PR #18, opened 2018-11-19
Fixes issue where callback was not invoked upon network error.

bradleykirwan:
Have kept the TODO in there, in case you want the reminder.


smtakeda:
@bradleykirwan thanks for PR. Could you elaborate how this change could help to catch the network error?

smtakeda:
ah, ok, I guess I started understanding the intention. By moving, the IF block to under else block, the network error will call the callback?

bradleykirwan:
@smtakeda Yes, currently network errors don't propagate through to the caller. I should really have commented out entire nested IF block, or perhaps remove it altogether.

smtakeda:
working on this.

PR #31, opened 2019-03-10
Fix toSfDate function by returning date object in correct timezone

### LINKED ISSUE #28: TIMESTAMP_NTZ column value is returned in local timezone instead of UTC
The `timestamp_ntz` column values are returned in the local timezone instead of UTC. 

The `convertRawTimestampNtz` [function](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0b25a78f3d5bae801e9b7c244d23ca75a5d18db6/lib/connection/result/column.js#L362) correctly passes the timezone to the `SfTimestamp` constructor and the moment object is in UTC timezone.

However, the [toSfDate](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0b25a78f3d5bae801e9b7c244d23ca75a5d18db6/lib/connection/result/sf_timestamp.js#L202) function constructs a Date object using the moment object.

``` 
var date = new Date(
      momentAsObject.years,
      momentAsObject.months,
      momentAsObject.date,
      momentAsObject.hours,
      momentAsObject.minutes,
      momentAsObject.seconds,
      momentAsObject.milliseconds);
```

But as per [Date object docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date), this would result in local time instead of UTC timezone-
> Note: Where Date is called as a constructor with more than one argument, the specified arguments represent local time. If UTC is desired, use new Date(Date.UTC(...)) with the same arguments

The bug can be resolved by using the `timezone` while constructing the date object. I can raise a PR with necessary fix, if required.

sksin28:
Fixes #28

### DIFF
--- a/lib/connection/result/sf_timestamp.js
+++ b/lib/connection/result/sf_timestamp.js
@@ -198,15 +198,7 @@ SfTimestamp.prototype.toString = function()
 SfTimestamp.prototype.toSfDate = function()
 {
   // create a Date from the moment
-  var momentAsObject = this.moment.toObject();
-  var date = new Date(
-      momentAsObject.years,
-      momentAsObject.months,
-      momentAsObject.date,
-      momentAsObject.hours,
-      momentAsObject.minutes,
-      momentAsObject.seconds,
-      momentAsObject.milliseconds);
+  var date = this.moment.toDate();
 
   var self = this;
 


smtakeda:
#40 will have the fix.

PR #35, opened 2019-04-09
changed winston.logger to winston.createLogger

Kevinyouk:
Accidentally closed last pull request.

### DIFF
--- a/lib/logger/node.js
+++ b/lib/logger/node.js
@@ -32,7 +32,7 @@ function Logger(options)
     // initialize the winston logger if needed
     if (!winstonLogger)
     {
-      winstonLogger = new winston.Logger(
+      winstonLogger = new winston.createLogger(
       {
         transports:
         [


lib/logger/node.js
```diff
@@ -32,7 +32,7 @@ function Logger(options)
     // initialize the winston logger if needed
     if (!winstonLogger)
     {
-      winstonLogger = new winston.Logger(
+      winstonLogger = new winston.createLogger(
```
smtakeda: @Kevinyouk can you elaborate what was the issue and how this change would address it?

notkriswagner: From reading the docs it seems like there was an interface change in Winston at v3.0.0 where they got rid of the `Logger` constructor and replaced it with `createLogger`. This seems like it could potentially be a breaking change for other customers

smtakeda: @notkriswagner , ok. but how could not we get any error or regression? And how can we verify this fix would work if no test failed?

notkriswagner: I guess there is no test that goes through this logic.

notkriswagner: I actually found a test where its failing to create the winston logger after the package update, so there was a test failure. You can find it here at line 991:

https://travis-ci.org/snowflakedb/snowflake-connector-nodejs/jobs/517462137

kavika-1: Just to confirm this is an issue:  Ran into this in production inspecting logs.  So when errors occur, like the recent OCSP issues, perhaps, there wasn't an a clear error message since the logging crashed.  Here's the stack.
````
2019-04-18T06:12:53.137Z 8ffe8b9f-e0c8-47d4-ada4-977629629774 TypeError: winston.Logger is not a constructor
at logMessage (/var/task/node_modules/snowflake-sdk/lib/logger/node.js:35:23)
at log (/var/task/node_modules/snowflake-sdk/lib/logger/core.js:312:7)
at Object.error (/var/task/node_modules/snowflake-sdk/lib/logger/core.js:193:7)
at Logger.error (/var/task/node_modules/snowflake-sdk/lib/logger/node.js:82:18)
at /var/task/node_modules/snowflake-sdk/lib/agent/socket_util.js:52:32
at /var/task/node_modules/snowflake-sdk/lib/agent/socket_util.js:168:9
at done (/var/task/node_modules/snowflake-sdk/lib/agent/check.js:53:5)
at ocspResponseVerify (/var/task/node_modules/snowflake-sdk/lib/agent/check.js:83:16)
at ClientRequest.done (/var/task/node_modules/ocsp/lib/ocsp/utils.js:26:7)
at emitOne (events.js:116:13)
at ClientRequest.emit (events.js:211:7)
at Socket.socketOnEnd (_http_client.js:391:9)
at emitNone (events.js:111:20)
at Socket.emit (events.js:208:7)
at endReadableNT (_stream_readable.js:1064:12)
at _combinedTickCallback (internal/process/next_tick.js:138:11)
at process._tickDomainCallback (internal/process/next_tick.js:218:9)
````

smtakeda:
PR will have the fix #36

PR #37, opened 2019-04-24
Properly destroy keepalive

veltman:
In v1.1.12 it looks like the `clearInterval()` call in `.destroy()` never executes, so a connection with keepalive set to `true` will keep the process running forever even after it's destroyed.

For example, this will hang:

```js
snowflake
  .createConnection({ ...connectionOptions, clientSessionKeepAlive: true })
  .connect((error, connection) => connection.destroy());
```

whereas this successfully destroys the connection:

```js
snowflake
  .createConnection({ ...connectionOptions, clientSessionKeepAlive: false })
  .connect((error, connection) => connection.destroy());
```

I've also found that although the docs say the default value for `clientSessionKeepAlive` is `false`, the default actually seems to be `true`, though it's not obvious to me where that's set.

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -28,8 +28,6 @@ function Connection(context)
   // generate an id for the connection
   var id = uuidv4();
 
-  var keepalive = null;
-
   /**
    * Returns true if the connection is active otherwise false
    *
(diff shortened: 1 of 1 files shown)


notkriswagner:
Thanks for this. I'll merge it in to the branch for our upcoming release

PR #56, opened 2019-09-08
SNOW-94651: adjust OCSP test case

smtakeda:


smtakeda:
This fixed the test failures Harsh fixed in SNOW-94651

PR #96, opened 2020-04-07
Workaround for SNOW-143538

snowstakeda:
SNOW-143538 Retry connection for ETIMEOUT tcp error.


snowstakeda:
The half of the above tests are not valid as they attempt to run tests from the original PR where no credential can be extracted.

This is the one.
https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/72870724

PR #112, opened 2020-07-06
Added request retry for EconnTimedout and a few other Econn failures using retry-request library

sfc-gh-cconner:


lib/http/node.js
```diff
@@ -3,7 +3,9 @@
  */
 
 const Url = require('url');
-const request = require('request');
+var request = require('retry-request', {
+  request: require('request')
+});
```
sfc-gh-kwagner, line 8: Can you add a default delay strategy to this (see https://www.npmjs.com/package/requestretry#how-to-define-your-own-delay-strategy and https://www.npmjs.com/package/requestretry#modifying-request-options). All of the other drivers use exponential backoff with jitter (see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)

sfc-gh-kwagner:
Closing this since it duplicates #117

PR #119, opened 2020-08-05
Added request retry for Econn failures using requestretry library and exp backoff w/ jitter using existing util.nextSleepTime

sfc-gh-cconner:


sfc-gh-kwagner:
Closing since its a duplicate

PR #126, opened 2020-09-23
SNOW-196924 Add precommit hooks for secret scanning

sfc-gh-hchaturvedi:


sfc-gh-stakeda (approved):

sfc-gh-kwagner (approved):

sfc-gh-hchaturvedi:
Hey @sfc-gh-kwagner I am seeing this unrelated test Windows (GCP) failing constantly with a schema not set error. Given that it is constantly failing is there something wrong you're aware of? I would think that this test is the same test just run on different cloud platforms which doesn't explain why it's failing as others are passing.

PR #140, opened 2021-01-30
Add external browser authentication

SimbaGithub:


lib/authentication/auth_web.js
```diff
@@ -0,0 +1,188 @@
+/*
+ * Copyright (c) 2015-2019 Snowflake Computing Inc. All rights reserved.
```
sfc-gh-hchaturvedi, line 2: Update Headers.

lib/connection/connection.js
```diff
@@ -88,14 +89,38 @@ function Connection(context)
...
+
   /**
    * Establishes a connection if we aren't in a fatal state.
    *
    * @param {Function} callback
    *
    * @returns {Object} the connection object.
    */
-  this.connect = function (callback)
+  this.connect = async function (callback)
```
sfc-gh-kwagner, line 123: Is making this async safe?

lib/util.js
```diff
@@ -443,3 +447,31 @@ const userAgent = 'JavaScript' + '/' + driverVersion
...
+      account = account.substring(0, account.indexOf('.'));
+    }
+    host = account + '.' + region + '.snowflakecomputing.com';
+  }
+  else
+  {
+    host = account + '.snowflakecomputing.com';
+  }
+  return host;
+};
```
sfc-gh-kwagner, line 477: nit: add newline at the end of file

sfc-gh-kwagner:
@SimbaGithub I ran these tests as well and it looks like your changes are in fact causing connections to Snowflake to take too long and timeout (as opposed to these test failures being caused by some other reason). Can you look into why connections aren't working?

lib/connection/connection.js
sfc-gh-kwagner: This could be why connections are timing out

SimbaGithub: Does having two connect() functions sound reasonable? An async connect() for external browser, and a non-async connect() for default, key-pair, and oauth

SimbaGithub:
> @SimbaGithub I ran these tests as well and it looks like your changes are in fact causing connections to Snowflake to take too long and timeout (as opposed to these test failures being caused by some other reason). Can you look into why connections aren't working?

Ok, will look into it

lib/connection/connection.js
sfc-gh-kwagner: Why do you need an async connect for external browser?

SimbaGithub: External browser's authentication returns a promise since retrieving the SAML token from the browser is done asynchronously.

With the async connect, it will wait for external browser's authenticate and services.sf.connect to finish execution before returning the connection. If it's not async, the connection may be returned without authenticate and services.sf.connect finishing their execution and making the changes it does to the connection

SimbaGithub: Passing services.sf.connect to authenticate() as a function parameter does remove the need for async on connect.
But executing a query afterwards fails, because a connection hasn't been established yet when the query is executed.

sfc-gh-kwagner: Okay that sounds reasonable. Btw, can you also add test cases for connecting with an external browser using this driver?

SimbaGithub: connect() will behave as before. And the async connect() is for external browser.
Sure, will add test cases for external browser connection

SimbaGithub:
@sfc-gh-kwagner I've added test cases for the default and external browser authenticator. Please review, thank you

sfc-gh-kwagner:
@SimbaGithub can you fix the remaining test failures?

SimbaGithub:
> @SimbaGithub can you fix the remaining test failures?

Tests are succeeding now. The integration tests for external browser are set to skipped, but can be tested manually

sfc-gh-kwagner (approved):

sfc-gh-hchaturvedi:
Not necessarily related to this PR alone but even though we don't specify a logging mechanism in our official docs, would it be possible for us to confirm whether turning on logging on the connector dumps anything sensitive?

sfc-gh-hchaturvedi (approved):
LGTM! Please ensure that the manual tests pass and the behavior is as expected.

SimbaGithub:
> LGTM! Please ensure that the manual tests pass and the behavior is as expected.

Confirming that manual tests passed on my end.

> Not necessarily related to this PR alone but even though we don't specify a logging mechanism in our official docs, would it be possible for us to confirm whether turning on logging on the connector dumps anything sensitive?

Using logLevel: trace, the log output shows the warehouse, database, and schema used in the connection string. I didn't see any sensitive info on there

PR #143, opened 2021-02-10
Add keypair/oauth authenticators

SimbaGithub:
Keypair authentication with private key or private key path with passphrase.
Oauth authentication with token.


SimbaGithub:
For keypair integration test, the "SNOWFLAKE_TEST_PRIVATE_KEY" environment value will have to be set with the private key value. The test case for passing private key path and passphrase are set to skipped. 

For oauth integration test, it's currently set to skip since it requires manual set-up to get a token

sfc-gh-kwagner:
@SimbaGithub can you look into and fix the test failures here?

SimbaGithub:
> @SimbaGithub can you look into and fix the test failures here?

@sfc-gh-kwagner Does this account have access to add environment variables or edit the parameters.json file? The two tests are failing because the private key has no value which it gets from SNOWFLAKE_TEST_PRIVATE_KEY.

I would need to add SNOWFLAKE_TEST_PRIVATE_KEY for the private key value and also SNOWFLAKE_JWT_TEST_USER for the private key user

SimbaGithub:
@sfc-gh-kwagner All tests are passing now. Also finished with manual testing on my end.

sfc-gh-kwagner (approved):

sfc-gh-kwagner:
@SimbaGithub can you rebase?

PR #153, opened 2021-03-10
Increase OCSP cache size

SimbaGithub:
By increasing the cache size, cache entries for already retrieved OCSP certificates will no longer be overwritten.

### DIFF
--- a/lib/agent/ocsp_response_cache.js
+++ b/lib/agent/ocsp_response_cache.js
@@ -23,18 +23,16 @@ const status = {
 };
 
 // validate input
-const capacity = GlobalConfig.getOcspResponseCacheMaxSize();
 const maxAge = GlobalConfig.getOcspResponseCacheMaxAge();
 
-Errors.assertInternal(Util.number.isPositiveInteger(capacity));
 Errors.assertInternal(Util.number.isPositiveInteger(maxAge));
 
 const cacheDir = GlobalConfig.mkdirCacheDir();
 const cacheFileName = path.join(cacheDir, "ocsp_response_cache.json");
-// create a cache to store the responses
-const cache = new SimpleCache({maxSize: capacity});
-// store previous cache's responses
-var prevCacheJSON;
+// create a cache to store the responses, dynamically changes in size
+var cache;
+// JSON object with previous cache's responses
+var prevCacheObj;
 
 function deleteCache()
 {
@@ -61,6 +59,7 @@ function OcspResponseCache()
 {
   let downloadStatus = status.NOT_START;
   let cacheUpdated = false;
+  let cacheInitialized = false;
 
   /**
    * Reads OCSP cache file.
@@ -76,24 +75,24 @@ function OcspResponseCache()
   {
     Logger.getInstance().debug("Reading OCSP cache file. %s", cacheFileName);
     const contents = fs.readFileSync(cacheFileName, 'utf-8');
-    let jsonParsed = JSON.parse(contents);
-    prevCacheJSON = jsonParsed;
-    for (let entry in jsonParsed)
-    {
-      if (jsonParsed.hasOwnProperty(entry))
-      {
-        const status = validateCacheEntry(entry, jsonParsed[entry]);
-        if (!status.err)
-        {
-          cache.set(status.key, status.value);
-        }
-      }
-    }
+    prevCacheObj = JSON.parse(contents);
+    cache = new SimpleCache({ maxSize: Object.keys(prevCacheObj).length });
+    setCacheEntries(prevCacheObj);
+    cacheInitialized = true;
   }
   catch (e)
   {
     Logger.getInstance().debug("Failed to read OCSP cache file: %s, err: %s", cacheFileName, e);
   }
+
+  /**
+   * Is OCSP Cache initialized?
+   * @returns {boolean}
+   */
+  this.isInitialized = function ()
+  {
+    return cacheInitialized;
+  };
 
   /**
    * Is OCSP Cache download finished?
@@ -131,7 +130,6 @@ function OcspResponseCache()
         cacheOutput[certIdInBase64] = [currentTime, ocspResponseInBase64];
       });
       const writeContent = JSON.stringify(cacheOutput);
-
       Logger.getInstance().debug("Writing OCSP cache file. %s", cacheFileName);
       try
       {
@@ -206,11 +204,7 @@ function OcspResponseCache()
       try
       {
         let jsonParsed = JSON.parse(cacheContent);
-        setCacheEntries(jsonParsed);
-        if (prevCacheJSON)
-        {
-          setCacheEntries(prevCacheJSON);
-        }
+        updateCache(jsonParsed);
         cacheUpdated = true;
         return cb(null, false);
       }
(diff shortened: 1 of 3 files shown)


sfc-gh-hchaturvedi:
@SimbaGithub I am a little surprised that 100 size cache is not enough for a single user. Besides that I feel that the size of the cache should be at least == size of the ocsp response cache downloaded from ocsp cache server 
( ocsp.snowflakecomputing.com ). The user must not in an ideal scenario need to have any more entries than that. Is there a chance that the cache has stale entries that are not being pruned?

SimbaGithub:
> @SimbaGithub I am a little surprised that 100 size cache is not enough for a single user. Besides that I feel that the size of the cache should be at least == size of the ocsp response cache downloaded from ocsp cache server
> ( ocsp.snowflakecomputing.com ). The user must not in an ideal scenario need to have any more entries than that. Is there a chance that the cache has stale entries that are not being pruned?

@sfc-gh-hchaturvedi The size of the OCSP response cache was 151-152. I found this out when I set the cache size to 1000, then the entries in the cache was either 151 or 152. I could set it to 152 exactly but I found that 150 was enough for OCSP certificate entries to not be overwritten

sfc-gh-hchaturvedi:
So you're saying that every time we downloaded the cache, some entries that were potentially interesting to the user were over written and hence the driver always has to go out and fetch them. This was causing the delay? And you've confirmed that if we increase the size the delay is reduced? Another thing I am curious about is what is the other solution you thought about?

SimbaGithub:
Yes, when it downloads the cache, some entries are overwritten. The sequence looks like this:
1. Fetch ocsp_response_cache.json from OCSP cache server
2. Write entries to cache (with cache size of 100, certificates are overwritten)
3. Driver contacts OCSP responders for certificates
4. Write certificates to cache (if one responder times out, then that certificate will be missing from the cache)
5. Driver contacts OCSP responder for missing certificate
6. Write certificate to cache (this overwrites already existing certificates in the cache)

The delay happens because of the loop between step 5 and 6 where some certificates have to be re-fetched which ends up overwriting the existing ones in the cache.

And yes, this reduces that delay. After increasing the cache size, the connector doesn't contact the OCSP responders anymore because it was able to keep the certificates retrieved from the OCSP cache server.
1. Fetch ocsp_response_cache.json from OCSP cache server
2. Write entries to cache (with cache size of 150, certificates are still in cache)
3. Driver doesn't need to contact OCSP responders

I attached a log file on the salesforce ticket for comparison

SimbaGithub:
The other solution was adding the entries from the previous cache on top of the new cache. The certificate entries will still be overwritten, but, the previous cache entries will be re-added afterwards. So all OCSP certificate entries will be in the cache.

1. Fetch ocsp_response_cache.json from OCSP cache server
2. Write entries to cache (with cache size of 100, certificates are overwritten)
3. Driver contacts OCSP responders for certificates
4. Write certificates to cache (if one responder times out, then that certificate will be missing from the cache)
5. Driver contacts OCSP responder for missing certificate
6. Write certificate to cache (this overwrites already existing certificates in the cache)
7. Write certificates from previous cache to cache

This alternative also prevents the loop from happening. But compared to increasing cache size, this was slower since the certificates are still overwritten and had to be fetched.

sfc-gh-hchaturvedi:
I have concerns about the scalability of this solution. What if we increase the size of the OCSP Cache so say tomorrow the cache size goes from 152 -> 200 .. Will we hit this problem again? Can we start off with a large buffer, add the ocsp cache to it and then append any **unexpired** entries from the older cache?

sfc-gh-dszmolka:
@sfc-gh-hchaturvedi  for completeness this is for SNOW-213402

SimbaGithub:
> Can we start off with a large buffer, add the ocsp cache to it and then append any **unexpired** entries from the older cache?

@sfc-gh-hchaturvedi  Yes, the two solutions together would be able to accomplish this. Is there a cache size you think it should start with or is 150 ok?

sfc-gh-hchaturvedi:
@SimbaGithub you mentioned that the ocsp cache downloaded has around 150 entries and we're planning to add older entries as well, do you think there can be some overwriting that might happen? Do we need a larger buffer?

SimbaGithub:
> @SimbaGithub you mentioned that the ocsp cache downloaded has around 150 entries and we're planning to add older entries as well, do you think there can be some overwriting that might happen? Do we need a larger buffer?

Yes, there are overwriting that happens, but the most used entries will take priority and remain in the cache.

I experimented with setting the OCSP cache size to only 4 (max number of certificates) and the certificates always remained on the cache. I'm assuming the reason is because they are the most used and take priority over the other entries.

SimbaGithub:
@sfc-gh-hchaturvedi Updated the fix to dynamically change the max size of the cache so existing entries are not overwritten.
Also, existing entries that are expired will be deleted

lib/agent/ocsp_response_cache.js
```diff
@@ -23,16 +23,16 @@ const status = {
...
 
-Errors.assertInternal(Util.number.isPositiveInteger(capacity));
 Errors.assertInternal(Util.number.isPositiveInteger(maxAge));
 
 const cacheDir = GlobalConfig.mkdirCacheDir();
 const cacheFileName = path.join(cacheDir, "ocsp_response_cache.json");
-// create a cache to store the responses
-const cache = new SimpleCache({maxSize: capacity});
+// create a cache to store the responses, dynamically changes in size
+var cache;
```
SimbaGithub, line 33: Removed the initialization here since it's now based on:
- the response cache's size or 
- the response cache's size + previous cache's size - # of overlap between the two caches

lib/agent/ocsp_response_cache.js
```diff
@@ -59,6 +59,7 @@ function OcspResponseCache()
 {
   let downloadStatus = status.NOT_START;
   let cacheUpdated = false;
+  let cacheInitialized = false;
```
SimbaGithub, line 62: Added a flag to indicate whether the cache is created, since the initialization is moved, it's possible the cache does not exist yet when OcspResponseCache is called.

lib/agent/socket_util.js
```diff
@@ -306,8 +306,13 @@ function validateCert(cert, cb)
       });
     }
 
-    // if we already have an entry in the cache, use it
-    let ocspResponse = getOcspResponseCache().get(cert);
+    let ocspResponse;
+    // check if cache is initialized before getting entry
+    if (getOcspResponseCache().isInitialized())
```
SimbaGithub, line 311: Added a check here because if "get(cert)" is called and the cache is not initialized, then an error is thrown due to cache being 'undefined'

lib/agent/ocsp_response_cache.js
```diff
@@ -320,6 +317,93 @@ function OcspResponseCache()
...
+    // Check if there are previous entries to append
+    if (prevCacheObj)
+    {
+      // Count overlap between previous cache and response cache
+      var cacheOverlapCount = 0;
+      for (let entry in jsonObject)
+      {
+        if (entryExists(prevCacheObj, entry))
+        {
+          cacheOverlapCount++;
```
sfc-gh-hchaturvedi, line 335: Is it possible to delete the entries in the older cache over here itself? This might save us the trouble of going through the entire older cache again at Line 351.

sfc-gh-hchaturvedi (commented):
Lets put an upper limit on the cache size that we don't think it will ever hit like 1000 or something. Just to avoid any corner case where the size of the cache turns about to be unrealistically big.

lib/agent/ocsp_response_cache.js
SimbaGithub: Done

SimbaGithub:
> Lets put an upper limit on the cache size that we don't think it will ever hit like 1000 or something. Just to avoid any corner case where the size of the cache turns about to be unrealistically big.

Sounds good. The PR is updated to have an upper limit on the cache size now

sfc-gh-hchaturvedi (approved):
LGTM! Please ensure all tests pass.

PR #155, opened 2021-03-25
Add host variable to connection config

SimbaGithub:
Set the host variable in the connection config to prevent part of its value being omitted


sfc-gh-kwagner (approved):
LGTM

PR #161, opened 2021-04-16
Draft - S3 library

SimbaGithub:


lib/connection/file_transfer_agent/s3_util.js
```diff
@@ -0,0 +1,206 @@
...
+  this.getFileHeader = async function (meta, filename)
+  {
+    var stageInfo = meta['stageInfo'];
+    var client = this.createClient(stageInfo);
+    var s3location = this.extractBucketNameAndPath(stageInfo['location']);
+
+    var params = {
+      Bucket: s3location.bucketName,
+      Key: filename
+    };
```
sfc-gh-kwagner, line 109: In the PythonConnector we don't use `filename` directly, we do this:

`s3path = s3location.s3path + filename.lstrip("/")`

And set that as the key. This could be why you're getting permission errors

SimbaGithub:
@sfc-gh-kwagner When copying data from the staged files, it just loads indefinitely. Would you know some possible causes as to why this is happening?

Here's what I found so far:
- Confirmed nodejs and python produce the same value for `enc_kek` and `iv_data` when encrypting given the same master key and random bytes
- Confirmed the compressed `.csv` file contains the correct data
- Different output size when compressing files with gzip for nodejs and python
- When the `AMZ_KEY` key is removed from the parameters, it returns an error: _Invalid data encountered during decompression for file: 'employees01.csv.gz',compression type used: 'GZIP', cause: 'data error'_

This is the command I use to copy data which works when files are staged with python:
```
copy into emp_basic
  from @%emp_basic
  file_format = (type = csv field_optionally_enclosed_by='"')
  pattern = '.*employees0[1-5].csv.gz'
  on_error = 'skip_file';
```

SimbaGithub:
After more testing, I found that copying data which is uploaded through S3 or Azure with the following query parameters have similar results:

- Uploading with both the `metadata` and `content-encoding` parameter: copying data results in loading indefinitely

- Uploading with only the `metadata` parameter: copying data results in loading indefinitely

- Uploading with only the `content-encoding` parameter: copying data returns the error `Invalid data encountered during decompression for file: 'employees01.csv.gz',compression type used: 'GZIP', cause: 'data error'`

- Uploading without the `metadata` and `content-encoding` parameter: copying data returns the error `Number of columns in file (2) does not match that of the corresponding table (6)`

SimbaGithub:
I was able to find what's causing the hanging issue:
`Unable to stream download: ('Received response with content-encoding: gzip, but failed to decode it.', error('Error -3 while decompressing data: incorrect header check'))`

I'm looking into how nodejs compresses files vs. python

SimbaGithub:
@sfc-gh-kwagner Looks like the 'hanging' issue happens when copying data for uncompressed uploads too. It results in the same error as compressed files:
`SQL execution internal error: Processing aborted due to error 300010:1489113183; incident 7589302.`

Here's the incident ID for uncompressed file:
`incident 7589302`

Incident ID for compressed file:
`incident 4985028`

Issue #165, opened 2021-05-14
SNOW-357899: Global variable leak: 'node', 'version'


jesse23:
Please answer these questions before submitting your issue. Thanks!

1. What version of NodeJS are you using (`node --version` and `npm --version`)?
v14.17.0

2. What operating system and processor architecture are you using?

Mac:
    > any version....

3. What are the component versions in the environment (`npm list`)?
     N/A

4. What did you do?
    host variable is leaking here:
    https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0c7eea1c3056c54f1a1a7b18b57a683393b8cb5b/lib/util.js#L484

https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0c7eea1c3056c54f1a1a7b18b57a683393b8cb5b/lib/connection/result/result.js#L50

It will cause unit test failure for global leak ( framework like mocha)

```
  1 failing

  1) "before each" hook:
     Error: global leak(s) detected: 'host', 'version'
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

5. What did you expect to see?
     define it ahead like `let host = ''`.

jesse23:
Could we get this addressed? Thx...

PR #166, opened 2021-05-17
SNOW-357899: Global variable leak: 'node', 'version'

jesse23:
SImple fix to 2 variable leak.

Issue link: https://github.com/snowflakedb/snowflake-connector-nodejs/issues/165


sfc-gh-dprzybysz:
It's already fixed in https://github.com/snowflakedb/snowflake-connector-nodejs/pull/277

PR #167, opened 2021-05-17
Check for privatelink and set as OCSP url if it exists

SimbaGithub:
Use privatelink for OCSP URL when the specified account is using privatelink.

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -11,6 +11,8 @@ var Statement = require('./statement');
 var Parameters = require('../parameters');
 var Authenticator = require('../authentication/authentication');
 
+const PRIVATELINK_URL_SUFFIX = ".privatelink.snowflakecomputing.com";
+
 /**
  * Creates a new Connection instance.
  *
@@ -139,7 +141,7 @@ function Connection(context)
       !Util.exists(callback) || Util.isFunction(callback),
       ErrorCodes.ERR_CONN_CONNECT_INVALID_CALLBACK);
 
-    if (connectionConfig.host.endsWith(".privatelink.snowflakecomputing.com"))
+    if (connectionConfig.host.endsWith(PRIVATELINK_URL_SUFFIX))
     {
       this.setupOcspPrivateLink(connectionConfig.host);
     }
@@ -199,7 +201,7 @@ function Connection(context)
       !Util.exists(callback) || Util.isFunction(callback),
       ErrorCodes.ERR_CONN_CONNECT_INVALID_CALLBACK);
 
-    if (connectionConfig.host.endsWith(".privatelink.snowflakecomputing.com"))
+    if (connectionConfig.host.endsWith(PRIVATELINK_URL_SUFFIX))
     {
       this.setupOcspPrivateLink(connectionConfig.host);
     }
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -100,6 +100,7 @@ function Connection(context)
   {
     var ocspCacheServer = `http://ocsp.${host}/ocsp_response_cache.json`;
     process.env.SF_OCSP_RESPONSE_CACHE_SERVER_URL = ocspCacheServer;
+    process.env.SF_OCSP_RESPONDER_URL = ocspCacheServer;
   }
 
   /**
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -89,6 +89,18 @@ function Connection(context)
     });
   };
 
+  /**
+  * Set the private link as the OCSP cache server's URL.
+  *
+  * @param {String} host
+  *
+  * @returns {null}
+  */
+  this.setupOcspPrivateLink = function (host)
+  {
+    var ocspCacheServer = `http://ocsp.${host}/ocsp_response_cache.json`;
+    process.env.SF_OCSP_RESPONSE_CACHE_SERVER_URL = ocspCacheServer;
+  }
 
   /**
   * Callback for connect() used to establish a connection.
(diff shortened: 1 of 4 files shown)


sfc-gh-hchaturvedi:
@SimbaGithub we also need to ensure that the ocsp responder url AND the ocsp cache server url both are updated to support privatelink. The URL received by `ocspRequestCallback` [here](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0c7eea1c3056c54f1a1a7b18b57a683393b8cb5b/lib/agent/check.js#L205) should also be updated for privatelink

SimbaGithub:
> @SimbaGithub we also need to ensure that the ocsp responder url AND the ocsp cache server url both are updated to support privatelink. The URL received by `ocspRequestCallback` [here](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0c7eea1c3056c54f1a1a7b18b57a683393b8cb5b/lib/agent/check.js#L205) should also be updated for privatelink

Done. Both `SF_OCSP_RESPONDER_URL` and `SF_OCSP_RESPONSE_CACHE_SERVER_URL` are updated with the privatelink

sfc-gh-hchaturvedi (changes requested):
@SimbaGithub OCSP URL and OCSP Cache Server URL are not the same. For privatelink OCSP URL can be built from privatelink OCSP CACHE SERVER URL by extracting the hostname and appending it with `retry/{actual ocsp url}/{actual ocsp request}` for example : 
`ocsp.abc.privatelink.snowflakecomputing.com/retry/ocsp.digicert.com/<BASE 64 OCSP Request>`
Consider how it's done in the python connector - https://github.com/snowflakedb/snowflake-connector-python/blob/aea85c48e1dcb5286bb6cefa45b537cd33a6cb20/src/snowflake/connector/ocsp_snowflake.py#L280

SimbaGithub:
@sfc-gh-hchaturvedi  Updated to set the OCSP responder url similar to how the python connector sets it. Am I understanding correctly that Python's `OCSP_RETRY_URL` is equivalent to NodeJS' `SF_OCSP_RESPONDER_URL`?

Note:
NodeJS' OCSP request contains an extra field `requestExtensions` that Python's OCSP request does not.

Here they are for comparison:
- NodeJS OCSP request
```
req = {
    tbsRequest: 
        version: 'v1',
        requestList: [ {
		        reqCert: certID
        } ],
        requestExtensions: [ {
		        extnID: rfc2560['id-pkix-ocsp-nonce'],
		        critical: false,
		        extnValue: rfc2560.Nonce.encode(crypto.randomBytes(16), 'der')
        } ]
  };
```  
- Python OCSP request
```
ocsp_request = OCSPRequest({
    'tbs_request': TBSRequest({
        'version': Version(0),
        'request_list': Requests([
	        Request({
		        'req_cert': cert_id,
	        })]),
	}),
})
```
I checked Python's base64 value for OCSP request and it's very similar to NodeJS' base64 value but NodeJS is longer due to the extra field.

Also, `parsedUrl.protocol` already has a ':' so the extra ':' is removed.

sfc-gh-hchaturvedi:
> @sfc-gh-hchaturvedi Updated to set the OCSP responder url similar to how the python connector sets it. Am I understanding correctly that Python's `OCSP_RETRY_URL` is equivalent to NodeJS' `SF_OCSP_RESPONDER_URL`?
> 
> Note:
> NodeJS' OCSP request contains an extra field `requestExtensions` that Python's OCSP request does not.
> 
> Here they are for comparison:
> 
> * NodeJS OCSP request
> 
> ```
> req = {
>     tbsRequest: 
>         version: 'v1',
>         requestList: [ {
> 		        reqCert: certID
>         } ],
>         requestExtensions: [ {
> 		        extnID: rfc2560['id-pkix-ocsp-nonce'],
> 		        critical: false,
> 		        extnValue: rfc2560.Nonce.encode(crypto.randomBytes(16), 'der')
>         } ]
>   };
> ```
> 
> * Python OCSP request
> 
> ```
> ocsp_request = OCSPRequest({
>     'tbs_request': TBSRequest({
>         'version': Version(0),
>         'request_list': Requests([
> 	        Request({
> 		        'req_cert': cert_id,
> 	        })]),
> 	}),
> })
> ```
> 
> I checked Python's base64 value for OCSP request and it's very similar to NodeJS' base64 value but NodeJS is longer due to the extra field.
> 
> Also, `parsedUrl.protocol` already has a ':' so the extra ':' is removed.

Since the extension's criticality is false setting or unsetting it shouldn't really make a difference. Anyways the fix looks good to me.  However let's add a test - let's force privatelink and mock an ocsp URL and a base64 request and check the final environment variable being exported for both the cache and the final SF_OCSP_RESPONDER_URL.

SimbaGithub:
> @SimbaGithub Since the extension's criticality is false setting or unsetting it shouldn't really make a difference. Anyways the fix looks good to me. However let's add a test - let's force privatelink and mock an ocsp URL and a base64 request and check the final environment variable being exported for both the cache and the final SF_OCSP_RESPONDER_URL.

New OCSP tests are added. I used one of the existing test accounts and added ".privatelink" to the account name so there's no need for new environment variables.

Now it also checks if the `SF_OCSP_RESPONSE_CACHE_SERVER_URL` value contains 'ocsp' because some of the tests from _testConnectionWithOCSP.js_ sets the value for that env variable and would fail.

The `Account without privatelink` test connects with a non-privatelink account and checks that the OCSP variables are unchanged. The `Account with privatelink` test connects with the privatelink account then calls the check function with the mock url and mock ocsp request.

lib/connection/connection.js
```diff
@@ -127,6 +139,11 @@ function Connection(context)
       !Util.exists(callback) || Util.isFunction(callback),
       ErrorCodes.ERR_CONN_CONNECT_INVALID_CALLBACK);
 
+    if (connectionConfig.host.endsWith(".privatelink.snowflakecomputing.com"))
```
sfc-gh-kwagner, line 142: nit: make this into a constant and use the constant here and below

sfc-gh-kwagner (approved):
One minor comment but approved overall

lib/connection/connection.js
SimbaGithub: Ok, constant for privatelink suffix is now made

SimbaGithub:
@sfc-gh-hchaturvedi just following up, please review when available thanks

sfc-gh-hchaturvedi (approved):
LGTM! Thanks @SimbaGithub .

sfc-gh-hchaturvedi:
Please confirm test failure reasons before merging.

PR #169, opened 2021-05-19
Implement PUT command

SimbaGithub:


SimbaGithub:
@sfc-gh-kwagner Please review when available, thank you

lib/connection/statement.js
```diff
@@ -29,23 +30,70 @@ var statementTypes =
     FILE_PRE_EXEC: 'FILE_PRE_EXEC'
```
sfc-gh-kwagner, line 30: We don't need a `FILE_POST_EXEC` flag here for when we're done with PUT command executing?

lib/connection/statement.js
```diff
@@ -472,7 +517,7 @@ function BaseStatement(
    * @param err
    * @param body
    */
-  context.onStatementRequestComp = function (err, body)
+  context.onStatementRequestComp = async function (err, body)
```
sfc-gh-kwagner, line 520: +1 for async/await changes. Always better to make things async

lib/file_transfer_agent/encrypt_util.js
```diff
@@ -0,0 +1,223 @@
...
+  * @param {String} tmpDir
+  * @param {Number} chunkSize
+  *
+  * @returns {Object}
+  */
+  this.encryptFile = async function (encryptionMaterial, inFileName,
+    tmpDir = null, chunkSize = blockSize * 4 * 1024)
+  {
+    // Get decoded key from base64 encoded value
+    var decodedKey = Buffer.from(encryptionMaterial['queryStageMasterKey'], 'base64');
```
sfc-gh-kwagner, line 92: nit: Turn 'queryStageMasterKey' into a named constant

sfc-gh-kwagner: Same with 'base64' since its used in a few places

test/unit/file_transfer_agent/encrypt_util_test.js
```diff
@@ -0,0 +1,134 @@
...
+
+  var EncryptionUtil;
+  var encrypt;
+  var filestream;
+  var temp;
+
+  this.beforeEach(function ()
+  {
+    encryptionMaterial = {
+      queryStageMasterKey: 'ztke8tIdVt1zmlQIZm0BMA==',
```
sfc-gh-kwagner, line 26: Is this a valid queryStageMasterKey or a dummy one that is just used for testing?

lib/util.js
```diff
@@ -489,3 +489,14 @@ exports.construct_hostname = function (region, account)
...
+
+/**
+ * Returns if command is a PUT command
+ *
+ * @param sqlText the query command
+ * @returns {boolean}
+ */
+exports.isPutCommand = function (sqlText)
+{
+  return (sqlText.substring(0, 3) === "put" || sqlText.substring(0, 3) === "PUT");
```
sfc-gh-kwagner, line 501: Alternatively you could capitalize `sqlText.substring(0, 3)` and just do one comparison

lib/connection/statement.js
SimbaGithub: I'll create a FILE_POST_EXEC flag. 

The ROW_POST_EXEC flag is used to check if Parameters.update() should be called or not, I'll add FILE_POST_EXEC to that check as well.

lib/file_transfer_agent/encrypt_util.js
SimbaGithub: Done

test/unit/file_transfer_agent/encrypt_util_test.js
SimbaGithub: It's based from a python test which I believe is a dummy: https://github.com/snowflakedb/snowflake-connector-python/blob/6cec12333927ac4f85bc0c9317dde7baf7097364/test/unit/test_encryption_util.py#L24

lib/util.js
SimbaGithub: That's a good idea, done

lib/file_transfer_agent/file_compression_type.js
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright (c) 2015-2021 Snowflake Computing Inc. All rights reserved.
+ */
+
+const Types = [
```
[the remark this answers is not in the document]
SimbaGithub: Ok, added test for uploading with different compression types

SimbaGithub:
> Overall looks good. Can you add a test for uploading some different compression types and then I can approve this.
> 
> Also in the future, could you split up non-essential changes from PRs. The async/await stuff could have been in a separate PR which would have helped me finish the PUT support PR faster

@sfc-gh-kwagner Sure thing, I'll split up the non-related changes to its own PR for the next one

The account being used in connOption.valid may need some permissions to be added. 
The test creates a temporary warehouse and database but it looks like there's "Insufficient privileges" to execute the query
```
       "before each" hook for "S3 upload - gzip":
     OperationFailedError: SQL access control error:
Insufficient privileges to operate on account 'SFCTEST0'
```

Twixes:
I suppose this doesn't support local PUT uploads yet? Would really love to be able to use that too

sfc-gh-kwagner:
@SimbaGithub can you look into and fix these test issues?

Twixes:
At PostHog we'd absolutely love to uploads files from local filesystem into an internal Snowflake stage this way, though seems this won't be possible with this SDK for quite a while. We'll probably drop it, or _maybe_ do this in a bespoke way in our service.

SimbaGithub:
> @SimbaGithub can you look into and fix these test issues?

@sfc-gh-kwagner The test failures are due to the SFCTEST0' account not having sufficient privilege:
`Insufficient privileges to operate on account 'SFCTEST0'`

There's more detail in the previous post above but essentially the test creates a temporary database for uploading. Is it possible for you to add privileges to that account or is there an existing database/warehouse that can be used for testing instead?

SimbaGithub:
> I suppose this doesn't support local PUT uploads yet? Would really love to be able to use that too

The PR would allow uploading local files onto an internal stage

SimbaGithub:
@sfc-gh-cshi The previous error is fixed and Linux and Mac are now passing.
Windows is failing due to this query `PUT file://C:\Users\RUNNER~1\...` which causes the error `syntax error line 1 at position 26 unexpected '~'.`

I can reproduce it locally and the error still occurs after adding backslash to escape the '~' character

lib/file_transfer_agent/file_transfer_agent.js
```diff
@@ -0,0 +1,713 @@
...
+  *
+  * @param {String} stageLocationType
+  *
+  * @returns {Object}
+  */
+  function getStorageClient(stageLocationType)
+  {
+    if (stageLocationType == LOCAL_FS)
+    {
+      return new SnowflakeLocalUtil();
```
sfc-gh-cshi, line 441: Uncaught TypeError: SnowflakeLocalUtil is not a constructor
      at getStorageClient (lib/file_transfer_agent/file_transfer_agent.js:441:14)

SimbaGithub: Could you please check if the last commit fixed this?

lib/file_transfer_agent/local_util.js
```diff
@@ -0,0 +1,58 @@
...
+      var reader = fs.createReadStream(meta['realSrcFilePath']);
+      var output = path.join(meta['stageInfo']['location'], meta['dstFileName']);
+
+      // expand '~' and '~user' expressions
+      if (process.platform !== "win32")
+      {
+        output = expandTilde(output);
+      }
+
+      var writer = fs.createWriteStream(output);
```
sfc-gh-cshi, line 44: The path might not exist at this moment.

SimbaGithub: Updated to create the directory if it doesn't exist.

sfc-gh-cshi (approved):
LGTM. I would like to see more test cases to cover more error scenarios.

PR #170, opened 2021-06-01
Add Secret Detector and unit tests

SimbaGithub:


SimbaGithub:
@sfc-gh-hchaturvedi The python connector uses Secret Detector for masking telemetry data but node connector doesn't seem to use telemetry. The logging in node doesn't contain sensitive information as well but if there are parts you think the Secret Detector should be used then please let me know, thanks

sfc-gh-hchaturvedi:
@SimbaGithub I am trying to understand how this module is being used? Do we have logging in nodeJS? If yes then all logs should pass through this secret detector. 

Another question - is it possible for us to add a custom pattern that can be added to the secret detector during init? I know this is not a part of the python connector but something I have been thinking might help quick mitigation in the future if something new comes up.  This question applies to the .NET fix as well. 

Thanks.

SimbaGithub:
> @SimbaGithub I am trying to understand how this module is being used? Do we have logging in nodeJS? If yes then all logs should pass through this secret detector.
> 
> Another question - is it possible for us to add a custom pattern that can be added to the secret detector during init? I know this is not a part of the python connector but something I have been thinking might help quick mitigation in the future if something new comes up. This question applies to the .NET fix as well.
> 
> Thanks.

@sfc-gh-hchaturvedi The logger now uses the SecretDetector to mask each message before they are logged.
And custom patterns can be added to the SecretDetector on init:
```
var customPatterns = {
  regex: [
	String.raw`(testCustomPattern\s*:\s*"([a-z]{8,})")`,
	String.raw`(testCustomPattern\s*:\s*"([0-9]{8,})")`
  ],
  mask: [
	'maskCustomPattern1',
	'maskCustomPattern2'
  ]
}
var SecretDetector = new (require('../secret_detector.js'))(customPatterns);
```

PR #173, opened 2021-06-11
Add "validateDefaultParameters" option

SimbaGithub:
Make connection params case insensitive.

Also added test cases for custom parameters:
- For external params, the value should be overridden
- For non-external params, the value should stay as the default


lib/connection/connection_config.js
```diff
@@ -11,42 +11,74 @@ const NativeTypes = require('./result/data_types').NativeTypes;
...
+  CLIENT_SESSION_KEEP_ALIVE: 'CLIENTSESSIONKEEPALIVE',
+  CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY: 'CLIENTSESSIONKEEPALIVEHEARTBEATFREQUENCY',
+  JS_TREAT_INTEGER_AS_BIGINT: 'JSTREATINTEGERASBIGINT',
+  SESSION_TOKEN: 'SESSIONTOKEN',
+  MASTER_TOKEN: 'MASTERTOKEN',
+  SESSION_TOKEN_EXPIRATION_TIME: 'SESSIONTOKENEXPIRATIONTIME',
+  MASTER_TOKEN_EXPIRATION_TIME: 'MASTERTOKENEXPIRATIONTIME',
+  AGENT_CLASS: 'AGENTCLASS'
+};
+
```
sfc-gh-kwagner, line 45: Can you change these to be snake case? Its a little hard to read as its currently written

SimbaGithub: Changing the values to snake case would require the connection string parameters to be typed in snake case as well. One word parameters will be unaffected but parameters with two or more will be affected
```
var connection = snowflake.createConnection({
  access_url: "test",
  private_key: "test",
  proxy_host: "test",
  account: "test",
  username: "test",
  password: "test",
```

It's possible to keep the connection string the same using a module to convert to snake case but it won't be case insensitive
```
var connection = snowflake.createConnection({
  accessUrl: "test",
  aCcount: "test",
```
In this case, "accessUrl" is correctly converted to "access_url" but "aCcount" would be converted to "a_C_count"

The parameters were converted to uppercase so it would be case insensitive but it does become hard to read. It could be converted to lowercase if it's easier

sfc-gh-kwagner:
@SimbaGithub is there a ticket for this change? Also can you fix the branch conflicts?

SimbaGithub:
> @SimbaGithub is there a ticket for this change? Also can you fix the branch conflicts?

@sfc-gh-kwagner This is the corresponding Salesforce ticket: 00335178
And will do on the branch conflicts

sfc-gh-kwagner:
@SimbaGithub I looked over the ticket and have talked to a couple people on the team and we are not going to merge this in for a couple of reasons:

1. All the other drivers have case-sensitive login parameter keys and merging this would cause inconsistent behavior between the drivers and in certain cases its not possible to make these login parameter keys case-insensitive in certain drivers
2. We have something in the Python connector that tells the customer they are submitting an incorrect login parameter key by printing to stderr. I think we should do something similar for the NodeJS driver. Here is the link to the Python connector feature https://docs.snowflake.com/en/release-notes/2020-07.html#snowflake-connector-for-python-report-invalid-parameters-passed-to-connect-method

lib/constants/error_messages.js
```diff
@@ -51,6 +51,7 @@ exports[404026] = 'Invalid private key. The specified value must be a string in
 exports[404027] = 'Invalid private key file location. The specified value must be a string';
 exports[404028] = 'Invalid private key passphrase. The specified value must be a string';
 exports[404029] = 'Invalid oauth token. The specified value must be a string';
+exports[404030] = 'Invalid validate default parameters. The specified value must be a boolean';
```
sfc-gh-kwagner, line 54: nit: Invalid validateDefaultParameters value. The specified value must be a boolean

sfc-gh-kwagner:
@SimbaGithub can you add a test case where you try to connect with a parameter that has the wrong case and the check through the logs to make sure that the warning messages function as expected

lib/constants/error_messages.js
SimbaGithub: Done

SimbaGithub:
> @SimbaGithub can you add a test case where you try to connect with a parameter that has the wrong case and the check through the logs to make sure that the warning messages function as expected

Added test cases for valid/invalid parameters that checks the console messages

lib/connection/connection_config.js
```diff
@@ -309,6 +343,32 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo)
...
+  {
+    for (const [key] of Object.entries(options))
+    {
+      if (!DEFAULT_PARAMS.includes(key))
+      {
+        var matches = stringSimilarity.findBestMatch(key, DEFAULT_PARAMS);
+        console.log(`"${key}" is an unknown connection parameter`);
+        if (matches.bestMatchIndex > 0)
+        {
+          console.log(`Did you mean "${matches.bestMatch.target}"`);
```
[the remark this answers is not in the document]
SimbaGithub: Ok

sfc-gh-kwagner (approved):

PR #185, opened 2021-08-17
Set 'keepAlive' option to true

SimbaGithub:
SF #00338640

Set keepAlive to true so the connector re-uses the sockets.

### DIFF
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -59,7 +59,7 @@ NodeHttpClient.prototype.getAgentAndProxyOptions = function (url, proxy, mock)
 {
   const isHttps = Url.parse(url).protocol === 'https:';
   let agentClass;
-  let agentOptions = { keepAlive: false };
+  let agentOptions = { keepAlive: true };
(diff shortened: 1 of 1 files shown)


sfc-gh-kwagner:
@SimbaGithub can you create/run some performance tests for this change so that we are sure that this improves performance?

SimbaGithub:
> @SimbaGithub can you create/run some performance tests for this change so that we are sure that this improves performance?

@sfc-gh-kwagner After creating a connection, the test asynchronously executes the query "SELECT L_COMMENT from SNOWFLAKE_SAMPLE_DATA.TPCH_SF100.LINEITEM limit 10000;" in a loop.

- keepAlive: false
connecting: 1.097s
query average of 100 runs: 1968.465s
execution: 3:17m 

- keepAlive: true
connecting: 1.052s
query average of 100 runs: 368.857s
execution: 37.953s

Setting keepAlive to false/true is done manually so it's skipped for CI tests.

PR #186, opened 2021-08-17
Add connection pool capability to connector

SimbaGithub:
SF #00284802

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -278,7 +278,8 @@ function Core(options)
     */
     this.validate = function (connection)
     {
-      return (connection.isTokenValid() && connection.isUp());
+      var heartbeat = await connection.heartbeatAsync();
+      return ((heartbeat[0][1] == 1) && connection.isUp());
     }
   }
 
--- a/lib/core.js
+++ b/lib/core.js
@@ -278,7 +278,7 @@ function Core(options)
     */
     this.validate = function (connection)
     {
-      return connection.isUp();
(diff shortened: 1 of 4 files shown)


sfc-gh-kwagner:
@SimbaGithub can you add more tests for failure modes like trying to acquire more than the max number of connections. Or changing the number of min/max connections in the pool.

Also you need to rebase

lib/core.js
```diff
@@ -196,6 +210,96 @@ function Core(options)
...
+    /**
+    * Returns the status of the connection.
+    *
+    * @param {Object} connection
+    *
+    * @returns {Boolean}
+    */
+    this.validate = function (connection)
+    {
+      return connection.isUp();
```
sfc-gh-kwagner, line 281: I don't think this is sufficient. The connection could be completely expired (both master and session token are expired) and `connection.isUp()` would return true (eventhough it should be false) since we haven't tried to make a request with the expired keys so we haven't moved into a new connection state. In the validate function you'll have to ensure that not only has the connection connected before, but that the session for that connection is still valid

SimbaGithub: _"In the validate function you'll have to ensure that not only has the connection connected before"_
 - In the create function, a new connection instance will attempt to connect so each connection in the pool will have connected before
 
_"but that the session for that connection is still valid"_
 - Is the new isTokenValid() enough? It checks if both session token and master token's time are greater than current time

sfc-gh-kwagner (approved):

PR #196, opened 2021-09-02
Add proxy user/password to connection settings

SimbaGithub:
SF# 00344819


sfc-gh-hchaturvedi:
Hey @SimbaGithub, I would suggest looking into supporting HTTPS protocol for proxy connect as well. Consider this PR in JDBC - https://github.com/snowflakedb/snowflake-jdbc/pull/557. 

If the driver tries to connect via proxy and the proxy is authenticated, if the driver is not using https then it will end up sending the proxy credentials over HTTP (not HTTPS) which would mean the credentials are communicated in the clear over the wire.  The customer should have the ability to configure the driver to initiate the initial connection to the proxy via HTTPS. 

The PR I shared adds the capability to JDBC.

SimbaGithub:
@sfc-gh-hchaturvedi I found the connector already had code that uses TLS when connecting to the proxy server but it wasn't being hit before because the "protocol" property was never set: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/903d6475364dda3bcb630223666adfa4b5c28e52/lib/agent/https_proxy_ocsp_agent.js#L52

The connector will now add a "protocol" property when doing authenticated proxy so that "secureProxy" will be true and it will connect using TLS
https://github.com/snowflakedb/snowflake-connector-nodejs/blob/903d6475364dda3bcb630223666adfa4b5c28e52/lib/agent/https_proxy_ocsp_agent.js#L85-L88

sfc-gh-hchaturvedi:
> @sfc-gh-hchaturvedi I found the connector already had code that uses TLS when connecting to the proxy server but it wasn't being hit before because the "protocol" property was never set:
> 
> https://github.com/snowflakedb/snowflake-connector-nodejs/blob/903d6475364dda3bcb630223666adfa4b5c28e52/lib/agent/https_proxy_ocsp_agent.js#L52
> 
> The connector will now add a "protocol" property when doing authenticated proxy so that "secureProxy" will be true and it will connect using TLS
> https://github.com/snowflakedb/snowflake-connector-nodejs/blob/903d6475364dda3bcb630223666adfa4b5c28e52/lib/agent/https_proxy_ocsp_agent.js#L85-L88

That's great! Can we please confirm that it actually works? Essentially would it would mean is set up an authenticated proxy and run a connection through it, do a packet capture and try to look at the credentials in the pcap with and without the protocol field set to https.

SimbaGithub:
> That's great! Can we please confirm that it actually works? Essentially would it would mean is set up an authenticated proxy and run a connection through it, do a packet capture and try to look at the credentials in the pcap with and without the protocol field set to https.

Is setting up an authenticated proxy something we do on our side? Would you have some resources on setting up a proxy for a Snowflake account?

SimbaGithub:
@sfc-gh-hchaturvedi I was able to test using HTTP and HTTPS to connect to proxy and confirm that the packets are encrypted when using TLS. The capability to specify which protocol to use when connecting to the proxy server is added as well. Please take a look, thanks

lib/connection/connection_config.js
```diff
@@ -193,11 +196,41 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo)
...
+      Errors.checkArgumentValid(Util.isString(proxyPassword),
+        ErrorCodes.ERR_CONN_CREATE_INVALID_PROXY_PASS);
+
+      proxy =
+      {
+        host: proxyHost,
+        port: proxyPort,
+        user: proxyUser,
+        password: proxyPassword,
+        protocol: proxyProtocol
```
sfc-gh-hchaturvedi, line 223: Is there any default value for `proxyProtocol`? What happens if the customer does not configure the `proxyProtocol`? Also It would be great if we can add a warning log statement if the customer is using an authenticated proxy but has not set `proxyProtocol` to `https`

SimbaGithub: _Is there any default value for proxyProtocol? What happens if the customer does not configure the proxyProtocol?_
- If no value is set for `proxyProtocol`, then the connector will use HTTP to connect to the proxy server. 

_Also It would be great if we can add a warning log statement if the customer is using an authenticated proxy but has not set proxyProtocol to https_
 - Sounds good, I'll add a warning log for that

SimbaGithub: @sfc-gh-hchaturvedi The warning message is added, please review again. Thanks

SimbaGithub:
> LGTM! I think we'd need a doc update for this one. Do you know how to get docs going for this @SimbaGithub or should I start the process on my side?

Thanks for the review, I'll update the docs team through the channel

Issue #193, opened 2021-08-24
Using `accessUrl` with `1.6.2` throws `Cannot read property 'endsWith' of undefined`


curtislarson:
Please answer these questions before submitting your issue. Thanks!

1. What version of NodeJS are you using (`node --version` and `npm --version`)?

```bash
$ node --version && npm --version
v14.16.0
7.20.2
```

2. What operating system and processor architecture are you using?

```bash
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: macOS 10.14.6 (18G9028)
      Kernel Version: Darwin 18.7.0
      Boot Volume: Macintosh HD
      Boot Mode: Normal
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
```

3. What are the component versions in the environment (`npm list`)?

```bash
$ npm list
@quackware/snowflake-undefined-host
├── @jest/types@27.0.6
├─┬ @quackware/snowflake-1.6.1@ -> ./packages/snowflake-1.6.1
│ ├── @types/snowflake-sdk@1.6.0
│ └── snowflake-sdk@1.6.1
├─┬ @quackware/snowflake-1.6.2@ -> ./packages/snowflake-1.6.2
│ ├── @types/snowflake-sdk@1.6.0 deduped
│ └── snowflake-sdk@1.6.2
├── @types/jest@26.0.24
├── jest@27.0.6
├── rimraf@3.0.2
├── ts-jest@27.0.5
└── typescript@4.3.5

```

4. What did you do?

Upgrading `snowflake-sdk` to `1.6.2` breaks `Connection.connect()` when using `accessUrl`.

Full runnable program: https://github.com/quackware/snowflake-undefined-host

5. What did you expect to see?

connect working

6. What did you see instead?

connect not working

7. Add this to get standard output.

Failing `connect` generates no output, see https://github.com/quackware/snowflake-undefined-host

PR #202, opened 2021-09-21
BUGFIX for #193: Populate options.host when account is inferred

thedwitter:
This is my first PR so please be gentle!!  My company uses Snowflake internally and I'm excited to contribute!

I took a look at #193 and noticed the root cause seems to be that options.host is typically inferred from options.accounts, and so when options.accounts is inferred from the options.accessUrl it will remain unset and result in a crash starting with 1.6.2.  The fix is to also set options.host when inferring the options.account.

I've also added a test case. 

~ Matt


thedwitter:
Hi @sfc-gh-abhatnagar, are you available to review this for me?  Thanks!!

~ Matt

thedwitter:
@SimbaGithub does this repository accept public pull requests?  Anything I can do to help get this merged?

~ Matt

Issue #217, opened 2021-10-16
NODE.js driver will always return an error when authenticating with externalbrowser


orellabac:
Please answer these questions before submitting your issue. Thanks!

1. What version of NodeJS are you using (`node --version` and `npm --version`)?
1.6.4 
2. What operating system and processor architecture are you using?

Linux:
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal

3. What are the component versions in the environment (`npm list`)?
/workspace/test
└── (empty)

4. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.
    // Create a Connection object that we can use later to connect.
    var connectionInfo = {
        account: account,
        username: user,
        password: password,
        database: database,
        warehouse: warehouse,
        role: role,
        authenticator: authenticator,
        schema: schema
    };
 connection.connectAsync(function (err, conn) {
            connecting = false;
            if (err) {
                logger.error({ message: 'unable to connect', reason: err.message });
                connectionError = err;
            }
            else {
                connected = true;
                logger.info('Successfully connected to Snowflake.');
            }
        });
5. What did you expect to see?
No errors and a successful connection

6. What did you see instead?

The browser is open to accept the authentication, but when there is a redirection to localhost the answers is that there was a responde error

7. Add this to get standard output.
 
```
var snowflake = require('./snowflake');
snowflake.configure(
{
  logLevel: 'trace'
});
```

PR #218, opened 2021-10-16
Update auth_web, to avoid error message #217

orellabac:
In the current implementation the success message that is returned will be reported as an error.

I looked at the Python Connector and adapted it to return the same msg


sfc-gh-dszmolka:
marking this old PR as closed; if it's still relevant please comment and link the issue it is aiming to solve. Thank you !

PR #221, opened 2021-11-01
Add GET command

SimbaGithub:


lib/file_transfer_agent/azure_util.js
```diff
@@ -236,6 +236,65 @@ function azure_util(azure, filestream)
...
+    var blockBlobClient = containerClient.getBlockBlobClient(blobName);
+
+    try
+    {
+      var downloadBlockBlobResponse  = await blockBlobClient.download(0);
+      var readableStream = downloadBlockBlobResponse.readableStreamBody;
+
+      await new Promise((resolve, reject) =>
+      {
+        var writer = fs.createWriteStream(fullDstPath);
```
sfc-gh-kwagner, line 266: Does this create the destination path if it doesn't exist? If not then we should be creating this directory before trying to download any files

sfc-gh-kwagner (approved):
Just one comment

lib/file_transfer_agent/azure_util.js
SimbaGithub: The folder is created before in file_transfer_agent.js
https://github.com/snowflakedb/snowflake-connector-nodejs/blob/ba5859a994757e46fbbeb5faeef5c10a6db8762d/lib/file_transfer_agent/file_transfer_agent.js#L118-L123

PR #227, opened 2021-11-11
SNOW-499880 fix docker image for build

sfc-gh-abhatnagar:
Build image got deleted in nexus . Fixing this dockerfile to push the image


sfc-gh-kdama:
I dont think we would need this anymore as the pipelines are working fine and this is a old PR.

Issue #222, opened 2021-11-04
Setting "application" field from SDK does not attach a "Client Application" field to queries


andrewpage:
Please answer these questions before submitting your issue. Thanks!

1. What version of NodeJS are you using (`node --version` and `npm --version`)?
```
$ node -v
v16.6.1
$ npm -v
7.20.3
```

2. What operating system and processor architecture are you using?
```
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: macOS 12.0.1 (21A559)
      Kernel Version: Darwin 21.1.0
      Boot Volume: Macintosh HD
      Boot Mode: Normal
      Computer Name: xxx
      User Name: xxx
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
      Time since boot: 4:09
```

3. What are the component versions in the environment (`npm list`)?
```
├── @babel/core@7.10.3
├── @babel/preset-env@7.10.3
├── @babel/preset-typescript@7.13.0
├── @databases/sqlite@3.0.0
├── @google-cloud/bigquery@5.6.0
├── @google-cloud/storage@5.13.1
├── @graphql-tools/merge@6.2.14
├── @looker/sdk@0.3.4-beta.2
├── @octokit/auth-app@3.5.3
├── @octokit/rest@18.5.3
├── @rudderstack/rudder-sdk-node@1.0.2
├── @slack/web-api@5.15.0
├── @stedi-oss/typesuite@0.0.0 invalid: "git+https://github.com/hightouchio/typesuite.git#main" from the root project
├── @types/btoa@1.2.3
├── @types/express-session@1.17.3
├── @types/express@4.17.11
├── @types/http-proxy-middleware@0.19.3
├── @types/jest@26.0.1
├── @types/jsforce@1.9.29
├── @types/jsonwebtoken@8.5.1
├── @types/lodash@4.14.169
├── @types/mysql@2.15.18
├── @types/node@13.13.52
├── @types/passport-google-oauth20@2.0.7
├── @types/passport@1.0.6
├── @types/pino@6.3.8
├── @types/promise.allsettled@1.0.3
├── @types/redis@2.8.28
├── @types/request-promise-native@1.0.17
├── @types/xml2js@0.4.8
├── abort-controller@3.0.0
├── airtable@0.9.0
├── alasql@0.6.7
├── apollo-cache-inmemory@1.6.6
├── apollo-client@2.6.10
[...]

PR #239, opened 2021-12-03
FIX 222 - Pass client application on connect

andrewpage:
https://github.com/snowflakedb/snowflake-connector-nodejs/issues/222

The "Client Application" field in Snowflake "Client Info" does not pull from `CLIENT_APP_NAME`, it pulls from `CLIENT_ENVIRONMENT.APPLICATION`.


sfc-gh-dszmolka:
hi and thank you for your contribution - i think this has been fixed earlier with 1.6.10

PR #251, opened 2022-01-13
BUGFIX for #193: Populate options.host when account is inferred #202 (try #2)

thedwitter:
I think I corrupted my first attempt at this pull request so I'm trying again.

This is my first PR so please be gentle!! My company uses Snowflake internally and I'm excited to contribute!

I took a look at #193 and noticed the root cause seems to be that options.host is typically inferred from options.accounts, and so when options.accounts is inferred from the options.accessUrl it will remain unset and result in a crash starting with 1.6.2. The fix is to also set options.host when inferring the options.account.

I've also added a test case.

~ Matt


sfc-gh-dheyman:
Closing due to issue expiration. Feel free to reopen if necessary.

PR #254, opened 2022-01-20
SNOW-499170 Bringing GH actions up-to-date with Python connector's

sfc-gh-mkeller:
Committed via https://github.com/asottile/all-repos


sfc-gh-kwagner (approved):
LGTM

PR #279, opened 2022-03-21
[Connection pool] skip heartbeat check in keep-alive mode

datbth:
## Summary
In connection pool, skip heartbeat check when keep-alive is enabled.

## Description
* Before:
    * Every `acquire()` call on the connection pool runs a heartbeat query (given that `testOnBorrow` of the pool is kept as `true`)
        * One big purpose of a connection pool is to avoid the overhead of the initial connection step. However, this heartbeat query execution results in an equivalent or even higher overhead, especially when the heartbeat query fails and the pool has to make a new connection.
 * After:
    * When `clientSessionKeepAlive` is true, skip the heartbeat query check and check only the client-side connection status.

## Other changes
* Refactor `heartbeatAsync` into `checkHeartbeat`:
    * When the heartbeat query fails, resolve the returned Promise into `false` instead of raising an Error.
    * Reuse the function `heartbeat` to reduce code duplication and ensure the consistency of "heartbeat" queries.
 * Check client-side connection status first (i.e. short-circuiting the logic) to avoid unnecessary heartbeat query executions.

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -98,38 +98,43 @@ function Connection(context)
     return id;
   };
 
-  this.heartbeat = function (self)
+  /**
+  * Runs a "select 1" query.
+  *
+  * @returns {undefined}
+  */
+  this.heartbeat = function (self, complete = function () {})
   {
     self.execute({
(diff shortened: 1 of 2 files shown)


datbth:

PR #284, opened 2022-04-04
fix web auth response

sfc-gh-tkojima:
Using http.createServer instead of net.createServer is more idiomatic and allows us to not have to worry about http specifics.

In addition the response being sent was never terminated, thus the browser would think the response was still pending.

image


sfc-gh-tkojima:

PR #288, opened 2022-04-10
Move aws-sdk to peerDependencies

### LINKED ISSUE #216: Move aws-sdk to peer/optional dependency
Currently the aws-sdk is specified as a dependency for a single util function related to S3. The aws-sdk adds over 90mb of uncompressed code. Moving to a peer/optional dependency would allow for a much smaller bundle size on platforms like aws-lambda.

Another option is to move to v3 of the aws-sdk which allows for installing just the s3 SDK.

markandrus:
aws-sdk is now a peer dependency. Users of this library's S3 functionality will now have to install a compatible version of aws-sdk alongside it. As such, you could argue this is a breaking change.

I haven't tried running the tests yet, and so it may be that aws-sdk should also be listed as a dev dependency, if it is used there.

Fixes #216.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,12 +2,14 @@
   "name": "snowflake-sdk",
   "version": "1.6.8",
   "description": "Node.js driver for Snowflake",
+  "peerDependencies": {
+    "aws-sdk": "^2.878.0"
+  },
   "dependencies": {
     "@azure/storage-blob": "^12.5.0",
     "agent-base": "^6.0.2",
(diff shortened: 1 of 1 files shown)


markandrus:

sfc-gh-dszmolka:
would possibly also fix https://github.com/snowflakedb/snowflake-connector-nodejs/issues/449

markandrus:
This PR will need some code changes, too. Not working yet, IIRC.

PR #290, opened 2022-04-12
Update 'async' package to 3.2.3

SimbaGithub:
Based on SF# 00379040

### DIFF
--- a/ci/container/package.json
+++ b/ci/container/package.json
@@ -6,7 +6,9 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1"
   },
-  "dependencies": {},
+  "dependencies": {
+    "async": "3.2.2"
+  },
   "devDependencies": {
     "async": "3.2.2",
     "insert-module-globals": "^7.2.0",
--- a/ci/container/package.json
+++ b/ci/container/package.json
@@ -16,6 +16,12 @@
     "vinyl-buffer": "^1.0.1",
     "vinyl-source-stream": "^1.1.2"
   },
+  "resolutions": {
+    "async": "3.2.2"
+  },
+  "overrides": {
+    "async": "3.2.2"
+  },
   "author": "Snowflake, Inc.",
   "license": "ISC"
 }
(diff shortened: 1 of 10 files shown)


sfc-gh-kwagner (dismissed):

package.json
```diff
@@ -37,7 +37,7 @@
     "winston": "^3.1.0"
   },
   "devDependencies": {
-    "async": "^2.6.1",
+    "async": "^3.2.3",
```
sfc-gh-kwagner, line 40: You also need to update this in the regular dependencies section since `ocsp` relies on `async` and the `ocsp` package is pulling a bad version of `async` which is causing the build failure.

sfc-gh-kwagner:
@SimbaGithub could you update this PR and get it merged ASAP? This issue is causing some of our testing to completely fail

ci/container/package.json
```diff
@@ -7,15 +7,21 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
+    "async": "3.2.2"
   },
   "devDependencies": {
-    "async": "^2.6.1",
+    "async": "3.2.2",
```
sfc-gh-kwagner, line 13: Should this be greater than 3.2.2 instead of equal to 3.2.2?

SimbaGithub: Setting to 3.2.3 had the same result

sfc-gh-kwagner: No like set it to `"async": "^3.2.2",`

SimbaGithub: Using `"async": "^3.2.2",` resulted in the same failure.

Replacing the `ocsp` package with a forked version that's using a later version of `async` seems to have fixed it

sfc-gh-kwagner:
@SimbaGithub you need to sign the CLA. Also going to add @sfc-gh-hchaturvedi since we're changing the OCSP module

SimbaGithub:

sfc-gh-kwagner (approved):

PR #326, opened 2022-07-11
Rename to 'README.md' for npm project description

SimbaGithub:
Based on SF# 00388805


sfc-gh-kdama:
File name seems to be readme.rd is that ok @SimbaGithub ?

SimbaGithub:
> File name seems to be readme.rd is that ok @SimbaGithub ?

You're right, I'll rename it to md

PR #328, opened 2022-07-13
adding arraybinding for nodejs

SimbaGithub:


lib/connection/bind_uploader.js
```diff
@@ -0,0 +1,150 @@
...
+	
+		var dataRows = new Array();
+		for(var i=0; i<bindings.length; i++)
+		{
+			var strbuffer = "";
+			for(var j=0; j< bindings[i].length; j++)
+			{
+				if (j>0)
+					strbuffer += ',';
+				var value = bindings[i][j];
```
sfc-gh-hchaturvedi, line 63: Isn't there possibility of special characters in here that might need to be escaped?

lib/connection/bind_uploader.js
```diff
@@ -0,0 +1,150 @@
...
+	
+		var startIndex = 0;
+		var rowNum = 0;
+		var curBytes = 0;
+		var fileCount = 0;
+		
+	
+		while (rowNum < dataRows.length)
+		{
+			Logger.getInstance().debug('rowNum='+rowNum+', dataRows='+dataRows.length);
```
sfc-gh-hchaturvedi, line 78: Can this entire loop not be done as part of the previous for loop where we're creating the CSV? Can the strbuffer created in the previous loop be ingested into data directly which is being done below?

lib/connection/bind_uploader.js
SimbaGithub: I tested with " and ', which work fine. User should escape it during input.

sfc-gh-hchaturvedi: JDBC Driver escapes it. It would be great if we can remain consistent.

SimbaGithub: nodejs took bind as an array, so it should handle the escape when user input to the array.

sfc-gh-hchaturvedi (approved):
LGTM. We can let the release move forward but please add a test for special characters.

lib/connection/bind_uploader.js
SimbaGithub: added cvs escape

test/integration/testArrayBind.js
```diff
@@ -0,0 +1,56 @@
...
+		arrBind.push(['string'+i, i, "2020-05-11", "12:35:41.3333333", "2022-04-01 23:59:59", "2022-07-08 12:05:30.9999999"]);
+	}
+	
+	var insertStatement = connection.execute({
+		sqlText: insertWithQmark,
+		binds: arrBind,
+		complete: function (err, stmt) {
+			testUtil.checkError(err);
+			assert.strictEqual(stmt.getNumUpdatedRows(), count);
+			done();
```
sfc-gh-igarish, line 52: Can you try the test with removing this "done()"? This must be called in after() also.

SimbaGithub (commented):
merge master to snow-373

SimbaGithub:
For this PR, The user will not see any different, but it will increase the performance when they do bulk insert, the driver will convert that bulk insert data into different files and upload to the server, instead of sending everything at once.

PR #337, opened 2022-08-18
Use v3 AWS SDK for S3 remote client

SimbaGithub:
Based off SF# 00394578 

- Use the modular version of the AWS SDK to reduce size
- Check for decoded key length when encrypting/decrypting to determine which AES algorithm to use
- The v3 AWS SDK requires a node version >= 12. One of the dockers used centos 6 is not compatible with later versions of node. The latest versions of ubuntu is used instead for testing.


sfc-gh-igarish (approved):
Looks good to me.

sfc-gh-kwagner (approved):

.github/workflows/build-test.yml
```diff
@@ -44,10 +44,15 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                image: [ 'nodejs-centos6-default', 'nodejs-centos7-node12', 'nodejs-centos7-fips' ]
                 cloud: [ 'AWS', 'AZURE', 'GCP' ]
```
sfc-gh-igarish, line 47: why we change this?

.github/workflows/build-test.yml
```diff
@@ -57,8 +62,7 @@ jobs:
               env:
                 PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
-                TARGET_DOCKER_TEST_IMAGE: ${{ matrix.image }}
-              run: ./ci/test.sh
+              run: ./ci/test_mac.sh
```
sfc-gh-igarish, line 65: why test_mac here?

ci/_init.sh
```diff
@@ -25,10 +25,10 @@ declare -A BUILD_IMAGE_NAMES=(
 )
 export BUILD_IMAGE_NAMES
 
-declare -A TEST_IMAGE_NAMES=(
-    [$DRIVER_NAME-centos6-default]=$DOCKER_REGISTRY_NAME/client-$DRIVER_NAME-centos6-default-test:$BUILD_IMAGE_VERSION
-    [$DRIVER_NAME-centos7-node12]=$DOCKER_REGISTRY_NAME/client-$DRIVER_NAME-centos7-node12-test:$BUILD_IMAGE_VERSION
-    [$DRIVER_NAME-centos7-fips]=$DOCKER_REGISTRY_NAME/client-$DRIVER_NAME-centos7-fips-test:$BUILD_IMAGE_VERSION
-)
-export TEST_IMAGE_NAMES
+#declare -A TEST_IMAGE_NAMES=(
```
sfc-gh-igarish, line 28: again why we remove some block and comment a new block?

ci/image/build.sh
```diff
@@ -14,10 +14,10 @@ for name in "${!BUILD_IMAGE_NAMES[@]}"; do
...
 done
 
-for name in "${!TEST_IMAGE_NAMES[@]}"; do
-    docker build \
-        --file $THIS_DIR/Dockerfile.$name-test \
-        --label snowflake \
-        --label $DRIVER_NAME \
-        --tag ${TEST_IMAGE_NAMES[$name]} .
-done
+#for name in "${!TEST_IMAGE_NAMES[@]}"; do
```
sfc-gh-igarish, line 17: Same comment as above.

sfc-gh-igarish (changes requested):
There are more code changes than title of the PR. Please see my comments.

.github/workflows/build-test.yml
SimbaGithub: The v3 AWS SDK requires a node version >= 12. These dockers use centos 6 which is not compatible with later versions of node so it results in test failures. 

When running the tests on centos 6, this is the warning stating the node version is unsupported by the package: https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/2884223384/jobs/4583575600#step:4:232

And this is the resulting error from the test run: https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/2884223384/jobs/4583575600#step:4:403

I tried to update to the latest node version but centos6 is incompatible. So instead of using the centos images, the latest ubuntu build provided by github is now used instead for testing.

ci/_init.sh
SimbaGithub: It's the same block but commented. They are commented out since the centos images are not being run anymore due to being incompatible with the later versions of node.

ci/image/build.sh
SimbaGithub: It's the same block but commented. They are commented out since the centos images are not being run anymore due to being incompatible with the later versions of node.

PR #338, opened 2022-08-23
Allow execute() with request id only

SimbaGithub:
Based off SF# 00393927

The connector requires a sqlText when calling execute. This PR allows passing only a requestId when calling the execute() function


lib/connection/statement.js
```diff
@@ -1113,10 +1117,17 @@ function sendRequestPreExec(statementContext, onResultAvailable)
...
+  {
+    disableOfflineChunks: false,
+  };
+  if (!statementContext.resubmitRequest)
+  {
+    json.sqlText = statementContext.sqlText;
+  }
+  else
+  {
+    json.sqlText = 'select 1;';
```
sfc-gh-igarish, line 1129: why we are hard coding SQL here?

SimbaGithub: If the user only passes a requestId when executing a query, the request fails due to not having a sqlText. The hardcoded SQL is for when the user only passes a requestId.

I've been discussing with Kris if it's possible to not have a sqlText when a requestId is already provided so that adding SQL wouldn't be necessary

sfc-gh-mkeller:
recheck

sfc-gh-igarish (approved):
LGTM. If tests pass, merge it.

SimbaGithub:
Failing tests are similar to the release PR https://github.com/snowflakedb/snowflake-connector-nodejs/pull/344 test failures
These failures also showed up in other PRs like https://github.com/snowflakedb/snowflake-connector-nodejs/pull/346

Merging PR as the test failures are not related to the changes

PR #356, opened 2022-11-02
Add no_proxy functionality

sfc-gh-ext-simba-lb:
Simba ticket 00406584.
This is a parity feature request to support NOPROXY/NO_PROXY as per JDBC/ODBC drivers.

Add noProxy connection parameter to specify a list of hosts that the driver should bypass the proxy and connect to directly. Hosts are separated with '|' symbol and '*' can be used as a wildcard.


sfc-gh-igarish (approved):
LGTM

Issue #355, opened 2022-10-28
Login/Role retry behaviour broken for connection pools


shhac:
Please answer these questions before submitting your issue. Thanks!

1. What version of NodeJS are you using (`node --version` and `npm --version`)?

`v16.17.1`
`8.15.0`

2. What operating system and processor architecture are you using?

MacOS `arm64` (dev machine), Ubuntu `x86_64` (deployment)

3. What are the component versions in the environment (`npm list`)?

`"snowflake-sdk": "^1.6.14",`

5. What did you do?

Trying to create a connection pool with an invalid credential or an invalid role causes an infinite loop of immediate retries; resulting in the whole account get locked

```ts
  const pool = snowflake.createPool(
    {
      account,
      password,
      role,
      username,
      warehouse,
      application: 'foobar',
      timeout: 30_000,
    },
    {
      evictionRunIntervalMillis: 30_000,
      idleTimeoutMillis: 30_000,
      max: poolSize,
      min: 0,
    },
  );
```

with e.g. `password: 'wrong password'` or `role: 'wrong role'`

Same behaviour happens when providing options `sfRetryMaxLoginRetries` and `sfRetryMaxNumRetries` which look like they should control this

```ts
    {
      evictionRunIntervalMillis: 30_000,
      idleTimeoutMillis: 30_000,
      max: poolSize,
      min: 0,
      ...({ sfRetryMaxLoginRetries: 0, sfRetryMaxNumRetries: 0 } as any),
  },
```

Also note that TypeScript does not recognise these as valid items on the options so to even try this we need to specify `as any`

6. What did you expect to see?

Single fail. No retries. Account not locked. `Ctrl` `C` shuts down app cleanly

8. What did you see instead?

Infinite retry loop. `Ctrl` `C` does not prevent additional errors being logged.

In LOGS:

With wrong password

```
Unable to connect: User temporarily locked. Contact your local system administrator or please create a case with Snowflake Support or reach us on our support line: 
```

or, with wrong role

```
Unable to connect: Role 'wrong role' specified in the connect string is not granted to this user. Contact your local system administrator, or attempt to login with another role, e.g. PUBLIC.
```

sfc-gh-dszmolka:
hi @shhac the fix for this issue should be available since 1.6.17 of the connector. if it indeed resolves the issue for you, could you please close this issue?

sfc-gh-dszmolka:
resolved by above PR, thank you for raising this issue !

PR #360, opened 2022-11-15
Fix connection pool login retry with incorrect pwd/role

sfc-gh-ext-simba-lb:
Fix for issue raised in #355 .
If using a connection pool and the wrong password/role is specified, the connector gets stuck in an infinite loop instead of returning an error.

The issue is reported on the node-pool module github page: https://github.com/coopernurse/node-pool/issues/273.
And a workaround here https://github.com/coopernurse/node-pool/pull/221.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -298,6 +298,11 @@ function Core(options)
       poolOptions
     );
 
+    // avoid infinite loop if factory creation fails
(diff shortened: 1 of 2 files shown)


sfc-gh-igarish (approved):
LGTM

PR #370, opened 2022-12-02
fix array binding master key issue

SimbaGithub:
move cleanTempFiles to steam on data to check the status.


sfc-gh-dszmolka:
another issue reported #374

sfc-gh-dszmolka:
adding @sfc-gh-igarish as a reviewer here

lib/connection/bind_uploader.js
```diff
@@ -29,7 +29,7 @@ const CREATE_STAGE_STMT = "CREATE OR REPLACE TEMPORARY STAGE "
  
 function BindUploader(options, services, connectionConfig, requestId)
 {
-	const MAX_BUFFER_SIZE = 1024 * 1024 * 100;
+	const MAX_BUFFER_SIZE = 1024 * 1024 * 1024 * 10;
```
sfc-gh-igarish, line 32: You are increasing from 100 MB to 10 GB. Did you measure memory consumption?

sfc-gh-igarish:
How about adding a test case?

sfc-gh-dszmolka:
array binding master key issue is fixed with #419

PR #380, opened 2022-12-12
fix missing arrayBindingThreshold on connection string

SimbaGithub:
fix missing arrayBindingThreshold on connection string


sfc-gh-igarish (approved):
LGTM

PR #384, opened 2022-12-15
Issue 66 array binding master key issue

SimbaGithub:
Issue 66


sfc-gh-igarish:
Why changes are in Jenkinfiles? Is the fileset only for this bug fix or you get other changes?

PR #385, opened 2022-12-16
Check OCSP response cache is initialised before setting cache entry

### LINKED ISSUE #349: TypeError: Cannot read properties of undefined (reading 'set') at OcspResponseCache.set
I am using node version v17.9.0. 
OS is Linux. 
I am using Kubernetes and launching a nodejs application to connect with snowflake and execute a query. 

version of snowflake-promise is 2.2.0.
version of snowflake-sdk is 1.6.14.

I am getting below error. Can you please suggest.

```
/app/node_modules/snowflake-sdk/lib/agent/ocsp_response_cache.js:157
    cache.set(certId, response);
          ^
TypeError: Cannot read properties of undefined (reading 'set')
    at OcspResponseCache.set (/app/node_modules/snowflake-sdk/lib/agent/ocsp_response_cache.js:157:11)
    at /app/node_modules/snowflake-sdk/lib/agent/socket_util.js:232:32
    at done (/app/node_modules/snowflake-sdk/lib/agent/check.js:142:5)
    at ocspResponseVerify (/app/node_modules/snowflake-sdk/lib/agent/check.js:201:7)
    at done (/app/node_modules/snowflake-sdk/lib/agent/check.js:71:7)
    at IncomingMessage.<anonymous> (/app/node_modules/snowflake-sdk/lib/agent/check.js:99:7)
    at IncomingMessage.emit (node:events:539:35)
    at IncomingMessage.emit (node:domain:475:12)
    at endReadableNT (node:internal/streams/readable:1345:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
```

sfc-gh-ext-simba-lb:
To fix #349 

```
/app/node_modules/snowflake-sdk/lib/agent/ocsp_response_cache.js:157
    cache.set(certId, response);
          ^
TypeError: Cannot read properties of undefined (reading 'set')
    at OcspResponseCache.set (/app/node_modules/snowflake-sdk/lib/agent/ocsp_response_cache.js:157:11)
    at /app/node_modules/snowflake-sdk/lib/agent/socket_util.js:232:32
    at done (/app/node_modules/snowflake-sdk/lib/agent/check.js:142:5)
    at ocspResponseVerify (/app/node_modules/snowflake-sdk/lib/agent/check.js:201:7)
    at done (/app/node_modules/snowflake-sdk/lib/agent/check.js:71:7)
    at IncomingMessage.<anonymous> (/app/node_modules/snowflake-sdk/lib/agent/check.js:99:7)
    at IncomingMessage.emit (node:events:539:35)
    at IncomingMessage.emit (node:domain:475:12)
    at endReadableNT (node:internal/streams/readable:1345:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
```

If getOcspResponseCache().set(cert, data.res) is called when the cache hasn't been initialized it will throw the undefined error. Check that the cache is initialized first before calling set. If it's undefined, initialize the cache with size 1 and set the entry.


lib/agent/ocsp_response_cache.js
```diff
@@ -166,8 +173,15 @@ function OcspResponseCache()
...
-    const certId = CertUtil.buildCertId(cert);
-    return cache.get(certId);
+    try
+    {
+      const certId = CertUtil.buildCertId(cert);
+      return cache.get(certId);
+    }
+    catch (e)
+    {
+      Logger.getInstance().debug("Failed to get certificate %s. err: %s", certId, e);
```
sfc-gh-igarish, line 183: Don't we have to return null or something after caching the exception?

sfc-gh-igarish (commented):
Do we have any test case? If not, can we add it?

sfc-gh-ext-simba-lb:
@sfc-gh-igarish I am not able to reproduce the issue even when I introduce a network delay. I added a unit test that tests attempting to set the certificate in the cache when the cache hasn't been initialized, which is the cause of the issue. This no longer breaks but handles the error gracefully.

There was a check isInitialized() added in #153, I added this check to where the set function is originally called in SocketUtil.js. This should prevent the issue from happening in the first place.

sfc-gh-igarish:
Please check with Harsh. We need a test case or double check with Harsh about OCSP changes. @sfc-gh-hchaturvedi

sfc-gh-ext-simba-lb:
@sfc-gh-igarish I am able to reproduce the issue now and was able to confirm the fix. 
I am not sure how to add a test because to reproduce we need to run 'sudo tc qdisc add dev eth0 root netem delay 5000ms' on Linux to create an artificial network delay.
adding @sfc-gh-hchaturvedi as a reviewer.

sfc-gh-igarish:
Can you run system command from node.js? https://stackoverflow.com/questions/53057672/which-are-the-inbuilt-modules-responsible-for-system-calls-in-node-js. If yes, will it work to repro the use case?

sfc-gh-ext-simba-lb:
Are we able to find out which network interface is used in Linux tests? For example I needed to change eth0 to ens33 in my vm to run the tc command. Otherwise we might need to make this a manual test.

sfc-gh-sfaber:
@sfc-gh-hchaturvedi pinging again, can you please take a look? Thanks

lib/agent/socket_util.js
```diff
@@ -229,7 +229,11 @@ function validateCertChain(cert, cb)
       // if we have an ocsp response, cache it
       if (data && (!data.err || isValidOCSPError(data.err)))
       {
-        getOcspResponseCache().set(cert, data.res);
+        // check if cache is initialized before setting entry
+        if (getOcspResponseCache().isInitialized())
```
sfc-gh-hchaturvedi, line 233: 1. If the cache hasn't been initialized we should initialize it as well and then set the cert. 
2. If we've already put this if condition do we still need the try catch block? I don't mind putting it just that I want to understand what is the actual cause of the exception so that we address that.

sfc-gh-ext-simba-lb: The cause of this issue is that the driver is timing out trying to reach ocsp.snowflakecomputing.com which prevents the cache from being initialized. Do you still want to try initialize again here incase of a one off failure? We don't need the try catch block for this specific error if we have the condition. I kept it incase some other error happened.

sfc-gh-hchaturvedi: Yes we should initialize the cache nevertheless so the driver does not have to continuously reach out to the OCSP Responders again n again. I don't mind the try catch block but I'd recommend to add an ability to put the driver in test mode so that in test mode the driver does not suppress exceptions. But this is a general advise.

sfc-gh-hchaturvedi (approved):
LGTM.

PR #398, opened 2023-01-13
force smkid to string

sfc-gh-ext-simba-dl:
for issue 66, array binding master key issue.
increased the upload file size (1024 * 1024 * 1024 * 10) to reduce the total number of files which send to the server. The required memory is almost the same, we write all those binding data to the files and upload it. After uploaded those files, the driver will delete it from the memory. 
force smkid to string


sfc-gh-dszmolka:
@sfc-gh-igarish can you please help us here, review the PR so it could get merged ? (smkId gets mangled when it's a large integer) thank you in advance !

sfc-gh-igarish (approved):
LGTM. If this fixes the customer issue, please comment on Simba ticket for the same.

PR #418, opened 2023-02-10
Add support for GCS access token for PUT/GET

sfc-gh-ext-simba-lf:
Regarding issue 93

Support using GCS access token for PUT/GET queries. 

Added test cases based on ODBC and JDBC connector similar to the [.NET PR](https://github.com/snowflakedb/snowflake-connector-net/pull/585#issuecomment-1416645309):
- The first test case based on ODBC is about executing a large query with the GCS downscoped token: https://github.com/snowflakedb/snowflake-odbc/pull/409/files#diff-9c59c382829b5ea28ce41ba69e6a64bd9e13bd29a1197f2c996710a68fc87976R2576
- The second test case based on JDBC uploads and downloads a file with the GCS downscoped token: https://github.com/snowflakedb/snowflake-jdbc/pull/409/files#diff-65666f89fd973572d96644d93d201eeb917372a9cbab1519ab487a9561c5a063R753


lib/services/sf.js
```diff
@@ -1121,6 +1121,12 @@ StateConnecting.prototype.continue = function ()
       this.connectionConfig.getJsTreatIntegerAsBigInt();
   }
 
+  if (Util.exists(this.connectionConfig.getGcsUseDownscopedCredential()))
```
sfc-gh-igarish, line 1124: I can see token pass, but where is the code to use GCS SDK for get/put instead of HTTPS call?

sfc-gh-ext-simba-lf: Currently, the connector doesn't use the GCS SDK so it makes HTTPS calls with the GCS access token in the authorization header: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/ea8db10873f1c38b3aeed84fa7b418af4a7730b4/lib/file_transfer_agent/gcs_util.js#L232

Similar to the python connector: https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/gcs_storage_client.py#L156

sfc-gh-ext-simba-lf: I'll update the PR to use the GCS SDK instead when the access token is available

sfc-gh-igarish: Please follow the JDBC code changes. As Python somehow going away from SDK.

sfc-gh-igarish: If you want to follow Python and not to add any dependency then that's fine also. As Python, Node.js classify as connectors rather than drivers.

sfc-gh-igarish (approved):
LGTM

sfc-gh-igarish (approved):
LGTM

PR #431, opened 2023-03-01
apply proxy settings for ocsp validation

sfc-gh-ext-simba-hx:
Fix for issue: NodeJS driver is not sending ocsp requests through proxies
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/57

The root cause is proxy agent is not used when connect to ocsp server.
To fix the issue, pass in proxy agent for ocsp validation.

No test case added since proxy is not available on github test environment. There is no way to set global proxy setting in nodejs so the trick we used for ODBC, having invalid proxy settings in environment variables, doesn't work either. Confirmed with local test through log (added in HttpsProxyAgent.connect() ) that proxy is used for ocsp validation and connection can be made.

### DIFF
--- a/lib/agent/https_ocsp_agent.js
+++ b/lib/agent/https_ocsp_agent.js
@@ -45,7 +45,7 @@ function HttpsOcspAgent(options)
     var socket = HttpsAgent.prototype.createConnection.apply(this, arguments);
 
     // secure the socket and return it
-    return SocketUtil.secureSocket(socket, host);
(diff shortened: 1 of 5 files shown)


sfc-gh-igarish (approved):
LGTM. Let's Harsh also review it.

sfc-gh-hchaturvedi (approved):
LGTM

PR #434, opened 2023-03-03
fix getSqlText() undefined on multi statement for issue 187

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/187
not return getSqlText() on multi statement


lib/connection/statement.js
```diff
@@ -688,6 +691,8 @@ function BaseStatement(
           Util.format('request id = %s', context.requestId) :
           Util.format('statement id = %s', context.statementId));
     }
+    this.hasNext = hasNextResult(this, context);
+    this.NextResult = createNextReuslt(this, context);
```
sfc-gh-igarish, line 695: Typo

lib/connection/statement.js
```diff
@@ -824,6 +832,7 @@ function createOnStatementRequestSuccRow(statement, context)
         context.isMulti = true;
         context.multiResultIds = this._resultIds;
         context.multiCurId = 0;
+        context.arrSqlText = context.sqlText.split(';');
```
sfc-gh-igarish, line 835: will this work in all cases? do we support store proc creation as multi statement? if yes,  one store proc definition can contains many semicolons. Also how about a comment contains a semi colon?

sfc-gh-ext-simba-dl: should we always return the whole multistatements for each sub statement?

sfc-gh-ext-simba-dl: I added code cover the store proc

sfc-gh-dszmolka:
fixed on another PR

PR #436, opened 2023-03-07
fix issue 184, incorrectly inserts double-quotes

sfc-gh-ext-simba-dl:


lib/connection/bind_uploader.js
```diff
@@ -138,10 +138,12 @@ function CleanFile(fileName)
 {
 	try
 	{
+		/*
```
sfc-gh-dszmolka, line 141: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/v1.6.19/lib/connection/statement.js#L174-L181 seems to call this method `CleanFile`. maybe I'm missing something ? why don't we want to delete the temp files after the upload is done?

sfc-gh-igarish (approved):
LGTM

PR #439, opened 2023-03-14
bug fix SNOW-719255: "Bind variable ? not set" error message

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/101

fixed the following case.
var insertStatement = connection.execute({
sqlText: 'insert into testTbl2(cola,colb) select value:stuff1, value:stuff2 from table(flatten(parse_json(?)))`,
binds: [JSON.stringify(arrBind)],
complete: function (err, stmt) {
if (err) {
console.error('1 Failed to execute statement due to the following error: ' + err.message);
}
else {
console.log('inserted rows=' + stmt.getNumUpdatedRows());
}
CloseConnection();
}
});

### DIFF
--- a/lib/connection/statement.js
+++ b/lib/connection/statement.js
@@ -1639,9 +1639,13 @@ function countBinding(binds)
   var count = 0;
   for(var index = 0; index < binds.length; index++)
   {
-    if(binds[index] != null)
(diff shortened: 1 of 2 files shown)


sfc-gh-igarish (approved):
LGTM

PR #442, opened 2023-03-16
Bug fix: refresh ocsp local cache when expired

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/129
The OCSP local cache won't be refreshed if the application run for a long time and the cache is expired during that, so query will fail after that.
The root cause is that the driver only check expiration and refresh the cache when the application start, no refresh logic afterwards.

### DIFF
--- a/lib/agent/ocsp_response_cache.js
+++ b/lib/agent/ocsp_response_cache.js
@@ -35,6 +35,9 @@ const cacheFileName = path.join(cacheDir, "ocsp_response_cache.json");
 var cache;
 // JSON object with previous cache's responses
 var prevCacheObj;
+// Cache updated time, initialized as current time.
+// Will be updated when load from local cache file or refresh by downloading
+var cacheUpdateTime = Date.now() / 1000;
 
 function deleteCache()
 {
@@ -68,6 +71,7 @@ function OcspResponseCache()
    * Reads OCSP cache file.
    */
   const currentTime = Date.now() / 1000;
+  cacheUpdateTime = currentTime;
 
   let OCSP_URL = process.env.SF_OCSP_RESPONSE_CACHE_SERVER_URL;
   if (!OCSP_URL)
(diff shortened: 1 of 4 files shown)


lib/agent/ocsp_response_cache.js
```diff
@@ -136,6 +140,30 @@ function OcspResponseCache()
...
+      return false;
+    }
+
+    const currentTime = Date.now() / 1000;
+
+    if ((currentTime - cacheUpdateTime) > maxAge)
+    {
+      Logger.getInstance().debug(
+        "OCSP local cache validity is out of range. currentTime: %s, timestamp: %s",
+        currentTime, cacheUpdateTime);
```
sfc-gh-igarish, line 160: Log maxAge time

lib/agent/ocsp_response_cache.js
```diff
@@ -136,6 +140,30 @@ function OcspResponseCache()
...
+  this.IsCacheExpired = function ()
+  {
+    if (!cacheInitialized)
+    {
+      return false;
+    }
+
+    const currentTime = Date.now() / 1000;
+
+    if ((currentTime - cacheUpdateTime) > maxAge)
```
sfc-gh-igarish, line 156: Are currentTime, cacheUpdateTime and maxAge in same time unit? I think first two are in milli seconds and maxAge is in seconds. Please double check and put unit in comments for variables.

sfc-gh-ext-simba-hx: Will add comments for unit.
Date.now() returns milliseconds so `currentTime = Date.now() / 1000` will be in seconds.
cacheUpdateTime is initialized with `Date.now() / 1000` as well so unit is seconds. It could be updated from the timestamp in local cache file, which is made here using currentTime.
https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/agent/ocsp_response_cache.js#L155

lib/agent/ocsp_response_cache.js
sfc-gh-ext-simba-hx: Log added

sfc-gh-igarish (approved):
LGTM

PR #443, opened 2023-03-17
Issue101, SNOW-719255: "Bind variable ? not set" error message after upgrade from 1.6.13 to a higher version

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/101
bug fix for stage array binding, handle the bind array case


sfc-gh-igarish (approved):
LGTM

PR #444, opened 2023-03-17
SNOW-762041: Bumped up the NodeJS Driver PATCH version from 1.6.19 to…

sfc-gh-anavalos:
… 1.6.20

@noreview - This is an automated process. No review is required.

PR #446, opened 2023-03-18
Update package.json

sfc-gh-kdama:


ci/container/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nodejs_driver_test",
-  "version": "1.6.18",
+  "version": "1.6.19",
```
sfc-gh-igarish, line 3: If upcoming release is 1.6.20 then it should be 1.6.20.

sfc-gh-dszmolka: looks to be already submitted with #444

PR #447, opened 2023-03-20
update package.json urllib from v2 to v3

mrspeiser:
Updating the urllib package version from 2 to 3.

urllib looks to only be called once inside lib/http/base.js
`request = require('urllib').request(requestOptions.url, requestOptions, async function (err, data, response)`

unit tests ran fine for me, but I don't have any snowflake credentials to run snowflake_test.js.

Will add additional comments if able to do more testing/investigation.

Thanks

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "string-similarity": "^4.0.4",
     "test-console": "^2.0.0",
     "tmp": "^0.2.1",
-    "urllib": "^2.38.0",
+    "urllib": "^3.11.0",
     "uuid": "^3.3.2",
     "winston": "^3.1.0"
   },


sfc-gh-dszmolka:
urllib 3 introduced a [breaking change with their refactor](https://github.com/node-modules/urllib/commit/26c44b75e7a6ac95dd0dd8853400f7d83d815f3b) which fundamentally breaks the operation of the snowflake node.js driver, starting from urllib 3.0.0. 
the simple package bump is not enough, we also need to figure out to refactor our code too or replace urllib (which is also a refactor)

i suggest closing this PR.

mrspeiser:
Thanks for the reply @sfc-gh-dszmolka. I will close this PR.

PR #453, opened 2023-03-22
Fix parsing issue with XML data from VARIANT columns

sfc-gh-ext-simba-lf:
Regarding issue 131

Adds an xml parser so the connector parses raw XML strings correctly

Test cases are added for both XML and JSON data types. The test loads data in an XML/JSON format into a table with a VARIANT column then calls a select query to verify the values are correct


lib/connection/result/column.js
```diff
@@ -550,7 +551,18 @@ function convertRawVariant(rawColumnValue, column, context)
   {
     try
     {
-      ret = eval("(" + rawColumnValue + ")");
+      // check if raw string is in XML format
+      // matches any string with an enclosed '<' and '>' character
+      if ((/^\s*<[\s\S]*>/).test(rawColumnValue))
```
sfc-gh-igarish, line 556: Can we do more accurate way to detect XML?

sfc-gh-ext-simba-lf: It's possible to use XMLValidator to check if the raw string is in valid XML format: https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/4.XMLValidator.md

- XMLValidator returns true if no issue is found.
- XMLValidator returns an error object if any issue is found.

sfc-gh-igarish: My question is more about how to differentiate a string value is an XML or not? Just by checking angle brackets won't be sufficient.

sfc-gh-ext-simba-lf: PR is updated to have a more accurate way of differentiating between an XML string and a non-XML string
It checks if each tag in the string is closed. It also checks if the attributes and element in each tag are valid

sfc-gh-ext-simba-lf: The JSON test case is also updated to include angle brackets in the value to test that it's still correctly detected as JSON

lib/connection/result/column.js
```diff
@@ -550,7 +551,18 @@ function convertRawVariant(rawColumnValue, column, context)
   {
     try
     {
-      ret = eval("(" + rawColumnValue + ")");
+      // check if raw string is in XML format
+      // ensure each tag is enclosed and all attributes and elements are valid
+      if (XMLValidator.validate(rawColumnValue) === true)
```
sfc-gh-igarish, line 556: Please run some perf tests with previous release of the driver and make sure this won't hurt the performance.

sfc-gh-igarish (approved):
LGTM

lib/connection/result/column.js
sfc-gh-ext-simba-lf: The execution time of convertRawVariant() increases by ~1.2 ms compared to previous version

sfc-gh-ext-simba-lf: Would it be better to parse the string using the JSON parser first, if it fails then try parsing it as an XML?

This would keep performance the same for JSON strings and fix the issue for XML strings

sfc-gh-igarish: Sure.

PR #457, opened 2023-03-27
SNOW-727386: bug fix for ocsp cache refresh

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/129
This is further fix for the changes made in [PR#442](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/442)
There was a typo in logging when changing the variable name from cacheUpdateTime to cacheUpdateTimeSec
The test case added in PR#442 didn't work as expected and passed with the bug while the functionality actually did work.
The reason is that the variable maxAgeSec (for the life time of local cache) is initialized when application starts and can't be updated later. So changing environment variable after it's initialized won't work.
Fix the typo, also update maxAgeSec per request to make the test case work.

### DIFF
--- a/lib/agent/ocsp_response_cache.js
+++ b/lib/agent/ocsp_response_cache.js
@@ -25,7 +25,7 @@ const status = {
 // validate input
 const sizeLimit = GlobalConfig.getOcspResponseCacheSizeLimit();
 // ocsp cache max age in second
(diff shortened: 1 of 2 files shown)


sfc-gh-igarish (approved):
LGTM

PR #459, opened 2023-03-29
fix Latency in NodeJS connector issue

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/188


sfc-gh-igarish:
Could you please add comment about how this will fix the latency?

sfc-gh-ext-simba-dl:
this changes won't work, the root cause it urllib didn't pass down requireOCSP to https and tls, which cause the agent createconnection never get call.

PR #461, opened 2023-03-30
fix: initialise maskedtxt variable prior to trying to mask secrets

bibli-alex:
Fixes a failure that occurs when logging is turned on. 

Without this, `maskedtxt` is never defined, so you get a `ReferenceError: maskedtxt is not defined`.

### DIFF
--- a/lib/secret_detector.js
+++ b/lib/secret_detector.js
@@ -138,6 +138,7 @@ function secret_detector(customPatterns, mock)
     }
 
     var masked = false;
+    var maskedtxt = '';
     var errstr = null;
     try
     {


steamwings:
@bibli-alex Was just about to open the same PR, thank you! Looks like you need to sign the CLA though.

sfc-gh-igarish (approved):
LGTM

sfc-gh-dszmolka:
thank you for submitting this pr @bibli-alex ! is there any specialty for causing the issue to happen? by simply adding
`snowflake.configure({ logLevel: "trace" })` to turn on arbitrary level of logging, it doesn't bail out for me. So probably I'm missing something.

Regardless, this is a valid issue and thank you for submitting the PR ! We'll take a look

sfc-gh-dszmolka:
if anyone could post an executable code snippet which leads to the issue, that would be highly appreciated!

sfc-gh-dszmolka:
implemented by #479  - thank you for catching this and the contribution!

PR #465, opened 2023-04-03
Replace eval() for parsing JSON strings

sfc-gh-ext-simba-lf:
Regarding issue 271

The PR replaces eval() which is considered [unsafe](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!) with JSON.parse() which is specifically for parsing and does not execute functions in strings

### DIFF
--- a/lib/connection/result/column.js
+++ b/lib/connection/result/column.js
@@ -550,7 +550,7 @@ function convertRawVariant(rawColumnValue, column, context)
   {
     try
     {
-      ret = eval("(" + rawColumnValue + ")");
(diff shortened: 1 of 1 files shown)


sfc-gh-ext-simba-lf:
The better-eval module creates objects in a VM which causes an assertion fail when assert.deepStrictEqual is used: https://github.com/nodejs/node/issues/44462

The assertions for test involving the eval function are changed so that only the object's values are checked for equality

sfc-gh-igarish (approved):
LGTM

PR #466, opened 2023-04-05
Nodejs183, crash when using full table path on array binding

sfc-gh-ext-simba-dl:
if the database and schema are missing on the session, the array binding will return error "missing database or schema" 
after this fix, the driver will catch any error which it may happen at create stage, write files, and upload files, when error happen, it will use the normal binding to handle the query.


lib/connection/statement.js
```diff
@@ -91,24 +91,30 @@ exports.createStatementPreExec = function (
...
+    }
+    catch(e)
+    {
+      Logger.getInstance().debug('bind upload error, use normal binding');
+      return createRowStatementPreExec(
+        options, context, services, connectionConfig);
+    }
     if(bindData != null)
     {
-      isUsingBindUpload = true;
```
sfc-gh-igarish, line 104: You removed this for some purpose? when this will be true now?

sfc-gh-igarish (approved):
LGTM

PR #467, opened 2023-04-05
Add retry on econnreset and etimedout error codes for row stream

sfc-gh-ext-simba-lf:
Regarding issue 268

Add retry on `ECONNRESET`and `ETIMEDOUT` error codes for row stream similar to large result set


sfc-gh-igarish (commented):
Please check other driver logic for retryable code. why 403 is retryable?

sfc-gh-ext-simba-lf:
> Please check other driver logic for retryable code

Other connectors do not have retry logic for socket errors like `etimedout` and `econnreset`.
Currently, the node connector is already retrying `econnreset` and `etimedout` for large result sets: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0c7eea1c3056c54f1a1a7b18b57a683393b8cb5b/lib/services/large_result_set.js#L29

> why 403 is retryable?

This comment explains why the connector retries on 403: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/0c7eea1c3056c54f1a1a7b18b57a683393b8cb5b/lib/services/large_result_set.js#L26

@sfc-gh-dszmolka Please feel free to share any more info that you know of regarding the question above

sfc-gh-dszmolka:
i personally have no additional information on why originally it has been decided to retry HTTP403 responses, i understand it might look strange. Its there in the code for at least 5 year when it was originally committed to Github. If needed, perhaps we should raise a separate issue/PR to address that separate question, to research if HTTP403 needs to be still retried after 5 years for large result set.

sfc-gh-igarish:
> 

Other driver do retry on 403 only for GCS. So it passes flag to check do driver wants to retry for 403 or not. Only GCS provider code set it true. This happens historically because GCS was throwing that error because of presigned URL intermittent issue.

sfc-gh-igarish (approved):
LGTM

PR #469, opened 2023-04-13
just for debug

sfc-gh-ext-simba-dl:


sfc-gh-dszmolka:
doesn't seem relevant now

PR #477, opened 2023-04-21
Display a more accurate error message when connection.execute() with requestId fails

sfc-gh-ext-simba-lf:
Regarding issue 137

The PR replaces the placeholder query that the connector uses when retrieving query results based on request id. If the connector is unable to retrieve the query results successfully, the new error message will be displayed instead


sfc-gh-igarish (approved):
LGTM

PR #478, opened 2023-04-21
Fix the error thrown when connect() is called using an okta/external browser authenticator

sfc-gh-ext-simba-lf:
Regarding issue 85

The connector currently throws an incorrect user/password error when connect() is called with an okta or external browser authenticator even when the credentials given are correct. With the PR, the connector will now throw an accurate error and will suggests to use connectAsync() instead.


sfc-gh-igarish (approved):
LGTM

Issue #301, opened 2022-05-09
SNOW-587447: Wrong parameter used as sleep timer cap?


timfenten:
https://github.com/snowflakedb/snowflake-connector-nodejs/blob/182b40f92af2ae6c129b1ded946aba74faa8212c/lib/services/sf.js#L1129

Hi,

its a very specific question but anyway ;-)

Why is PARAM_RETRY_SF_MAX_NUM_RETRIES used as timeout cap here?
Shouldn't it be PARAM_RETRY_SF_MAX_SLEEP_TIME?

sfc-gh-dszmolka:
hi, reviewing the issues stumbled on this one. and to be honest, i'm not sure why :) it's a bit suspicious for sure, but i see it's the same at least for 3.5 years now since https://github.com/snowflakedb/snowflake-connector-nodejs/commit/5ddd0df706006658b548b21f64121003ecffd512

that doesn't mean it's correct, but, wasn't changed for a while now (since implementation, as i see it)
anyways we'll check it and see if a change is needed.

sfc-gh-dszmolka:
fixed with #480 , which is already merged into main and thus will be part of the next release - expected around the second half of May

sfc-gh-dszmolka:
fix released with version [1.6.22](https://github.com/snowflakedb/snowflake-connector-nodejs/releases/tag/v1.6.22)

PR #480, opened 2023-04-25
bug fix issue 139, Wrong parameter used as sleep timer cap

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/139

used incorrect parameter for cap.


sfc-gh-dszmolka:
#301

lib/services/sf.js
```diff
@@ -1136,7 +1136,8 @@ StateConnecting.prototype.continue = function ()
     Date.now() : 'FIXEDTIMESTAMP';
   const maxLoginRetries = connectionConfig.getRetrySfMaxLoginRetries();
   let sleep = connectionConfig.getRetrySfStartingSleepTime();
-  const cap = connectionConfig.getRetrySfMaxNumRetries();
+  const cap = connectionConfig.getRetrySfMaxSleepTime();
```
sfc-gh-igarish, line 1139: is it possible sleeptime as 0? Do we have test already exist for sleep time? are we still using max num of retries some where? Make sure "cap" variable meaning change shouldn't cause any side effect.

sfc-gh-ext-simba-dl: we don't have sleeptime = 0. we still use the max num of retries. for example, retry backoff is 1s, 2s, 4s, 8s, 16s, so this is max retry = 5, if max retry = 6, the backoff will looks like 1s, 2s, 4s, 8s, 16s, 16s. Just the max backoff is 16s. And this setting we didn't allow customer modify it.

sfc-gh-igarish (approved):
LGTM

PR #482, opened 2023-04-26
Add missing dependencies 'bn.js' and 'https-proxy-agent'

sfc-gh-ext-simba-lf:
Regarding issue 132

Add missing dependencies `bn.js` and `https-proxy-agent`


sfc-gh-igarish (approved):
LGTM

PR #492, opened 2023-05-16
fix issue 217, and 279, fix performance issue on stag binding and not return statement issue.

sfc-gh-ext-simba-dl:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/217
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/279

for performance issue, using file stream for encode and upload to the server instead of writing to file and encode on file then send to server. This fix also includes returning statement object after connection execute. Added test case for statement object return and statement cancel.


sfc-gh-igarish (approved):
LGTM

PR #495, opened 2023-05-16
Use the session token when closing the connection

sfc-gh-ext-simba-lf:
Regarding issue 358

When multiple connections are opened using the OAuth authenticator, closing one of those connections lead to the other connections being terminated as well

The PR closes the connection using the session token instead of the master token so that all OAuth connections will not be terminated if one of the connections are closed

### DIFF
--- a/lib/services/sf.js
+++ b/lib/services/sf.js
@@ -1378,11 +1378,14 @@ StateConnected.prototype.request = function (options)
  */
 StateConnected.prototype.destroy = function (options)
 {
-  // send out a master token request to terminate the current connection
-  this.createMasterTokenRequest(
+  var requestID = uuidv4();
+  var requestGUID = uuidv4();
+
+  // send out a session token request to terminate the current connection
+  this.createSessionTokenRequest(
     {
       method: 'POST',
-      url: '/session/logout-request',
+      url: `/session?delete=true&requestId=${requestID}&request_guid=${requestGUID}`,
       scope: this,
       callback: function (err, body)
       {
--- a/lib/services/sf.js
+++ b/lib/services/sf.js
@@ -1379,13 +1379,12 @@ StateConnected.prototype.request = function (options)
 StateConnected.prototype.destroy = function (options)
 {
   var requestID = uuidv4();
-  var requestGUID = uuidv4();
 
   // send out a session token request to terminate the current connection
   this.createSessionTokenRequest(
     {
       method: 'POST',
-      url: `/session?delete=true&requestId=${requestID}&request_guid=${requestGUID}`,
+      url: `/session?delete=true&requestId=${requestID}`,
       scope: this,
       callback: function (err, body)
       {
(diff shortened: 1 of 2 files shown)


lib/services/sf.js
```diff
@@ -1378,11 +1378,14 @@ StateConnected.prototype.request = function (options)
...
-  this.createMasterTokenRequest(
+  var requestID = uuidv4();
+  var requestGUID = uuidv4();
+
+  // send out a session token request to terminate the current connection
+  this.createSessionTokenRequest(
     {
       method: 'POST',
-      url: '/session/logout-request',
+      url: `/session?delete=true&requestId=${requestID}&request_guid=${requestGUID}`,
```
sfc-gh-wshangguan, line 1388: Where do we replace the master token with session token? the `createSessionTokenRequest` function would do that for us?

lib/services/sf.js
```diff
@@ -1378,11 +1378,14 @@ StateConnected.prototype.request = function (options)
  */
 StateConnected.prototype.destroy = function (options)
 {
-  // send out a master token request to terminate the current connection
-  this.createMasterTokenRequest(
+  var requestID = uuidv4();
+  var requestGUID = uuidv4();
```
sfc-gh-wshangguan, line 1382: What is the requestGUID? I didn't see it in JDBC code.

test/unit/mock/mock_http_client.js
```diff
@@ -40,8 +40,17 @@ MockHttpClient.prototype.request = function (request)
       buildRequestToOutputMap(buildRequestOutputMappings(this._clientInfo));
   }
 
+  // Closing a session includes a unique uuid for requestID and requestGUID as query parameters in the url
```
sfc-gh-wshangguan, line 43: Where do we draw the conclusion of `requestGUID`?

sfc-gh-wshangguan (commented):
Have we run any end to end tests to make sure the change works?

sfc-gh-ext-simba-lf:
> Have we run any end to end tests to make sure the change works?

Yes, I ran the repro code with this change and the issue was fixed. I tested with the other authenticator types and they executed successfully as well

lib/services/sf.js
sfc-gh-ext-simba-lf: Yes, the createSessionTokenRequest() function will use the session token instead of the master token for the Snowflake Token header value

lib/services/sf.js
sfc-gh-ext-simba-lf: It's from the .NET connector: https://github.com/snowflakedb/snowflake-connector-net/blob/cfc916791313c510b55e9b2b78ab6ebec6528f39/Snowflake.Data/Core/SFSession.cs#LL266C36-L266C57

The GUID seems to be .NET specific so that could be why it's not on JDBC

test/unit/mock/mock_http_client.js
sfc-gh-ext-simba-lf: I'm not aware why .NET connector includes a requestGUID, but since JDBC connector doesn't use it I could remove the requestGUID parameter on node so it will have parity with JDBC

sfc-gh-wshangguan (commented):
Let's remove the GUID and rerun the end to end test to see if it works. Others LGTM. Thanks for making the fix.

sfc-gh-wshangguan (approved):
LGTM

PR #499, opened 2023-05-23
Fix: Updating the way how we invoke the request from urllib

anandhud:
**Details**

- I have been continuously hitting the below error while running a lot of tests on my app. 
- Changed the request requirement approach and tested it


sfc-gh-dszmolka:
thank you for your contribution! `urllib` has been obsoleted and replaced by `axios` in driver version 1.9.0 so marking this as closed since the fix targeting `urllib` which doesn't exist anymore in this library. If still relevant, please comment and we can reopen.

PR #505, opened 2023-05-30
Add backpressure functionality to row stream

sfc-gh-ext-simba-lf:
Regarding issue 219

Add backpressure functionality to row stream so the connector will stop pushing data to the internal buffer when the highWaterMark value is reached ensuring the internal buffer will not overflow with data


lib/connection/result/row_stream.js
```diff
@@ -173,7 +173,40 @@ function RowStream(statement, context, options)
     // add the next row to the read queue
     process.nextTick(function ()
     {
-      self.push(externalizeRow(row, columns, mapColumnIdToExtractFnName));
+      // check if there are still rows available in the rowBuffer
+      if (rowBuffer)
+      {
+        rowIndex--; // decrement the index to include the previous row in the while loop
```
sfc-gh-dprzybysz, line 179: are we sure that `rowIndex` is not negative after decrement? some lines above it's set to `0`

sfc-gh-dprzybysz (commented):
Could we add some tests to check this behaviour?

lib/connection/result/row_stream.js
```diff
@@ -173,7 +173,40 @@ function RowStream(statement, context, options)
...
+        // push() data to readable stream until highWaterMark threshold is reached or all rows are pushed
+        while (rowIndex < rowBuffer.length)
+        {
+          row = rowBuffer[rowIndex];
+
+          // if buffer has reached the threshold based on the highWaterMark value then
+          // push() will return false and pause sending data to the buffer until the data is read from the buffer
+          if (!self.push(externalizeRow(row, columns, mapColumnIdToExtractFnName)))
+          {
+            rowIndex++;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Good idea, done

lib/connection/result/row_stream.js
sfc-gh-ext-simba-lf: The line before it always increments the rowIndex by 1: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/505/files/7e6cb500a769f6bb561440f9cd327dcc0c2e4e2c#diff-2e25c6e02363450501da256765b53482373c3976dfdc0dee01bd1c1160666c87R151

It's possible to do some refactoring so the decrement is no longer needed

sfc-gh-igarish: @sfc-gh-ext-simba-lf in one condition it set to 0 also in above code.

sfc-gh-ext-simba-lf: In the condition that rowIndex is set to 0, the rowBuffer is also set to null: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/505/files/7e6cb500a769f6bb561440f9cd327dcc0c2e4e2c#diff-2e25c6e02363450501da256765b53482373c3976dfdc0dee01bd1c1160666c87L158

If the rowBuffer is null, the decrement does not occur since it goes to the [else statement](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/505/files/7e6cb500a769f6bb561440f9cd327dcc0c2e4e2c#diff-2e25c6e02363450501da256765b53482373c3976dfdc0dee01bd1c1160666c87R205)

If the [rowBuffer is not null](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/505/files/7e6cb500a769f6bb561440f9cd327dcc0c2e4e2c#diff-2e25c6e02363450501da256765b53482373c3976dfdc0dee01bd1c1160666c87R177) then the rowIndex was not set to 0, and the decrement happens

sfc-gh-ext-simba-lf:
Added tests to check that the number of rows read per 'readable' event never exceeds the highWaterMark threshold

test/integration/testStreamRows.js
```diff
@@ -372,4 +373,107 @@ describe('Test Stream Rows API', function ()
...
+              });
+            }
+          });
+        }
+      ],
+      callback
+    );
+  };
+
+  const highWaterMarkValue = 10; // default parameter value is 10 (based on PARAM_ROW_STREAM_HIGH_WATER_MARK)
```
sfc-gh-ext-simba-lf, line 442: Currently the [highWaterMark parameter](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/connection/connection_config.js#L846) is not an external parameter so only the default value is tested

Only parameters that are external can be modified through the connection options: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/connection/connection_config.js#L743

Here's the [timeout parameter](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/connection/connection_config.js#L813) which is an external parameter as a reference

test/integration/testStreamRows.js
```diff
@@ -372,4 +373,107 @@ describe('Test Stream Rows API', function ()
...
+
+  it('test1000Rows', function (done)
```
sfc-gh-dprzybysz, line 444: Could we simplify parameterized tests with loop? 
```
[1000, 10000, 100000, 100000].forEach(rowCount => 
  it(`test ${rowCount} rows`, done => {
    testingFunc(highWaterMarkValue, 1000, done); // this function could be inlined here
  });
);

File: test/integration/testStreamRows.js [hunk 5]
```diff
@@ -372,4 +373,107 @@ describe('Test Stream Rows API', function ()
...
+  it('test100000Rows', function (done)
+  {
+    testingFunc(
+      highWaterMarkValue,
+      100000,
+      done
+    );
+  });
+
+  it('test1000000Rows', function (done)
```
sfc-gh-dprzybysz, line 471: do we need so many rows in tests?

test/integration/testStreamRows.js
sfc-gh-ext-simba-lf: Sure

test/integration/testStreamRows.js
sfc-gh-ext-simba-lf: It was included to test highWaterMark with large result sets as well

test/integration/testStreamRows.js
```diff
@@ -372,4 +373,107 @@ describe('Test Stream Rows API', function ()
       }
     });
   });*/
+
+});
+
+describe('testHighWaterMark', function ()
+{
+  this.timeout(300000);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: It's based off the timeout from the test suite above which was [increased](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/152/commits/d532aad2b547f0bca01da73814669b4613a16919#diff-787e9e5d2b97d8c09003c790fd43df0e12f6a2d9a2c65488b798890bf112a327R12) before

test/integration/testStreamRows.js
```diff
@@ -372,4 +373,107 @@ describe('Test Stream Rows API', function ()
       }
     });
   });*/
+
+});
+
+describe('testHighWaterMark', function ()
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: 1. The describe block above was for stream rows API, the new test suite is still related to stream rows but in regards to the highWaterMark property. I can put it in the same describe suite if you think it should be contained in that API block

2. Updated to follow name convention

test/integration/testStreamRows.js
```diff
@@ -372,4 +373,107 @@ describe('Test Stream Rows API', function ()
...
+                rowIndex = 0;
+
+                while (this.read() !== null)
+                {
+                  actualRowCount++;
+                  rowIndex++;
+                }
+
+                // assert the amount of rows read per loop never exceeds the highWaterMark threshold
+                assert.ok(rowIndex <= highWaterMark);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Currently, if the assert fails, the `callback()` from the complete event eventually gets called and the test fails with the assertion error

Added a try/catch which leads to invoking `callback(err)` from the error event

lib/connection/result/row_stream.js
sfc-gh-ext-simba-lf: Would you like for this to be refactored to remove the decrement?

sfc-gh-dprzybysz: It would be nice to have this decrement refactored or in `if` above we could verify if it's `>0` next to `rowBuffer` existance check

sfc-gh-ext-simba-lf: Ok, added a check to the if statement to make sure rowIndex is > 0

sfc-gh-igarish (approved):
LGTM

lib/connection/result/row_stream.js
```diff
@@ -173,7 +173,37 @@ function RowStream(statement, context, options)
...
+        {
+          // reset the buffer and index
+          rowBuffer = null;
+          rowIndex = 0;
+        }
+      }
+      else // No more rows left in the buffer
+      {
+        // Push the last row in the buffer
+        self.push(externalizeRow(row, columns, mapColumnIdToExtractFnName));
```
sfc-gh-igarish, line 205: Why special code for last row? what are chances we keep adding last row multiple times?

PR #509, opened 2023-05-31
SNOW-830232: Fix regression tests

sfc-gh-dprzybysz:
### Description
Fix 2 regression tests that fail in internal build.

### Checklist

### DIFF
--- a/test/integration/testArrayBindCustomerTable.js
+++ b/test/integration/testArrayBindCustomerTable.js
@@ -1,66 +1,57 @@
 /*
  * Copyright (c) 2015-2019 Snowflake Computing Inc. All rights reserved.
  */
-var async = require('async');
 var assert = require('assert');
 var testUtil = require('./testUtil');
-
-const sourceRowCount = 30000;
 
 describe('Test Stage Binding with Customer Table', function ()
 {
   this.timeout(300000);
-  var connection;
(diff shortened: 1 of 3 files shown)


test/integration/testArrayBindCustomerTable.js
```diff
@@ -1,66 +1,57 @@
...
+    await testUtil.destroyConnectionAsync(connection);
   });
 
   it('testArrayBindCustomerTable', function (done)
   {
-    var arrBind = [];
-    var count = 100000;
-    for(var i = 0; i<count; i++)
+    const arrBind = [];
+    const rowsToInsert = 10000; // 100000 rows causes "Statement exhausted compilation resources and was canceled. Please contact support." in regression environment
```
sfc-gh-pmotacki, line 39: Probably you tried investigate what does mean "Statement exhausted compilation resources". It is a problem of test environment?

sfc-gh-pmotacki (approved):

test/integration/testArrayBindCustomerTable.js
sfc-gh-dprzybysz: It seems to be problem of the regression environment. There is a question if we need so many records added in this test to check the behaviour. Maybe this test should be part of performance tests in the future.

sfc-gh-igarish (approved):
LGTM

PR #525, opened 2023-06-06
SNOW-358605: Make isValidAsync always resolve

sfc-gh-dprzybysz:
### Description
Catch `heartbeat` errors and return `false` then from `connection.isValidAsync`

### Checklist

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -140,14 +140,25 @@ function Connection(context)
     });
   };
 
+  /**
+   * @return {Promise<boolean>}
+   */
   this.isValidAsync = async () =>
   {
     if (!this.isUp())
     {
       return false;
     }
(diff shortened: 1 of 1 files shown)


sfc-gh-igarish (approved):
LGTM

sfc-gh-pfus (approved):

lib/connection/connection.js
```diff
@@ -140,14 +140,25 @@ function Connection(context)
     });
   };
 
+  /**
+   * @return {Promise<boolean>}
```
sfc-gh-pmotacki, line 144: This is like partially documented. By the name of the function I know what the function implements, but only the return type is different than in other places

sfc-gh-pmotacki (approved):

PR #535, opened 2023-06-13
SNOW-358605: Let heartbeatAsync return the same result as with select 1

sfc-gh-dprzybysz:
### Description

### Checklist

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -136,7 +136,8 @@ function Connection(context)
   {
     return new Promise((resolve, reject) =>
     {
-      this.heartbeat((err, body) => err ? reject(err) : resolve(body));
+      // previous version of driver called `select 1;` which result in `[ { '1': 1 } ]`
+      this.heartbeat((err) => err ? reject(err) : resolve([ { '1': 1 } ]));
     });
   };
 
@@ -151,8 +152,8 @@ function Connection(context)
     }
     try
     {
-      const heartbeatResult = await this.heartbeatAsync()
-      return heartbeatResult.success;
+      await this.heartbeatAsync()
+      return true;
     }
     catch (e)
     {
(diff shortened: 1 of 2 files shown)


lib/connection/connection.js
```diff
@@ -136,7 +136,8 @@ function Connection(context)
   {
     return new Promise((resolve, reject) =>
     {
-      this.heartbeat((err, body) => err ? reject(err) : resolve(body));
+      // previous version of driver called `select 1;` which result in `[ { '1': 1 } ]`
+      this.heartbeat((err) => err ? reject(err) : resolve([ { '1': 1 } ]));
```
sfc-gh-igarish, line 140: why the new code care about previous version of driver?

test/integration/testConnection.js
```diff
@@ -1159,7 +1159,7 @@ describe('Connection Test - Heartbeat', () =>
   it('call heartbeat url as promise', async () =>
   {
     const rows = await connection.heartbeatAsync();
-    assert.ok(rows.success);
+    assert.deepEqual(rows,  [ { '1': 1 } ]);
```
sfc-gh-igarish, line 1162: In JDBC we have certain tests for latest driver and few tests for older driver. Do you think we can do the same here, if we are running some tests with older version?

lib/connection/connection.js
sfc-gh-dprzybysz: method `heartbeatAsync` was available in previous version of driver (the new method we added is `isValidAsync`) and someone could use this method and parse output from succesfully resolved promise - which is different now since it is using SF heartbeat, not `Select 1`

test/integration/testConnection.js
sfc-gh-dprzybysz: do you mean running such compatibility tests during each build or have such tests running only on upcoming release? we can think about such tests to ensure that everything is fine between releases

sfc-gh-igarish: For each build. i.e. for each PR.

sfc-gh-igarish:
LGTM. Once you raise the PR, I can approve it.

sfc-gh-igarish (approved):
LGTM

PR #536, opened 2023-06-14
Replace better-eval with vm

sfc-gh-ext-simba-lf:
Regarding issue 366

The PR replaces better-eval with vm for converting raw variant strings. The better-eval package creates a new VM for each conversion while this change creates a single VM and re-uses it for all conversions


sfc-gh-ext-simba-lf:
Added test case similar to the repro that calls select on a variant column containing a large result set

sfc-gh-igarish (approved):
Please do pref test and put numbers in this ticket before and after the change.

sfc-gh-ext-simba-lf:
> Please do pref test and put numbers in this ticket before and after the change.

Average execution time of testVariantLarge over 10 runs: 
 - better-eval: 22.6s
 - vm: 4.4s

package.json
```diff
@@ -37,6 +36,7 @@
     "tmp": "^0.2.1",
     "urllib": "^2.38.0",
     "uuid": "^3.3.2",
+    "vm": "0.1.0",
```
sfc-gh-dprzybysz, line 39: Do we need to add this library? It was released 7 years ago https://www.npmjs.com/package/vm?activeTab=readme and it has information that we should not install it since it's part of node now https://github.com/DiegoRBaquero/node-vm
(I removed this dependency and it seems to work fine)

test/integration/testLargeResultSet.js
```diff
@@ -113,4 +113,126 @@ describe('Large result Set Tests', function ()
...
+          complete: function (err, stmt)
+          {
+            if (err)
+            {
+              callback(err);
+            }
+
+            try
+            {
+              assert.strictEqual(stmt.getNumUpdatedRows(), sourceRowCount);
```
sfc-gh-dprzybysz, line 184: if this assertion fails then callback in line below won't be called and the test could stuck till timeout?

test/integration/testLargeResultSet.js
```diff
@@ -113,4 +113,126 @@ describe('Large result Set Tests', function ()
...
+      {
+        connection.execute({
+          sqlText: insertVariant,
+          binds: [JSON.stringify(arrJSON)],
+          complete: function (err, stmt)
+          {
+            if (err)
+            {
+              callback(err);
+            }
```
sfc-gh-dprzybysz, line 180: shouldn't we wrap next lines in `else`?

test/integration/testLargeResultSet.js
```diff
@@ -113,4 +113,126 @@ describe('Large result Set Tests', function ()
...
+              {
+                callback(err);
+              }
+            });
+          }
+        });
+      },
+      function (callback)
+      {
+        testUtil.executeCmd(connection, dropTableWithVariant, callback);
```
sfc-gh-dprzybysz, line 233: the main focus of the test is to insert and select from table with variant - maybe we can wrap this test in describe and create table put in `before` and drop table in `after`?

test/integration/testLargeResultSet.js
```diff
@@ -113,4 +113,126 @@ describe('Large result Set Tests', function ()
...
+            });
+            stream.on('error', function (err)
+            {
+              callback(err);
+            });
+            stream.on('end', function ()
+            {
+              try
+              {
+                assert.strictEqual(rowCount, sourceRowCount);
```
sfc-gh-dprzybysz, line 220: if this assertion fails than callback in line below won't be called and we will wait till timout?

test/integration/testLargeResultSet.js
```diff
@@ -113,4 +113,126 @@ describe('Large result Set Tests', function ()
...
   });
+
+  it('testVariantLarge', function (done)
+  {
+    const createTableWithVariant = 'create or replace table testVariantTable(colA variant)';
+    const insertVariant = 'insert into testVariantTable select value from table(flatten(parse_json(?)))';
+    const selectVariant = 'select * from testVariantTable';
+    const dropTableWithVariant = 'drop table if exists testVariantTable';
+
+    var arrJSON = [];
```
sfc-gh-dprzybysz, line 124: array could also be `const`

test/integration/testLargeResultSet.js
```diff
@@ -113,4 +113,126 @@ describe('Large result Set Tests', function ()
       }
     });
   });
+
+  it('testVariantLarge', function (done)
+  {
+    const createTableWithVariant = 'create or replace table testVariantTable(colA variant)';
+    const insertVariant = 'insert into testVariantTable select value from table(flatten(parse_json(?)))';
```
sfc-gh-dprzybysz, line 120: We are sending one gigantic array in json and let Snowflake to parse it and store as one json per row. Why don't we send binds as array of arrays? is it important for the test?

test/integration/testLargeResultSet.js
```diff
@@ -113,4 +113,126 @@ describe('Large result Set Tests', function ()
       }
     });
   });
+
+  it('testVariantLarge', function (done)
```
sfc-gh-dprzybysz, line 117: could we name this test in more elaborated way? to tell what is the core of the test? is it inserting or selecting?

package.json
sfc-gh-ext-simba-lf: It's removed now

test/integration/testLargeResultSet.js
sfc-gh-ext-simba-lf: Sure, test is renamed now

test/integration/testLargeResultSet.js
sfc-gh-ext-simba-lf: Done

test/integration/testLargeResultSet.js
sfc-gh-ext-simba-lf: You're right, it would've called callback again otherwise. Added else clause

test/integration/testLargeResultSet.js
sfc-gh-ext-simba-lf: It goes to the catch clause and calls callback(err)

test/integration/testLargeResultSet.js
sfc-gh-ext-simba-lf: Ok, placed the test in its own describe and create/drop table are now in the before/after of the test

test/integration/testLargeResultSet.js
sfc-gh-ext-simba-lf: This is how we were advised by the previous engineer when inserting multiple rows of variant data

I tried binding an array of arrays but got `OperationFailedError: SQL compilation error: error line 1 at position 53 Bind variable ? not set.`
```
sqlText: "insert into testVariantTable(colA) select parse_json(?)",
binds: [ [JSON.stringify({ 'root': 'value' })], [JSON.stringify({ 'root': 'value' })] ],
```

I then tried using insert into values but got a `Error [OperationFailedError]: SQL compilation error: Unsupported data type 'VARIANT'.` instead
```
sqlText: "insert into testVariantTable(colA) values(?)",
binds: [ [{ 'root': 'value' }], [{ 'root': 'value' }] ],
```

However, that error did not show up when array binding non-variant data
```
sqlText: 'insert into testTable values(?)',
binds: [[1], [2], [3]],
```

And running the first query without array bind was successful
```
sqlText: 'insert into testVariantTable(colA) select parse_json(?)',
binds: [JSON.stringify({ 'key': 'value' })],
```

Is there a different combination of sql and binds that I should be using instead?

lib/connection/result/column.js
```diff
@@ -552,7 +553,8 @@ function convertRawVariant(rawColumnValue, column, context)
   {
     try
     {
-      ret = betterEval("(" + rawColumnValue + ")");
+      // run code on a different scope
+      ret = vm.runInContext("(" + rawColumnValue + ")", VM_CONTEXT);
```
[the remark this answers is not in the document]
yanavais: Additional input that might be helpful. We noticed that `better-eval` is very slow in JSON parsing comparing to JSON.parse(). Most, if not all of our `variant` fields are JSON. Maybe it makes sense to apply `JSON.parse() first and leave `better-eval` as a fallback.

test/integration/testDataType.js
```diff
@@ -31,7 +31,8 @@ describe('Test DataType', function ()
   var insertDouble = 'insert into testDouble values(123.456)';
   var insertLargeNumber = 'insert into testNumber values (12345678901234567890123456789012345678)';
   var insertRegularSizedNumber = 'insert into testNumber values (100000001)';
-  var insertVariant = 'insert into testVariant select parse_json(\'{a : 1 , b :[1 , 2 , 3, -Infinity, undefined], c : {a : 1}}\')';
+  var insertVariantJSON = 'insert into testVariant select parse_json(\'{a : 1 , b :[1 , 2 , 3, -Infinity, undefined], c : {a : 1}}\')';
```
sfc-gh-dprzybysz, line 34: let's add new code with const instead of var

test/integration/testDataType.js
```diff
@@ -244,6 +245,34 @@ describe('Test DataType', function ()
       );
     });
 
+    it('testVariantXML', function (done)
+    {
+      async.series(
+        [
+          function (callback)
+          {
+            testUtil.executeCmd(connection, createTableWithVariant, callback);
```
sfc-gh-dprzybysz, line 254: can we create describe for variant and in before/afterEach/after add createTable/truncateTable/dropTable?

test/unit/snowflake_config_test.js
```diff
@@ -0,0 +1,215 @@
...
+    }
+  });
+
+  after(function ()
+  {
+    snowflake.configure(originalConfig);
+  });
+
+  ///////////////////////////////////////////////////////////////////////////
+  //// Test invalid arguments                                            ////
```
sfc-gh-dprzybysz, line 34: instead of comment let's wrap test cases in describe

test/unit/snowflake_config_test.js
```diff
@@ -0,0 +1,215 @@
...
+
+  var index, length, testCase;
+  for (index = 0, length = negativeTestCases.length; index < length; index++)
+  {
+    testCase = negativeTestCases[index];
+    it(testCase.name, createNegativeITCallback(testCase));
+  }
+
+  ///////////////////////////////////////////////////////////////////////////
+  //// Test valid arguments                                              ////
```
sfc-gh-dprzybysz, line 96: let's create describe instead of comments

test/unit/snowflake_config_test.js
```diff
@@ -0,0 +1,215 @@
...
+      finally
+      {
+        assert.ok(error);
+        assert.strictEqual(error.code, testCase.errorCode);
+      }
+    };
+  };
+
+  var index, length, testCase;
+  for (index = 0, length = negativeTestCases.length; index < length; index++)
```
sfc-gh-dprzybysz, line 89: we can simplify test cases to 
```
negativeTestCases.forEach(testCase => {it(testCase.name, createNegativeITCallback(testCase))})
```
or even inline createNegativeITCallback in test body

The same could be done with positive cases

test/unit/snowflake_config_test.js
```diff
@@ -0,0 +1,215 @@
...
+        options:
+        {
+          jsonColumnVariantParser: rawColumnValue => require('vm').runInNewContext("(" + rawColumnValue + ")")
+        }
+      },
+      {
+        name: 'xml parser',
+        options:
+        {
+          xmlColumnVariantParser: rawColumnValue => (require("fast-xml-parser").XMLParser()).parse(rawColumnValue)
```
sfc-gh-dprzybysz, line 175: is `new` missing when XMLParser is created? here we give example where exception is probably thrown by who would catch it in our production code when variant is parsed?

sfc-gh-dprzybysz (changes requested):
Let's add run lint (`npm run lint:check -- FILENAME`) for new code. Especially new files could be even changed with `npm run lint:fix -- FILENAME`.

test/integration/testDataType.js
```diff
@@ -216,7 +217,7 @@ describe('Test DataType', function ()
 
   describe('testSemiStructuredDataType', function ()
   {
-    it('testVariant', function (done)
+    it('testVariantJSON', function (done)
```
sfc-gh-dprzybysz, line 220: can we add additional describe and tests where we override default json and xml parser?

alternative implementation for JSON could be just `JSON.parse` and for xml we can leave todo to implement the test in https://github.com/snowflakedb/snowflake-connector-nodejs/pull/538 (e.g. `TODO SNOW-830291: add custom xml parser test` should be enough)

odedpeer:
> > Please do pref test and put numbers in this ticket before and after the change.
> 
> Average execution time of testVariantLarge over 10 runs:
> 
> * better-eval: 22.6s
> * vm: 4.4s

what is the average execution time of `eval`? I'd like to see the difference between `vm` and `eval`

lib/connection/result/column.js
```diff
@@ -552,22 +552,16 @@ function convertRawVariant(rawColumnValue, column, context)
...
-      else
-      {
-        // TODO: log the error
+      ret = GlobalConfig.xmlColumnVariantParser(rawColumnValue);
 
-        // throw the error
+      // If both json/xml parser was unsuccessful, then throw the parsing error
+      if (!ret)
+      {
+        Logger.getInstance().debug("Unable to parse raw variant string: " + parseError);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: When logging with `%o`, the error stack shows the data that was being parsed. Should we remove that data before logging?

Currently, only the error message is logged

lib/connection/result/column.js
```diff
@@ -552,22 +552,16 @@ function convertRawVariant(rawColumnValue, column, context)
...
-      // ensure each tag is enclosed and all attributes and elements are valid
-      if (XMLValidator.validate(rawColumnValue) === true)
-      {
-        // use XML parser
-        ret = new XMLParser().parse(rawColumnValue);
-      }
-      else
-      {
-        // TODO: log the error
+      ret = GlobalConfig.xmlColumnVariantParser(rawColumnValue);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Oops sorry I didn't realize I deleted it. I saw that comment but I wasn't sure whether to keep the logging/catch inside the default xml parser function since the default parser could be replaced

I thought about wrapping it in its own try/catch so it doesn't rely on the custom function to do the logging/catching. Do you think a nested try/catch for the xml parsing is ok or keeping it in the default function is enough?

lib/global_config.js
```diff
@@ -10,6 +10,12 @@ const Util = require('./util');
 const Errors = require('./errors');
 const Logger = require('./logger');
 
+const vm = require('vm');
+const VM_CONTEXT = vm.createContext() // create a new context so VM does not have to make a new one for each conversion
+
+const { XMLParser, XMLValidator } = require("fast-xml-parser");
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Done

lib/global_config.js
```diff
@@ -10,6 +10,12 @@ const Util = require('./util');
...
+const xmlParser = new XMLParser();
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Ok

lib/global_config.js
```diff
@@ -157,3 +173,44 @@ const rest = {
...
+// The default XML parser
+exports.xmlColumnVariantParser = rawColumnValue =>
+{
+  // check if raw string is in XML format
+  // ensure each tag is enclosed and all attributes and elements are valid
+  if (XMLValidator.validate(rawColumnValue) === true)
+  {
+    // use XML parser
+    return xmlParser.parse(rawColumnValue);
+  }
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Alright, added the else and throw

lib/global_config.js
```diff
@@ -157,3 +173,44 @@ const rest = {
...
+};
+
+/**
+ * Updates the value of the 'xmlColumnVariantParser' parameter.
+ *
+ * @param {function} value
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Added the expected signature. I put down `any` as the expected result, let me know if it should be `Object`

lib/global_config.js
```diff
@@ -157,3 +173,44 @@ const rest = {
...
 };
 exports.rest = rest;
+
+// The default JSON parser
+exports.jsonColumnVariantParser = rawColumnValue => vm.runInContext("(" + rawColumnValue + ")", VM_CONTEXT);
+
+/**
+ * Updates the value of the 'jsonColumnVariantParser' parameter.
+ *
+ * @param {function} value
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Added the expected signature

test/integration/testDataType.js
sfc-gh-ext-simba-lf: Ok

lib/core.js
```diff
@@ -171,7 +171,7 @@ function Core(options)
         {
           // check that the specified value is a valid tag
           Errors.checkArgumentValid(LoggerCore.isValidLogTag(logTag),
-            ErrorCodes.ERR_GLOGAL_CONFIGURE_INVALID_LOG_LEVEL);
+            ErrorCodes.ERR_GLOBAL_CONFIGURE_INVALID_LOG_LEVEL);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Do you prefer closing that PR after this is merged? Or close it now

test/integration/testDataType.js
sfc-gh-ext-simba-lf: Sure, I also copied the way you used `executeCmdAsync` from the large result test case

test/unit/snowflake_config_test.js
sfc-gh-ext-simba-lf: Ok

test/unit/snowflake_config_test.js
sfc-gh-ext-simba-lf: Ok

test/unit/snowflake_config_test.js
sfc-gh-ext-simba-lf: Nice, I'll use this for both the test cases. I put the callback function's body inside if that's what you meant

test/unit/snowflake_config_test.js
sfc-gh-ext-simba-lf: You're right, it was missing. Added `new` when creating XMLParser and corrected the parenthesis placement

I mentioned in a comment above about wrapping the xml parsing in its own try/catch if possible

test/integration/testDataType.js
sfc-gh-ext-simba-lf: I tried using `JSON.parse()` but it failed because the test data contained `Infinity` and `undefined` which doesn't seem to be supported by JSON

I made another insert query with only valid JSON and it works

lib/global_config.js
```diff
@@ -157,3 +172,51 @@ const rest = {
...
+    throw validateResult;
+  }
+};
+
+/**
+ * Updates the value of the 'xmlColumnVariantParser' parameter.
+ *
+ * @param {function: (rawColumnValue: string) => any} value
+ */
+exports.setXmlColumnVariantParser = function (value)
```
sfc-gh-pmotacki, line 216: In changes of SNOW-830291 I've prepared possibility to set custom xmlParserConfiguration:  
/**
 * Set custom xmlParser configuration.
 * Parametrs that you can override:
 *   ignoreAttributes - default true,
 *   attributeNamePrefix - default '@',
 *   attributesGroupName - default null,
 *   alwaysCreateTextNode - default false
 *
 * @param {object} value
 */
Examle:
```
snowflake.configure({
    xmlParserConfiguration: {
        ignoreAttributes: true,
        attributeNamePrefix: '@'
    },
})
``` 
and it is used during creation of new XMLParser
I wonder if it will be useful to have both ways configured? This is more to discuss with @sfc-gh-igarish. When we decide I will implement it in my task

lib/core.js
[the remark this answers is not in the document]
sfc-gh-dprzybysz: Definitely after with comment like "It is fixed in ..."

lib/connection/result/column.js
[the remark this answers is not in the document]
sfc-gh-dprzybysz: it's hard question... but maybe we should keep it consistent with json parser and log also here

Maybe we should have such logic here (is pseudo js):

```
try {
  return parseJson()
} catch (jsonParseError) {
  try {
     return parseXml()
  } catch(xmlParseError) {
    log(variant cannot be parsed neither as JSON (jsonParseError.message) nor as XML (xmlParseError.message)
    throw new VariantParseError(jsonParseError, xmlParseError)
  }
}
```

@sfc-gh-igarish WDYT?

lib/global_config.js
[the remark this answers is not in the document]
sfc-gh-dprzybysz: It's interesting question - if we have sth declared as variant then we can put as json `{}`, `[]` or `5` and all of them are  considered by parser as valid jsons? (https://www.json.org/json-en.html)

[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: It looks like they are considered valid, no errors were thrown by the default json parser when parsing those 3 values

lib/connection/result/column.js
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: This could work, info on both error messages is provided through the logging and VariantParseError

sfc-gh-dprzybysz (approved):

lib/connection/result/column.js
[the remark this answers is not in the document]
sfc-gh-ext-simba-lf: Hello, apologies for the delay, the decision to not use `eval` was based on the security team's advise. The parser can be configured now though so if there is a parser you'd like to use over `vm`, then it is possible to configure:

- Configure the connector to use `eval` when parsing json

```
snowflake.configure({
  jsonColumnVariantParser: rawColumnValue => eval("(" + rawColumnValue + ")");
})
```
- Or use `JSON.parse()` instead
```
snowflake.configure({
  jsonColumnVariantParser: rawColumnValue => JSON.parse(rawColumnValue)
})
```

PR #547, opened 2023-06-22
fix process crash on put file command

nvedenyuk:
put command uses getDigestAndSizeForFile() to generate hash for a file - this function read all file contents to buffer
and only after passed for hash generation. Issue is more critical if disable AUTO_COMPRESS and try to upload file. Current change uses stream instead to prevent crashes on a bigger files


nvedenyuk:

sfc-gh-dszmolka:
hi and thank you for your contribution! if this is still relevant with the latest driver version of 1.9.0, can you please link the corresponding Github issue which the PR is aiming to solve ? A repro is more than welcome too :) Thank you !

PR #579, opened 2023-07-18
revert word-wrap upgrade

sfc-gh-ext-simba-ll:
Latest word-wrap is released and is automatically downloaded by urllib. we no longer needs to override it.


sfc-gh-igarish (approved):
LGTM

PR #585, opened 2023-07-19
Revert "SNOW-847264: Always run tests on node v12 and drop centos6 test builds"

sfc-gh-kdama:
Reverts snowflakedb/snowflake-connector-nodejs#573


sfc-gh-igarish (approved):
LGTM

Issue #486, opened 2023-05-03
Add the ability to execute query without fetching the result


TimShilov:
## What is the current behavior?
At the moment there seems to be no way to execute a query and only get queryId. Every time you execute the query the whole response is still loaded which leads to memory issues (like [this one](https://github.com/snowflakedb/snowflake-connector-nodejs/issues/43)).

## What is the desired behavior?
I would like to execute the request, wait for its completion and only get queryId in response for further usage with `RESULT_SCAN()`.

## How would this improve `snowflake-connector-nodejs`?
That will greatly optimize the network usage of the client and will allow complex transformations to be done.

## References, Other Background
I was talking about it with Support (Case Number 00510759) and they have pointed me to https://github.com/snowflakedb/snowflake-connector-nodejs/issues/43 which seems to be in the works and should help resolve memory issues as I will be able to stream data (to nowhere I guess) more effectively without using too much memory but I feel like it is wrong to fetch something I don't need in the first place. I want to further process the data inside Snowflake (as mentioned in [documentation](https://docs.snowflake.com/en/user-guide/querying-persisted-results#post-processing-query-results)). Responses in my case can be huge, tens of millions of rows and even memory aside, that's going to waste time and network traffic.

## What is your Snowflake account identifier, if any?
xe29001.us-central1.gcp

sfc-gh-dszmolka:
hello and thank you for submitting this issue ! if i understand correctly, you would like async query execution capabilities for the nodejs driver, something similar to what for example the [JDBC driver](https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-using#label-jdbc-asynchronous-query) and the [SQL REST API](https://docs.snowflake.com/en/developer-guide/sql-api/reference#query-parameters) already has .

this request is already with the driver team under the internal jira ID with which i just updated the Issue title. sadly, at this moment the improvement is not prioritised and therefore i am unable to give even an estimation for the expected implementation and availability timeline. Since you're a Snowflake customer, i recommend as of next step to contact your Sales rep. and tell them how important this capability would be for you. This helps the product team with prioritisation and resource allocation between the many requests. 

Closing this issue for now but if you feel your request would be something entirely different from the aforementioned async query execution capability (submit request -> immediately return queryID and control to the caller -> execute further calls in arbitrary later time to retrieve the results/do something with the already-executed query) , please comment and I can reopen if necessary.

TimShilov:
[...]

Issue #208, opened 2021-10-06
SNOW-502598 Unable to execute statements in async mode (add async query execution capabilities)


gievln:
Please answer these questions before submitting your issue. Thanks!

1. What version of NodeJS are you using (`node --version` and `npm --version`)?
node: `v14.18.0`
yarn: `1.22.10`

2. What operating system and processor architecture are you using?
For Prod env: [AWS Lambda](https://aws.amazon.com/lambda/)
For local testing:
```
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:      focal
```

3. What are the component versions in the environment (`npm list`)?
```
"snowflake-sdk": "1.6.3",
"@types/snowflake-sdk": "1.6.1"
```

Hey, is there a way to execute statements in async manner so that after starting a heavy procedure I would be able to monitor its status using snowflake's `query_history()` calls?
Looks like `snowflake-connector-nodejs`  doesn't allow to set `async` execution parameter to the inner SnowFlake API call on statement execution…
In general Snowflake REST API allows to execute sql statement in async mode `async:true` (see [here](https://docs.snowflake.com/en/developer-guide/sql-api/reference.html#query-parameters)) - this is an url parameter but it’s not allowed in NodeJs snowflake driver. I believe this is the place where we should be able to set async param:
```
...
url: Url.format(
        {
          pathname: '/queries/v1/query-request',
          search: QueryString.stringify(
            {
              requestId: statementContext.requestId,
              async: statementContext.async // <--- NEW LINE THAT DOESN'T EXIST AT THE MOMENT
            })
}),
```
source code is here: [https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/connection/statement.js#L1134](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/connection/statement.js#L1134)

Is there a plan to introduce this async flag support in `snowflake-connector-nodejs`? Or maybe someone could recommend other approach to achieve this asynchronous behavior?

bjacques:
I see that [async queries are available for the Python driver](https://docs.snowflake.com/en/user-guide/python-connector-example.html#performing-an-asynchronous-query), but not Nodejs

TomNachman:
will there be any support for NodeJS driver in the future for this?

sfc-gh-dszmolka:
[...]

PR #592, opened 2023-07-24
SNOW-502598: Enable executing queries with `asyncExec: true`

markandrus:
### Description

My team is using the JavaScript Snowflake SDK. We need to be able to execute queries without waiting for the result. I see that the JavaScript Snowflake SDK has functionality for polling query results, a Result class, a RowStream class, etc.; however, it would be sufficient for us if we could just issue a request with `asyncExec: true`, get the Snowflake query ID back, and then monitor the query ourselves (for example, using [RESULT_SCAN](https://docs.snowflake.com/en/sql-reference/functions/result_scan)).

With this in mind, I've tried to create a small patch that does just that. This may partially address SNOW-502598 (see #486 and #208). I've also prepared a patch for @types/snowflake-sdk:

```diff
--- a/index.d.ts
+++ b/index.d.ts
@@ -571,7 +571,13 @@ export type Connection = NodeJS.EventEmitter & {
          * - {@link https://docs.snowflake.com/en/user-guide/nodejs-driver-use.html#fetching-data-types-as-strings Fetching Data Types As Strings}
          */
         fetchAsString?: Array<'String' | 'Boolean' | 'Number' | 'Date' | 'JSON' | 'Buffer'> | undefined;
-        complete?: (err: SnowflakeError | undefined, stmt: Statement, rows: any[] | undefined) => void;
+
+        /**
+         * Set to `true` to execute the query asynchronously, and the `complete` callback will receive an object with the query ID.
+         */
+        asyncExec?: boolean
+
+        complete?: (err: SnowflakeError | undefined, stmt: Statement, result: any[] | { queryId: string } | undefined) => void;
     }): Statement;
 
     /**
```

If you use Yarn v2 or later, you can try these changes out by

1. Updating your dependency to `git@github.com:propeldata/snowflake-connector-nodejs.git#head=asyncExec`
2. Using `yarn patch` to apply the @types/snowflake-sdk patch above.

### Warning

**This is definitely a hack,** because the JavaScript Snowflake SDK's `execute` method is supposed to return a Statement object with APIs you can invoke, like `getColumns` and `streamRows`. These are probably not working with this patch; however, my team does not use these APIs anyway. We typically invoke the JavaScript Snowflake SDK as follows:

```ts
const { queryId } = new Promise((resolve, reject) => {
  client.execute({
    sqlText: 'CALL SYSTEM$WAIT(5, \'MINUTES\')',
    asyncExec: true,
    complete (err, _stmt, result) {
      if (err != null) {
        reject(err)
      } else {
        resolve(result)
      }
    }
  })
})
```

I would look to the JavaScript Snowflake SDK team for ideas on how to better integrate this functionality with the Statement object. Perhaps, when `asyncExec: true` is set, we should just synthesize a Result with a RowStream containing a singular row. That singular row could contain a singular column with the query ID. Maybe this would make sense?

### Checklist

### DIFF
--- a/lib/connection/statement.js
+++ b/lib/connection/statement.js
@@ -198,6 +198,7 @@ exports.createStatementPostExec = function (
   statementContext.complete = complete;
   statementContext.streamResult = statementOptions.streamResult;
   statementContext.fetchAsString = statementOptions.fetchAsString;
+  statementContext.asyncExec = statementOptions.asyncExec;
   statementContext.multiResultIds = statementOptions.multiResultIds;
   statementContext.multiCurId = statementOptions.multiCurId;
 
@@ -361,6 +362,7 @@ function createContextPreExec(
   statementContext.complete = complete;
   statementContext.streamResult = statementOptions.streamResult;
   statementContext.fetchAsString = statementOptions.fetchAsString;
+  statementContext.asyncExec = statementOptions.asyncExec;
   statementContext.multiResultIds = statementOptions.multiResultIds;
   statementContext.multiCurId = statementOptions.multiCurId;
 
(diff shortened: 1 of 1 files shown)


sfc-gh-pmotacki:
@markandrus In version 1.9.1 of snowflake-sdk we added feature: "Added support for executing asynchronous queries".

PR #596, opened 2023-07-25
Test nodejs16 and 19 with different cloud provider

sfc-gh-ext-simba-ll:


.github/workflows/build-test.yml
```diff
@@ -130,18 +130,47 @@ jobs:
                 TARGET_DOCKER_TEST_IMAGE: ${{ matrix.image }}
               run: ./ci/test.sh
 
-    test-mac-optional:
+    test-mac-azure-optional:
```
sfc-gh-dprzybysz, line 133: why cannot we just run tests with and don't duplicate build configuration?
```
matrix:
  cloud: [ 'AZURE', 'GCP' ]
  nodeVersion: ['16.x', '18.x']
```

can we take it for windows and mac?

.github/workflows/build-test.yml
```diff
@@ -130,18 +130,47 @@ jobs:
...
         needs: test-mac-required
         name: Tests on Mac
```
sfc-gh-dprzybysz, line 135: name should be changed (or parameters applied as in my first comment) - the same also in other changed builds

sfc-gh-dprzybysz (changes requested):

.github/workflows/build-test.yml
sfc-gh-dprzybysz: and let's add `fail-fast: false` to not cancel other builds when one matrix item fails

PR #598, opened 2023-07-27
Add manual test script for connectAsync

sfc-gh-ext-simba-ll:
These tests needs manual interaction and cannot be added as part of automated test.

### DIFF
--- a/EnableConnectAsyncTests.py
+++ b/EnableConnectAsyncTests.py
@@ -1,0 +1,15 @@
+import os
+f = open("test/integration/testConnection.js", "r")
+data = f.read()
+data = data.replace("describe(", "describe.skip(")
+data = data.replace("describe.skip('Connection test - external browser'", "describe('Connection test - external browser'")
+data = data.replace("describe.skip('Connection test - oauth'", "describe('Connection test - oauth'")
+data = data.replace("describe.skip('Connection test - okta'", "describe('Connection test - okta'")
+data = data.replace("const connOption = require('./connectionOptions');", "const connOption = require('../integration/connectionOptions');")
+data = data.replace("const testUtil = require('./testUtil');", "const testUtil = require('../integration/testUtil');")
+
+if not os.path.exists('test/manual'):
+   os.makedirs('test/manual')
+
+f = open("test/manual/testConnection.js", "w")
+f.write(data)
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ npm test
 ```
 or
 ```
-npm test:unit
+npm run test:unit
 ```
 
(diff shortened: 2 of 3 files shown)


EnableConnectAsyncTests.py
```diff
@@ -0,0 +1,15 @@
+import os
```
sfc-gh-dprzybysz, line 1: Why do we use python script to rewrite the tests? one has to have python installed. Cannot we use for example environment variable and keep everything in node js?

we can wrap connection manual tests in side if like that:
```
if (process.env.RUN_MANUAL_TESTS_ONLY === 'true') {
  describe.only('Run manual tests', () => {
     describe('Connection test - external browser'", "describe('Connection test - external browser', () => ...)
     // and other tests here without skip
  })
}
```

and then in package json:
```
"test:manual": "RUN_MANUAL_TESTS_ONLY=true npm run test:single -- test/integration/testConnection.js"
```

sfc-gh-dprzybysz (changes requested):

EnableConnectAsyncTests.py
sfc-gh-ext-simba-ll: Updated. This looks better than generate a new file using python.
Although RUN_MANUAL_TESTS_ONLY cannot be set in package.json like that. 
So I made a change to readme to set this variable before test.

test/integration/testConnection.js
```diff
@@ -10,1044 +10,1179 @@ const Util = require('./../../lib/util');
...
+          done(`Max after ${maxSleep} it's expected to complete ${totalConnections} but completed ${completedConnection}`)
+        }
+      }, sleepMs);
+
+      timeout();
     });
   });
 
+  // Skipped - requires manual interaction to enter credentials on browser
+  describe.skip('Connection test - external browser', function () {
```
sfc-gh-dprzybysz, line 276: those tests are moved to if branch - do we need them here with skip? the same also for other cases where the tests are duplicated but skipped

maybe for cleaner implementation let's create new file e.g. `testManualConnection.js` and move all those tests (with env variable checking if) to new file and in `testConnection` let's leave only automatic tests

sfc-gh-ext-simba-jy:

sfc-gh-ext-simba-jy:
Hello Team, I'm John who took over  Linda's PR.

sfc-gh-igarish:
> Hello Team, I'm John who took over Linda's PR.

Hi John, welcome aboard. If you have address the Dominic question in the PR then resolve it and ping him again to review it.

sfc-gh-dprzybysz (approved):

Issue #609, opened 2023-08-08
SNOW-889050: NextJS Weird Warning Message


detailedghost:
## Summary
Working with an NX monorepo with NextJS. I'm able to query the database, but grabbing a weird warning message when I launch the application. Figured it might be worth flagging, sorry if there's a duplicate issue. Regardless, I appreciate any response.

## Requested info
1. What version of NodeJS driver are you using?
    - v18.16.0
2. What operating system and processor architecture are you using? 
    - Windows 10 with WSL 2 Ubuntu
4. What version of NodeJS are you using?
    - v18.16.0
5. What are the component versions in the environment (`npm list`)?
```bash
├── @nx/cypress@16.6.0
├── @nx/eslint-plugin@16.6.0
├── @nx/jest@16.6.0
├── @nx/js@16.6.0
├── @nx/linter@16.5.5
├── @nx/next@16.6.0
├── @nx/react@16.5.5
├── @nx/workspace@16.6.0
├── @radix-ui/react-navigation-menu@1.1.3
├── @radix-ui/react-slot@1.0.2
├── @swc/cli@0.1.62
├── @swc/core@1.3.75
├── @swc/helpers@0.5.1
├── @tanstack/react-query@4.32.6
├── @testing-library/react@14.0.0
├── @types/jest@29.5.3
├── @types/node-fetch@2.6.4
├── @types/node@20.4.8
├── @types/react-dom@18.2.7
├── @types/react@18.2.18
├── @types/snowflake-sdk@1.6.13
├── @typescript-eslint/eslint-plugin@5.62.0
├── @typescript-eslint/parser@5.62.0
├── autoprefixer@10.4.14
├── babel-jest@29.6.2
├── class-variance-authority@0.7.0
├── clsx@2.0.0
├── cypress@4.12.1
├── eslint-config-next@13.4.12
├── eslint-config-prettier@8.1.0
├── eslint-plugin-cypress@2.14.0
├── eslint-plugin-import@2.28.0
├── eslint-plugin-jsx-a11y@6.7.1
├── eslint-plugin-react-hooks@4.6.0
├── eslint-plugin-react@7.32.2
├── eslint@8.46.0
├── jest-environment-jsdom@29.6.2
├── jest-environment-node@29.6.2
├── jest@29.6.2
├── lucide-react@0.263.1
├── next@13.4.1
├── node-fetch@3.3.2
├── nx-cloud@16.2.0
├── nx@16.5.5
[...]

PR #346, opened 2022-09-26
Override node-fetch version of @azure/storage-blob to 3.2.10

SimbaGithub:
Based off SF# 00402210 and snyk issue https://github.com/snowflakedb/snowflake-connector-nodejs/issues/334

The @azure/storage-blob package uses node-fetch as a dependency. Override node-fetch to 3.2.10 to fix the vulnerability

PR #610, opened 2023-08-09
adding encoding as a devDependency

sfc-gh-dszmolka:
should fix https://github.com/snowflakedb/snowflake-connector-nodejs/issues/609

the older versions of `node-fetch` on which some of our dependencies depend, use `encoding` as an optional peerDependency. v3 of `node-fetch` eliminated this dependency and due to [an override](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/346), we are already using it, however some of our dependencies still depend on the older `node-fetch` versions:
```
# npm ls node-fetch
snowflake-sdk@1.7.0 /node/snowflake-connector-nodejs
+-- @azure/storage-blob@12.15.0 overridden
| `-- @azure/core-http@3.0.2
|   `-- node-fetch@3.3.2.  <<< v3
`-- @google-cloud/storage@6.12.0
  +-- gaxios@5.1.3
  | `-- node-fetch@2.6.12.  <<< v2
  `-- teeny-request@8.0.3
    `-- node-fetch@2.6.12.  <<< v2
```

As a result, in certain cases users of `snowflake-sdk` can see a warning message like
```
Module not found: Can't resolve 'encoding' in 'path/node_modules/node-fetch/lib'
```

which is literally nothing but a warning message as everything works correctly, but it would be nice perhaps to get rid of this warning too.

I thought perhaps we could address this issue by bumping the dependencies (e.g. the GCP Storage) but even in their latest versions they seem to depend on `gaxios` which for now depends on v2 of `node-fetch` and looking at issues gaxios#508 and gaxios#445, I'm not sure how realistic is to think it will be moved to `node-fetch` v3 in the near future. Did not look at `teeny-request` therefore.

Users around the world are working around this issue with running `npm i -D encoding` in their projects, see links in the linked issue.

So maybe we can address this in `snowflake-sdk`, by adding `encoding` as an (unused) devDependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "winston": "^3.1.0"
   },
   "devDependencies": {
+    "encoding": "^0.1.13",
     "eslint": "^8.41.0",
     "mocha": "^10.1.0",
     "mock-require": "^3.0.3",


sfc-gh-dszmolka:
closing this for now. The issue is that node-fetch requires some module as a peer-dependency, but does not provide it in any ways. This does not seem to cause any issue in the Node.JS framework for which this driver is built and tested against. 

For other frameworks like Next.JS, solution is to install the missing package (`npm i -D encoding`, for example)

PR #622, opened 2023-08-18
SNOW-726738: [HTAP] LAST QUERY ID - Driver API

sfc-gh-ext-simba-jy:
### Description
Modifying getStatementId to getQueryId
- [x] changed the function name "getStatementId" to "getQueryId" in Result and Statement 

### Checklist


lib/connection/result/result.js
```diff
@@ -2,14 +2,14 @@
...
 
-var EventEmitter = require('events').EventEmitter;
-var Util = require('../../util');
-var Errors = require('../../errors');
-var Chunk = require('./chunk');
-var ResultStream = require('./result_stream');
-var ChunkCache = require('./chunk_cache');
-var Column = require('./column');
-var StatementType = require('./statement_type');
+const EventEmitter = require('events').EventEmitter;
```
sfc-gh-ext-simba-hx, line 5: are we planning to replace var with const/let now? If so, better to make it in separated PR to ensure all var get replaced an no mixed stuff left.

lib/connection/result/result.js
```diff
@@ -696,8 +664,7 @@ Result.prototype.getColumn = function (columnIdentifier)
  *
  * @returns {string}
  */
-Result.prototype.getStatementId = function ()
-{
+Result.prototype.getQueryId = function () {
```
sfc-gh-ext-simba-hx, line 667: I think better to have a new function of getQueryId() to return the same thing in case getStatementId() is used by existing customer application.

sfc-gh-ext-simba-jy: I will apply it and commit again soon.

sfc-gh-ext-simba-jy: Applied

lib/connection/result/result.js
sfc-gh-ext-simba-jy: Okay, I reverted them, and pushed the commit again.

sfc-gh-igarish (approved):
LGTM

test/integration/testStatement.js
```diff
@@ -205,7 +205,7 @@ describe('Statement Tests', function ()
 
                 assert.strictEqual(statement.getNumRows(), 1);
                 assert.ok(statement.getSessionState());
-                assert.ok(statement.getStatementId());
+                assert.ok(statement.getQueryId());
```
sfc-gh-dprzybysz, line 208: shouldn't we keep both assertions for statementId and queryId? and even check that they are equal as a next assertion?

test/unit/connection/statement_test.js
```diff
@@ -2,425 +2,362 @@
  * Copyright (c) 2015 Snowflake Computing Inc. All rights reserved.
  */
 
-var Statement = require('./../../../lib/connection/statement');
-var ErrorCodes = require('./../../../lib/errors').codes;
-var assert = require('assert');
+var Statement = require("./../../../lib/connection/statement");
```
sfc-gh-dprzybysz, line 5: do you change anything related to last query id in this file or is it just reformatting?

test/unit/snowflake_test.js
```diff
@@ -791,7 +791,7 @@ describe('connection.execute() statement successful', function ()
 
                 assert.strictEqual(statement.getNumRows(), 1);
                 assert.ok(Util.isObject(statement.getSessionState()));
-                assert.ok(Util.string.isNotNullOrEmpty(statement.getStatementId()));
+                assert.ok(Util.string.isNotNullOrEmpty(statement.getQueryId()));
```
sfc-gh-dprzybysz, line 794: here also we should keep both assertions for backward compatibility tests

lib/connection/statement.js
```diff
@@ -553,6 +553,10 @@ function BaseStatement(
     return context.statementId;
   };
 
+  this.getQueryId = function ()
+  {
+    return context.statementId;
```
sfc-gh-dprzybysz, line 558: do we really need the new functions in result and statement? are we planning to make one of them deprecated?

sfc-gh-dprzybysz (changes requested):

lib/connection/statement.js
sfc-gh-ext-simba-jy: At the first commit, I just replaced getStatementId to getQueryId. However, Harry worried about the existing users who use getStatementId. Therefore, we consider making a new function(getQueryId) which returns the same result is better. Please let me know if you want us to replace getStatementId to getQueryId

test/unit/snowflake_test.js
sfc-gh-ext-simba-jy: I will add them

test/unit/connection/statement_test.js
sfc-gh-ext-simba-jy: No, nothing was changed. Just reformatting happened.

lib/connection/statement.js
sfc-gh-dprzybysz: we can add a new method but we should mark the old one as deprecated probably? we can also replace it with BCR if we want

but my question is: cannot we just live with statementId in the whole code? do we have to have the method named with queryId?

test/unit/connection/statement_test.js
sfc-gh-dprzybysz: let's reformat it fully (with lint:fix) or remove it from this PR and put in separate one since it's not related to the implemented feature

sfc-gh-ext-simba-jy: I will remove it from this PR. I think it was added by mistake.

lib/connection/statement.js
sfc-gh-dprzybysz: ok, let's then add new method and mark the old one as deprecated (in nodejs you can do this by a comment probably) and until we delete it with the BCR release in the future test both methods since both can be used by our users

lib/connection/result/result.js
```diff
@@ -701,6 +701,11 @@ Result.prototype.getStatementId = function ()
   return this._statementId;
 };
 
+Result.prototype.getQueryId = function ()
+{
+  return this._statementId;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I will replace statemnetId field with queryId.

test/integration/testStatement.js
sfc-gh-ext-simba-jy: added testing for both getStatementId and getQueryId

lib/connection/result/result.js
```diff
@@ -701,6 +701,11 @@ Result.prototype.getStatementId = function ()
   return this._statementId;
 };
 
+Result.prototype.getQueryId = function ()
```
sfc-gh-pfus, line 704: Just asking - we need to have this value on:
1. prepared statements
2. SELECT queries
3. Non-SELECT (DML, DDL etc) queries
I see only two new functions. Can you confirm, that it is available for all cases?

sfc-gh-ext-simba-jy: Hello, it works in all three conditions.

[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Applied eslint rules for all the files that I changed.

test/unit/connection/statement_test.js
sfc-gh-ext-simba-jy: As we need to replace the field name "statementId" with "queryId", this file has some changes. I applied eslint rules and added this in this PR.

test/unit/snowflake_test.js
sfc-gh-dprzybysz: how about this test?

lib/connection/result/result.js
```diff
@@ -694,20 +634,29 @@ Result.prototype.getColumn = function (columnIdentifier)
  * Returns the statement id generated by the server for the statement that
  * produced this result.
  *
+ * @deprecated
```
sfc-gh-dprzybysz, line 637: let's add a comment above that `getQueryId` should be used instead

sfc-gh-dprzybysz (commented):
Do we really need to apply all those reformating in this PR? the change is rather small but we are changing thousands of lines - I think we should reformat the code in separate PR

lib/connection/result/result.js
sfc-gh-ext-simba-jy: Applied.

sfc-gh-ext-simba-jy:
@sfc-gh-dprzybysz Hello Domink, I created a new PR that only added feature changes without reformatting whole files:
[https://github.com/snowflakedb/snowflake-connector-nodejs/pull/628](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/628).

sfc-gh-ext-simba-jy:
Closed this PR because we created a new PR.

PR #623, opened 2023-08-21
SNOW-896805: Replace urllib with axios

sfc-gh-dprzybysz:
### Description
Replace urllib@2 with axios

### Checklist


lib/http/base.js
```diff
@@ -51,8 +52,12 @@ HttpClient.prototype.request = function (options)
   var body;
   var request;
 
-  // normalize the headers
-  headers = normalizeHeaders(options.headers);
+  headers = {
+    ...normalizeHeaders(options.headers),
+    // axios does not know how to unzip response without content-length and  https://github.com/axios/axios/issues/5298
```
sfc-gh-pfus, line 57: Can you explain more? As far as I see, this issue you mention has been resolved for a few months.

lib/http/base.js
```diff
@@ -51,8 +52,12 @@ HttpClient.prototype.request = function (options)
...
+    // and google api returns such gzipped response https://github.com/apache/libcloud/issues/1544 and https://issuetracker.google.com/issues/177896087
```
sfc-gh-pfus, line 58: Same here, the issue was resolved 2 years ago.

lib/http/base.js
```diff
@@ -51,8 +52,12 @@ HttpClient.prototype.request = function (options)
...
+    'Accept-Encoding': options.url.includes('storage.googleapis.com') ? 'application/json' : 'gzip'
```
sfc-gh-pfus, line 59: Are you sure? This is something like mixing two worlds - a world of response content type and response compression...

lib/services/large_result_set.js
```diff
@@ -31,7 +31,7 @@ function LargeResultSetService(connectionConfig, httpClient)
     return err && (
       err.code === 'ECONNRESET' || err.code === 'ETIMEDOUT' ||
       // errors named with TimoutError suffix are thrown from urllib
-      err.name === 'ResponseTimeoutError' || err.name === 'ConnectionTimeoutError'
+      err.name === 'AxiosError' && err.code === 'ECONNABORTED'
```
sfc-gh-pfus, line 34: Can we have some parenthesis here as this is not obvious when mixing || and &&.

lib/http/base.js
sfc-gh-dprzybysz: We found different root cause and applied different solution - check it out now

lib/http/base.js
sfc-gh-dprzybysz: it was an workaround - instead of application/json it could be any string, but we don't need it anymore

lib/services/large_result_set.js
sfc-gh-dprzybysz: it's not necessary, but we can

lib/http/base.js
```diff
@@ -97,21 +103,38 @@ HttpClient.prototype.request = function (options)
...
+          }
+        } else {
+          return options.callback(null, normalizeResponse(response), response.data);
         }
-        else
-        {
-          response.body = response.data.toString();
-          await options.callback(null, normalizeResponse(response), data.toString());
+      } else {
+        return null;
```
sfc-gh-pfus, line 126: No log here?

lib/services/large_result_set.js
```diff
@@ -30,8 +30,8 @@ function LargeResultSetService(connectionConfig, httpClient)
   {
     return err && (
       err.code === 'ECONNRESET' || err.code === 'ETIMEDOUT' ||
```
sfc-gh-pfus, line 32: Can we split it into two lines? Each alternative branch in a new line.

sfc-gh-pfus (approved):

lib/services/large_result_set.js
sfc-gh-dprzybysz: yes, we can

sfc-gh-pmotacki (approved):
lgtm

PR #647, opened 2023-09-24
Query Context Cache integration testing fix

sfc-gh-ext-simba-jy:
### Description
- [x]  I modified the test to generate random DB names in the test.

### Checklist

### DIFF
--- a/test/integration/testHTAP.js
+++ b/test/integration/testHTAP.js
@@ -7,24 +7,36 @@ const async = require('async');
 const connOption = require('./connectionOptions').valid;
 const testUtil = require('./testUtil');
 
-// Only the AWS servers support the hybrid table in the GitHub action.
+function getRandomDBNames() {
+  const dbName = 'qcc_test_db';
+  const arr = [];
+  const randomNumber = Math.floor(Math.random() * 10000);
+  for (let i = 0; i < 3; i++){
+    arr.push(dbName + (randomNumber + i));
+  }
+  return arr;
+}
+
 if (process.env.CLOUD_PROVIDER === 'AWS') {
   describe('Query Context Cache test', function () {
+    this.retries(3);
     let connection;
-    
-    before(async () => {
+    const dbNames = getRandomDBNames(); 
+
+    beforeEach(async () => {
       connection = testUtil.createConnection(connOption);
-      await testUtil.connectAsync(connection);
+      testUtil.connectAsync(connection);
     });
 
     after(async () => {
+      await testUtil.dropDBsIgnoringErrorsAsync(connection, dbNames);
       await testUtil.destroyConnectionAsync(connection);
     });
 
     const querySet = [
       {
         sqlTexts: [
-          'create or replace database db1',
+          `create or replace database ${dbNames[0]}`,
           'create or replace hybrid table t1 (a int primary key, b int)',
           'insert into t1 values (1, 2), (2, 3), (3, 4)'
         ],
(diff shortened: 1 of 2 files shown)


test/integration/testUtil.js
```diff
@@ -101,6 +101,23 @@ module.exports.dropTablesIgnoringErrorsAsync = async (connection, tableNames) =>
...
 };
 
+/**
+ * Drop databases one by one if exist - any connection error is ignored
+ * @param connection Connection
+ * @param dbNames string[]
+ * @return {Promise<void>}
+ */
+module.exports.dropDBsIgnoringErrorsAsync = async (connection, dbNames) => {
+  for (const dbIdx in dbNames) {
```
sfc-gh-dprzybysz, line 111: let's prefer `dbNames.forEach` and don't use index when unnecessary

test/integration/testHTAP.js
```diff
@@ -7,50 +7,63 @@ const async = require('async');
 const connOption = require('./connectionOptions').valid;
 const testUtil = require('./testUtil');
 
+function getRandomDBNames() {
+  const dbName = 'qcc_test_db';
+  const arr = [];
+  const randomNumber = Math.floor(Math.random() * 10000);
```
sfc-gh-dprzybysz, line 13: instead of taking generating random number we can use uuid as a suffix - it should be random enough

sfc-gh-dprzybysz (approved):

test/integration/testHTAP.js
```diff
@@ -7,50 +7,63 @@ const async = require('async');
 const connOption = require('./connectionOptions').valid;
 const testUtil = require('./testUtil');
 
+function getRandomDBNames() {
```
sfc-gh-pmotacki, line 10: What do you think about creating common method in testUtil.js - with functionality of generating and concatenate suffix with original db name?

test/integration/testHTAP.js
sfc-gh-ext-simba-jy: Using uuid instead.

sfc-gh-ext-simba-jy: It seems like uuid has a problem in order to use it as a dbName. db663b9a3d-40a0-4baf-90c1-aaaaed9fb86d: {"name":"OperationFailedError","code":"001003","message":"SQL compilation error:\nsyntax error line 1 at position 43 unexpected '-'.","sqlState":"42000","data":{"internalError":false,"errorCode":"001003","age":0,"sqlState":"42000","queryId":"01af3b8e-0506-b1e4-0000-53b1070a75ce","line":-1,"pos":-1,"type":"COMPILATION"}}

test/integration/testUtil.js
sfc-gh-ext-simba-jy: My codes are async and it seems like async is not working with forEach. https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971. Instead of  the foreach, I used map function isntead.

test/integration/testHTAP.js
sfc-gh-ext-simba-jy: It seems like Dominik wants to use a uuid, but it does not work with the db or table Name because of the hyphen(-) in the uuid. If we need a function to create a random number as a suffix with the original name, I will create it.

sfc-gh-pmotacki: We can add this function next time

sfc-gh-pmotacki (approved):

PR #649, opened 2023-09-25
SNOW-922989 Replace deprecated bignumber.js floor function

mblakley:
### Description
Replaced the deprecated `floor` function in new versions of bignumber.js (>= 6.0) with the updated solution of calling `.integerValue(BigNumber.ROUND_FLOOR)`

### Checklist
- [X] Format code according to the existing code style (run `npm run lint:check -- CHANGED_FILES` and fix problems in changed code) - I ran this, and there are a lot of warnings in the existing column.js file.
- [X] Make all tests (unit and integration) pass (`npm run test:unit` and `npm run test:integration`) - Ran unit tests, no new failures

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "asn1.js-rfc5280": "^3.0.0",
     "axios": "^1.5.1",
     "big-integer": "^1.6.43",
-    "bignumber.js": "^6.0.0",
+    "bignumber.js": ">=6.0.0",
     "binascii": "0.0.2",
     "bn.js": "^5.2.1",
     "browser-request": "^0.3.3",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "asn1.js-rfc5280": "^3.0.0",
     "axios": "^1.5.1",
     "big-integer": "^1.6.43",
-    "bignumber.js": ">=6.0.0",
+    "bignumber.js": "^9.1.2",
     "binascii": "0.0.2",
     "bn.js": "^5.2.1",
     "browser-request": "^0.3.3",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "asn1.js-rfc5280": "^3.0.0",
     "axios": "^1.5.1",
     "big-integer": "^1.6.43",
-    "bignumber.js": "^2.4.0",
+    "bignumber.js": "^6.0.0",
     "binascii": "0.0.2",
     "bn.js": "^5.2.1",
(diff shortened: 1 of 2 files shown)


mblakley:

mblakley:
Thank you for reviewing! Another possible solution would be to set an upper bound on the bignumber.js version in package.json - my initial thought was that my proposed solution would allow moving to the newest versions of that package instead of sticking with the older versions.

sfc-gh-pbulawa:
Hi @mblakley,

Thanks for your contribution! I have ran your branch in #656 and it seems that the newer version of the `bignumber.js` is having different API:
```javascript
Uncaught TypeError: valFracSecsWithTzBig.greaterThanOrEqualTo is not a function
      at Object.convertRawTimestampTz [as convert] (/mnt/host/lib/connection/result/column.js:462:30)
```

Could you please take a look into it and update the code?

mblakley:
I made updates to my branch.

Luckily, the API name updates for `greaterThanOrEqualTo` were also done in v6.0 of the bignumber.js package, so there's no conflict with using that version as the minimum.

package.json
```diff
@@ -12,7 +12,7 @@
     "asn1.js-rfc5280": "^3.0.0",
     "axios": "^1.5.0",
     "big-integer": "^1.6.43",
-    "bignumber.js": "^2.4.0",
+    "bignumber.js": "^6.0.0",
```
sfc-gh-pmotacki, line 15: Current version of bignumber.js is 9.1.2. Should we set this version in package.json?

sfc-gh-pmotacki (changes requested):

package.json
mblakley: It doesn't need to be set to the latest version, but could be.  I actually think setting this to ">=" would be better, since it would allow for major versions beyond 6.  I'll make that change.

sfc-gh-pmotacki: Do we really need to support 4 versions? In my opinion during upgrading more interesting for us is confirm of using version 9. It is better for supporting in future. My proposal is to use current latest version (9.1.2) with possibility tu upgrade patch and minor -> ^9.1.2. WDYT?

mblakley: Fine with me, should make testing and support easier - updated.  Thanks again!

sfc-gh-pmotacki: Thank you

sfc-gh-pmotacki (approved):

sfc-gh-pmotacki:
@mblakley thank you for your contribution.

PR #663, opened 2023-10-10
SNOW-824882 MFA token cache

sfc-gh-ext-simba-jy:
### Description
testing storage secure new library (keytar)

### Checklist


lib/constants/error_messages.js
```diff
@@ -71,6 +71,7 @@ exports[404043] = 'Invalid clientConfigFile value. The specified value must be a
 exports[404044] = 'Invalid retryTimeout value. The specified value must be a number.';
 exports[404045] = 'Invalid account. The specified value must be a valid subdomain string.';
 exports[404046] = 'Invalid region. The specified value must be a valid subdomain string.';
+exports[404047] = 'Invalid clientStoreTemporaryCredential. The specified value must be a boolean.';
```
sfc-gh-ext-simba-hx, line 74: Sounds weird to throw error on this and it makes more sense to me to fallback to external browser authentication. Have you checked the behavior of other drivers?

lib/authentication/secureStorage.js
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2015-2023 Snowflake Computing Inc. All rights reserved.
+ */
+
+const path = require('path');
+const GlobalConfig = require('../global_config');
+const Logger = require('../logger');
+const keytar = require('keytar');
```
sfc-gh-ext-simba-hx, line 8: just checking this line is fine when we don't have the necessary libs on Linux?

lib/constants/error_messages.js
sfc-gh-ext-simba-jy: Hmmm, I see. I am not sure other drivers have the same checking argument function. The reason why I put this error message is the node js driver usually creates an error message for each connection configuration. If you have a look at https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/connection/connection_config.js#L214,  the driver has validated for each configuration and they have each error code.

lib/authentication/secureStorage.js
sfc-gh-ext-simba-jy: I redesigned the secure Storage. Please have a look again.

lib/constants/error_messages.js
sfc-gh-ext-simba-hx: Ah I see. OK then.

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+async function writeCredential(host, user, credType, token) {
+  if (checkForNull(host, user, credType, token)) {
+    Logger.getInstance().debug('Token cannot be saved. One of the paramenters is null or undefined');
+    return;
+  } else {
+    try {
+      try {
+        await secureStorage.writeCredential(host, user, credType, token);
```
sfc-gh-ext-simba-hx, line 28: Check OS here and call secure/local storage accordingly.
https://stackoverflow.com/questions/8683895/how-do-i-determine-the-current-operating-system-with-node-js
No local storage on Win/Mac and if anything wrong we should fallback to use browser authentication.

sfc-gh-ext-simba-jy: I fixed it.

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {String} credType 
+ * @returns 
+ */
+
+function checkForNull(...parameters){
+  return parameters.some((element) => element === undefined || element === null);
+}
+
+function isLinuxMachine() {
+  return os.platform() !== 'win32' || os.platform() !== 'darwin';
```
sfc-gh-dheyman-1, line 23: this should be `&&` instead of `||`

lib/services/sf.js
```diff
@@ -1214,6 +1232,16 @@ StateConnecting.prototype.continue = function ()
       // update all token-related information
       parent.tokenInfo.update(body.data);
 
+      if (connectionConfig.getClientStoreTemporaryCredential()) {
+        if(body.data.idToken){
+          CredentialManager.writeCredential(connectionConfig.account, body.data.displayUserName,
```
sfc-gh-dheyman-1, line 1237: We should pass username from `connectionConfig`, because it's the actual source of the username that we use when we are reading the credential. We can't rely on `displayUserName` being the same as config username, because it's a field in Snowflake that can be set to whatever the customer wants.

package.json
```diff
@@ -36,6 +36,9 @@
     "uuid": "^8.3.2",
     "winston": "^3.1.0"
   },
+  "optionalDependencies": {
```
sfc-gh-dheyman-1, line 39: what will happen if user installs using `npm install --omit=optional`? Do we handle such a case?

lib/services/sf.js
sfc-gh-ext-simba-jy: Okay, I will fix them. Then, the driver will not save the token if users do not provide the username. Could you please confirm that this is the correct design?

lib/authentication/SecureStorage/credentialManager.js
sfc-gh-ext-simba-jy: Fixed.

package.json
sfc-gh-ext-simba-jy: I added try catch just in case when users do not install the 'keytar' library. In this case, the driver returns null, and they can see the warning message in the driver logs.

sfc-gh-ext-simba-jy: As we decided not to use 'keytar' library, this conversation was resolved.

lib/services/sf.js
sfc-gh-ext-simba-jy: Fixed

lib/global_config.js
```diff
@@ -278,4 +279,8 @@ exports.getKeepAlive = function () {
   return keepAlive;
 };
 
+exports.CredentialManager = CredentialManager;
```
sfc-gh-dheyman-1, line 283: I think the approach we are looking for is to simplify this a little. Our global config logic could look more like this: 
```
let credentialManager = createJsonFileCredentialManager();

exports.setCustomCredentialManager = function (customCredentialManager) {
  // here we should also add checks if the object actually has these read(), write(), remove() methods. If not then we should throw an error.
  credentialManager = customCredentialManager;
};

exports.getCredentialManager = function () {
  return credentialManager;
};
```
Also note that this method `createJsonFileCredentialManager()` creates a default manager (it's logic currently is in credentialManager.js.

lib/authentication/SecureStorage/customCredentialManager.js
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2015-2023 Snowflake Computing Inc. All rights reserved.
+ */
+
+function CustomCredentialManager(storage) {
```
sfc-gh-dheyman-1, line 5: this object creates unnecessary layer of abstraction so I think we can get rid of this file. Also our internal configuration parameters (like host or credType) shouldn't leak to a manager supplied by user - instead we should have as simple API as possible, hence instead of `read(host, user, credType)` we should have something more like `read(key)`

lib/global_config.js
sfc-gh-dheyman-1: So in effect user would set the manager something like this (forgive me my pseudocode skills):
```
setCustomCredentialManager(function () {
  this.read = function (key) {
    return keytar.read(key);
  };
  this.write = function (key, credential) { ... }
  this.remove = function (key) { ... }
});
```
That being said, obviously our default manager set via `createJsonFileCredentialManager()` should have the same interface so they can be used interchangeably.

lib/services/sf.js
```diff
@@ -694,6 +695,17 @@ function StateAbstract(options)
             if (body && !body.success)
             {
               var data = body.data;
+              const auth = getAuthenticator(connectionConfig, httpClient);
+
+              if(body.code === '390195' && data.authnMethod === 'TOKEN') {
+                Logger.getInstance().trace("ID Token being used has expired. Reauthenticating");
+                const {CredentialManager} = require('../global_config');
+                CredentialManager.remove(connectionConfig.account, connectionConfig.username, authenticationTypes.ID_TOKEN_AUTHENTICATOR);
```
sfc-gh-dheyman-1, line 703: After changes in global_config.js, here instead we can have:
```
let key = buildCredentialCacheKey(connectionConfig.account, connectionConfig.username, authenticationTypes.ID_TOKEN_AUTHENTICATOR); //this method can be put in `util.js` or wherever you think is appropriate :) 
GlobalConfig.getCredentialManager().remove(key)
```

sfc-gh-ext-simba-jy: Fixed.

lib/authentication/SecureStorage/customCredentialManager.js
sfc-gh-ext-simba-jy: Fixed

lib/global_config.js
sfc-gh-ext-simba-jy: Replaced. I have a question about this. For the local JSON credential, I still left two different files: 'localStorage' and 'credentialManager' just in case when we add different credentialManager methods in the future. However, for now, do you think it would be better to merge those two files?

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * 
+ * @param {String} host 
+ * @param {String} user 
+ * @param {String} credType 
+ * @returns 
+ */
+
+function checkForNull(...parameters){
```
sfc-gh-dheyman-1, line 16: can this function be extracted to util? Seems like something that can be reused throughout project.

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,58 @@
+/*
```
sfc-gh-dheyman-1, line 1: I think it's good idea change the name to jsonCredentialManager.js and here import localStorage.js logic

lib/global_config.js
```diff
@@ -278,4 +278,29 @@ exports.getKeepAlive = function () {
   return keepAlive;
 };
 
+let credentialManager = createJsonFileCredentialManager();
+
+exports.setCustomCredentialManager = function (customCredentialManager) {
+  Errors.checkArgumentValid(Util.checkValidCustomCrednetialManager(customCredentialManager),
```
sfc-gh-dheyman-1, line 284: typo `Crednetial` :)

lib/global_config.js
```diff
@@ -278,4 +278,29 @@ exports.getKeepAlive = function () {
...
+      await customCredentialManager.remove(key);
+    };  
+  };  
+};
+
+exports.getCredentialManager = function () {  
+  return credentialManager;  
+};  
 
+function createJsonFileCredentialManager() {
```
sfc-gh-dheyman-1, line 304: Now that I think about it, IMO this would be cleaner: 
```
function createDefaultFileCredentialManager() {
  return require('./authentication/SecureStorage/jsonCredentialManager');
} 
```

lib/global_config.js
```diff
@@ -278,4 +278,29 @@ exports.getKeepAlive = function () {
...
 };
 
+let credentialManager = createJsonFileCredentialManager();
+
+exports.setCustomCredentialManager = function (customCredentialManager) {
+  Errors.checkArgumentValid(Util.checkValidCustomCrednetialManager(customCredentialManager),
+  ErrorCodes.ERR_GLOBAL_CONFIGURE_INVALID_CUSTOM_CREDENTIAL_MANAGER);
+  
+  credentialManager =  new function () {  
+    this.read = async function (key) {
```
sfc-gh-dheyman-1, line 288: why are we wrapping these functions instead of just assigning the whole object?

lib/global_config.js
sfc-gh-dheyman-1: yup, we can merge it.

lib/core.js
```diff
@@ -222,6 +222,14 @@ function Core(options)
 
           GlobalConfig.setKeepAlive(keepAlive);
         }
+
+        let customCredentialManager = options.customCredentialManager;
```
sfc-gh-dheyman-1, line 226: i assume this is to cover a scenario where user sets it via connectionConfig(options) instead of GlobalConfig?

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,62 @@
...
+  } 
+  try {
+    localStorage.deleteCredential(host, user, credType);
+  } catch (err){
+    Logger.getInstance().error('Failed to delete Credential: ', err.message);
+  } 
+}
+
+module.exports = {
+  checkForNull,
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I exported this for the testing purpose.

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,62 @@
...
+    return null;
+  } 
+  try {
+   return localStorage.readCredential(host, user, credType);
+  } catch (err){
+    Logger.getInstance().error('Failed to read Credential: ', err.message);
+  } 
+}
+
+function remove(host, user, credType) {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: This is already fixed.

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,62 @@
...
+  } else {
+    try {
+      localStorage.writeCredential(host, user, credType, token);
+    } catch (err){
+      Logger.getInstance().error('Failed to save Credential: ', err.message);
+    } 
+  }
+}
+
+function read(host, user, credType) {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: fixed.

lib/authentication/SecureStorage/credentialManager.js
sfc-gh-ext-simba-jy: Fixed.

lib/core.js
sfc-gh-ext-simba-jy: Do you mean this option needs to go to the connection config?

lib/global_config.js
sfc-gh-ext-simba-jy: I merged it.

ci/image/Dockerfile.nodejs-centos7-node14-test
```diff
@@ -4,7 +4,7 @@ FROM centos:7
 # update OS
 RUN yum -y update && \
     yum -y install epel-release && \
-    yum -y install centos-release-scl
+    yum -y install centos-release-scl
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: There was a space. I removed it.

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,62 @@
+/*
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Replaced.

lib/authentication/SecureStorage/credentialManager.js
sfc-gh-ext-simba-jy: I changed the name 'checkForNull' to 'checkParametersDefined', and moved to the Util.js.

lib/authentication/SecureStorage/credentialManager.js
sfc-gh-ext-simba-jy: Done, and merged.

lib/global_config.js
sfc-gh-ext-simba-jy: Done.

lib/global_config.js
sfc-gh-ext-simba-jy: I did this because users can use the async functions. However, as you mentioned, it would be better to assign the whole object and I added the 'await' command on the connection and the sf.

lib/core.js
sfc-gh-dheyman-1: Sorry, my bad - misunderstood this code. We don't want to do it via connection config.

lib/global_config.js
```diff
@@ -256,4 +260,19 @@ exports.getKeepAlive = function () {
...
+  ErrorCodes.ERR_GLOBAL_CONFIGURE_INVALID_CUSTOM_CREDENTIAL_MANAGER);
+  
+  credentialManager =  customCredentialManager;
+};
+
+exports.getCredentialManager = function () {  
+  return credentialManager;  
+};  
 
+function createJsonFileCredentialManager() {
```
sfc-gh-dheyman-1, line 276: what do you think about `createDefaultCredentialManager()`?

lib/global_config.js
```diff
@@ -114,9 +115,12 @@ exports.getOcspResponseCacheMaxAge = function () {
  *
  * @returns {string}
  */
-exports.mkdirCacheDir = function () {
-  let cacheRootDir = process.env.SF_OCSP_RESPONSE_CACHE_DIR;
-  if (!Util.exists(cacheRootDir)) {
+
+exports.mkdirCacheDir = function (envPath)
```
sfc-gh-dheyman-1, line 119: using this logic with tokens creates a security vulnerability, since we don't want to store tokens in /tmp dir that can be accessed also by other users. Our logic should be that the directory can be only either `homedir()`or the one specified by the user (`envPath`), otherwise the feature should be disabled.

sfc-gh-dheyman-1 (dismissed):

lib/global_config.js
sfc-gh-ext-simba-jy: More reasonable.

lib/services/sf.js
```diff
@@ -623,6 +625,18 @@ function StateAbstract(options) {
             // success flag is false, the operation we tried to perform failed
             if (body && !body.success) {
               var data = body.data;
+              const auth = getAuthenticator(connectionConfig, httpClient);
+
+              if (body.code === '390195' && data.authnMethod === 'TOKEN') {
```
sfc-gh-dheyman-1, line 630: Can we extract this magic number? (390195)

test/integration/testManualConnection.js
```diff
@@ -70,6 +71,38 @@ if (process.env.RUN_MANUAL_TESTS_ONLY == 'true') {
...
+        //Testing  with the id token.
+        const idTokenConnection = snowflake.createConnection(connectionOption);
+        idTokenConnection.connectAsync(function (err) {
+          assert.ok(!err);
+        });
+        await testUtil.destroyConnectionAsync(idTokenConnection);
+
+        //Testing reauthentication.
+        await GlobalConfig.getCredentialManager().write(connectionOption.host, connectionOption.username, 'ID_TOKEN", "WRONG Token');
+        const wrongTokneConnection = testUtil.connectAsync(connOption);
```
sfc-gh-dheyman-1, line 100: typo

lib/global_config.js
sfc-gh-ext-simba-jy: The reason why I stored the token in this location is because this location is the place where we are saving the ocsp response cache file. Is the reason why we need to save it in homedir() or envPath because the 'securestorage' is more secure than ocsp_response_cache? If you confirm, I will redesign this.

test/unit/authentication/authentication_test.js
```diff
@@ -25,6 +25,7 @@ var connectionOptionsKeyPair = mockConnectionOptions.authKeyPair;
 var connectionOptionsKeyPairPath = mockConnectionOptions.authKeyPairPath;
 var connectionOptionsOauth = mockConnectionOptions.authOauth;
 var connectionOptionsOkta = mockConnectionOptions.authOkta;
+const connectionOptionsidToken = mockConnectionOptions.authIdToken;
```
sfc-gh-dheyman-1, line 28: connectionOptionsIdToken

lib/global_config.js
sfc-gh-dheyman-1: Also it would be great to create unit test that checks what is the actual cache directory

sfc-gh-dheyman-1: I think token leak introduces much greater risk, but we should consider to exclude shared directories from  ocsp response cache too.

lib/services/sf.js
sfc-gh-ext-simba-jy: I think this number is necessary. If I remove it, it impacts other token authentication or workflow. When I remove it and run a unit testing, the testing is stuck. I tried other properties from the response, but the code '390195) is unique.

test/unit/authentication/authentication_test.js
sfc-gh-ext-simba-jy: Fixed

test/integration/testManualConnection.js
sfc-gh-ext-simba-jy: Fixed

lib/global_config.js
sfc-gh-ext-simba-jy: Fixed

lib/global_config.js
sfc-gh-ext-simba-jy: I redesigned it. Instead of using mkdirCachedir in the global config, I make the default credential manager use home dir or env path. Because I need to check that either homedir() or the one specified is valid, I redesigned the json credential manager as a class and I implemented that it checks that token cache directory is whether valid or not. If it is not valid, all the methods (write, read, remove) just return null. Also, I added the unit testing for the token directory.

lib/services/sf.js
sfc-gh-ext-simba-jy: Okay, it is fixed.

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,91 @@
...
+ * Copyright (c) 2015-2023 Snowflake Computing Inc. All rights reserved.
+ */
+
+const path = require('path');
+const Logger = require('../../logger');
+const fs = require('fs');
+const os = require('os');
+const Util = require('../../util');
+
+function JsonCrednetialManager() {
```
sfc-gh-dheyman-1, line 11: typo

test/unit/util_test.js
```diff
@@ -890,5 +890,158 @@ describe('Util', function () {
...
+        }
+    
+        this.write = function (key ,credential) {
+        }
+    
+        this.remove = function (key) {
+        }
+    }
+    
+      const testCases = [
```
sfc-gh-dheyman-1, line 962: where are these testCases tested?

lib/global_config.js
```diff
@@ -256,4 +259,20 @@ exports.getKeepAlive = function () {
   return keepAlive;
 };
 
+let credentialManager = createDefaultCredentialManager();
+
+exports.setCustomCredentialManager = function (customCredentialManager) {
```
sfc-gh-dheyman-1, line 264: it would be great to simulate users behaviour and create a test where we supply our own  custom implementation of credential manager. To keep it simple, perhaps we could do an in-memory implementation of it?

sfc-gh-ext-simba-jy: In my opinion, the testing for this credential manager is the one I did not finish that you commented on. Those test cases are testing the valid credential manager and scenario of expected users input.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: Fixed

test/unit/util_test.js
sfc-gh-ext-simba-jy: Added

lib/global_config.js
sfc-gh-ext-simba-jy: I added testing for the custom credential manager. Please have a look at https://github.com/snowflakedb/snowflake-connector-nodejs/pull/663/files#diff-c92bed55d478910c2b3d0fa5930295b793b9d866ee42e7411ad6ac4fc15531adR12

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,91 @@
...
+    let jsonCredential = null;
+  
+    if (fs.existsSync(this.tokenDir)){
+      jsonCredential = this.readJsonCredentialFile() || {};
+    } else {
+      jsonCredential = {}; 
+    }
+    jsonCredential[key] = token;
+   
+    fs.writeFileSync(this.tokenDir, JSON.stringify(jsonCredential), 'utf8');
```
sfc-gh-dheyman-1, line 51: This is the place where we also create a file in case it doesn't exist? If so, then what are the read/write access rights of this file? It should only be accessed by the account user.

sfc-gh-ext-simba-jy: I added the mode which makes the account users only read and write.

test/unit/snowflake_config_test.js
```diff
@@ -20,7 +19,7 @@ describe('Snowflake Configure Tests', function () {
       ocspFailOpen: GlobalConfig.getOcspFailOpen(),
       keepAlive: GlobalConfig.getKeepAlive(),
       jsonColumnVariantParser: GlobalConfig.jsonColumnVariantParser,
-      xmlColumnVariantParser: GlobalConfig.xmlColumnVariantParser
+      xmlColumnVariantParser: GlobalConfig.xmlColumnVariantParser,
```
sfc-gh-pmotacki, line 22: It shouldn't be change

sfc-gh-pmotacki (changes requested):

lib/authentication/SecureStorage/credentialManager.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+const Logger = require('../../logger');
+const localStorage = require('./localStorage');
+
+/**
+ * 
+ * @param {String} host 
+ * @param {String} user 
+ * @param {String} credType
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Not required anymore as we changed the design.

lib/authentication/SecureStorage/localStorage.js
```diff
@@ -0,0 +1,66 @@
...
+function readCredential(key) {
+  if (!fs.existsSync(getCredentialCacheDir())){
+    Logger.getInstance().info('Cannot find the credential file from the directory');
+    return null;
+  }
+  const jsonCredential = readCredentialFile();
+  return jsonCredential[key] || null;
+}
+
+function deleteCredential(key) {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Not required anymore as we changed the design.

lib/authentication/SecureStorage/credentialManager.js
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Outdated.

lib/authentication/SecureStorage/localStorage.js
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Outdated.

package.json
```diff
@@ -81,4 +81,4 @@
     "url": "https://www.snowflake.com/"
   },
   "license": "Apache-2.0"
-}
+}
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: There is an extra space at the end of the file, but I am not sure we really need it. I know we can fix it later, not in this PR.

lib/authentication/auth_idtoken.js
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2015-2023 Snowflake Computing Inc. All rights reserved.
+ */
+
+const AuthWeb = require('./auth_web');
+
+/**
+ * Creates an oauth authenticator.
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Fixed.

lib/authentication/auth_default.js
```diff
@@ -21,7 +21,7 @@ function AuthDefault(password) {
   this.updateBody = function (body) {
     body['data']['PASSWORD'] = password;
   };
-
+
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I reverted back.

lib/global_config.js
```diff
@@ -256,4 +257,21 @@ exports.getKeepAlive = function () {
...
 };
 
+let credentialManager = createDefaultCredentialManager();
 
+exports.setCustomCredentialManager = function (customCredentialManager) {
+  Errors.checkArgumentValid(Util.checkValidCustomCredentialManager(customCredentialManager),
+    ErrorCodes.ERR_GLOBAL_CONFIGURE_INVALID_CUSTOM_CREDENTIAL_MANAGER);
+  
+  credentialManager =  customCredentialManager;
+  Logger.getInstance().debug('Custom credential manager is set by a user.');
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Fixed.

lib/services/sf.js
```diff
@@ -623,6 +625,18 @@ function StateAbstract(options) {
...
+                Logger.getInstance().trace('ID Token being used has expired. Reauthenticating');
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Fixed.

test/unit/snowflake_config_test.js
sfc-gh-ext-simba-jy: Could you please confirm whether we need to do the trailing comma or not? Dominik said we need to do it ,so that is why I put the comma at the end. I edited it first.

lib/authentication/auth_web.js
```diff
@@ -48,7 +48,9 @@ function AuthWeb(connectionConfig, httpClient, webbrowser) {
    *
    * @returns {null}
    */
+
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: removed.

lib/authentication/authentication.js
```diff
@@ -1,12 +1,14 @@
 /*
- * Copyright (c) 2015-2021 Snowflake Computing Inc. All rights reserved.
+ * Copyright (c) 2015-2023 Snowflake Computing Inc. All rights reserved.
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: If we need to change the copyright date for all the files, please let me know. I will make a different PR.

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright (c) 2015-2023 Snowflake Computing Inc. All rights reserved.
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: This was hold for few months, so this was outdated. I fixed it and let me know if we need to change all the files.

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,91 @@
...
+const os = require('os');
+const Util = require('../../util');
+
+function JsonCredentialManager() {
+  this.getTokenDir = function () {
+    let tokenDir = process.env.SF_TEMPORARY_CREDENTIAL_CACHE_DIR;
+    if (!Util.exists(tokenDir)) {
+      tokenDir = os.homedir();
+    }
+    if (!Util.exists(tokenDir)) {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I added at https://github.com/snowflakedb/snowflake-connector-nodejs/pull/663/files#diff-3181d20ffd0eab4ed1603db94abce7ca627cbb18905cabd59aef05ac4ef1e431R33.

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,91 @@
...
+      Logger.getInstance().error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location(home). 
+      Please assign the enviroment variable value SF_TEMPORARY_CREDENTIAL_CACHE_DIR to enable the default credential manager.`);
+      return null; 
+    }
+    const tokenCacheFile = path.join(tokenDir, 'temporary_credential.json');
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: This might be outdated. According to the security review, we tried to save the token file in the home directory, so we do not need to add a username for the file unless the customer uses the same space.

sfc-gh-ext-simba-jy:
This PR has been open for a long time with many comments, so I made a new PR : https://github.com/snowflakedb/snowflake-connector-nodejs/pull/820

PR #669, opened 2023-10-16
SNOW-926029: Check for renewing/connecting state when validating a connection

sfc-gh-ext-simba-lf:
### Description
Regarding issue 678

The PR checks if a connection is renewing/connecting to determine if it is valid. This can prevent an `Already connected` error or `Connection already in progress` error which happens when a user calls the `isUp` or `isValidAsync` function and it returns false and the user then tries to connect manually while the connection is already renewing/connecting

### Checklist

### DIFF
--- a/lib/services/sf.js
+++ b/lib/services/sf.js
@@ -486,7 +486,9 @@ function SnowflakeService(connectionConfig, httpClient, config)
   };
 
   this.isConnected = function () {
-    return currentState === stateConnected || currentState === stateRenewing;
+    return currentState === stateConnected ||
+      currentState === stateConnecting ||
+      currentState === stateRenewing;
   };
 
   this.getServiceName = function ()
--- a/lib/services/sf.js
+++ b/lib/services/sf.js
@@ -485,8 +485,7 @@ function SnowflakeService(connectionConfig, httpClient, config)
     operationQueue.length = 0;
   };
 
-  this.isConnected = function ()
-  {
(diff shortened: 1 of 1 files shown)


lib/services/sf.js
```diff
@@ -485,9 +485,8 @@ function SnowflakeService(connectionConfig, httpClient, config)
     operationQueue.length = 0;
   };
 
-  this.isConnected = function ()
-  {
-    return currentState === stateConnected;
+  this.isConnected = function () {
+    return currentState === stateConnected || currentState === stateRenewing;
```
sfc-gh-pmotacki, line 489: It is correct. I'm still wondering if state 'Connetcing' should be added too? WDYT?

sfc-gh-pmotacki (changes requested):

lib/services/sf.js
sfc-gh-ext-simba-lf: It looks like `stateConnecting` occurs from `statePristine`. And `statePristine` occurs when there's no tokens like the first time connect is called or after the connection is terminated

So it's possible the user checks if the connection is valid while it's already connecting. And if they try to manually connect while it's in the connecting state, they'll get an error as well so I think we can add `stateConnecting` to prevent that

sfc-gh-pmotacki (commented):
LGTM

sfc-gh-pmotacki (approved):

PR #697, opened 2023-11-10
SNOW-926008 observability enhancement for proxy use-cases

sfc-gh-dszmolka:
* detecting if HTTP_PROXY/HTTPS_PROXY environmental variable is set, log them on DEBUG loglevel
* added logging for proxy port and proxy user (if set), had only proxy host
* if both HTTPS_PROXY and proxyHost is set, can lead to not-so-trivial-to-debug situations ; log this on DEBUG loglevel for more intuitive troubleshooting
* generic observability: instead of swallowing the underlying error upon connectivity issue and emit a hard-to-debug generic error message, log that underlying error stack on DEBUG loglevel


lib/agent/https_proxy_ocsp_agent.js
```diff
@@ -98,7 +98,25 @@ function connect(req, opts, fn)
...
+  if (process.env.HTTP_PROXY || process.env.HTTPS_PROXY) {
+    envHttpProxy = process.env.HTTP_PROXY ? 'HTTP_PROXY: ' + process.env.HTTP_PROXY : 'HTTP_PROXY: <unset>'
+    envHttpsProxy = process.env.HTTPS_PROXY ? 'HTTPS_PROXY: ' + process.env.HTTPS_PROXY : 'HTTPS_PROXY: <unset>'
+    envNoProxy = process.env.NO_PROXY ? 'NO_PROXY: ' + process.env.NO_PROXY : 'NO_PROXY: <unset>'
+    Logger.getInstance().debug('PROXY environment variables: %s %s %s'
+        , envHttpProxy, envHttpsProxy, envNoProxy)
+  }
+
+  // log proxy host, port, and user for observability
+  let prxPort = proxy.port ? ':' + proxy.port : ''
```
sfc-gh-dprzybysz, line 112: let's prefer `const`s where it is possible

lib/agent/https_proxy_ocsp_agent.js
```diff
@@ -98,7 +98,25 @@ function connect(req, opts, fn)
 {
   var proxy = this.proxy;
   var agent = this;
-  Logger.getInstance().debug("Using proxy=%s for host %s", proxy.host, opts.host);
+
+  // try and detect if there's proxy envvar set, currently it's not logged anywhere
+  if (process.env.HTTP_PROXY || process.env.HTTPS_PROXY) {
+    envHttpProxy = process.env.HTTP_PROXY ? 'HTTP_PROXY: ' + process.env.HTTP_PROXY : 'HTTP_PROXY: <unset>'
```
sfc-gh-dprzybysz, line 104: we should declare new variables with let/const

lib/agent/https_proxy_ocsp_agent.js
```diff
@@ -98,7 +98,25 @@ function connect(req, opts, fn)
 {
   var proxy = this.proxy;
   var agent = this;
-  Logger.getInstance().debug("Using proxy=%s for host %s", proxy.host, opts.host);
+
+  // try and detect if there's proxy envvar set, currently it's not logged anywhere
```
sfc-gh-dprzybysz, line 102: do we support env variables for proxy in NodeJs driver?

sfc-gh-dszmolka: we do not, to my knowledge. 
that does not keep users from trying to still use it (most of the times inadvertently, e.g. it's already set in the env for other application. or trying to use it since our other drivers support it), then get in all sorts of trouble, ending up with Snowflake Support who now doesn't have a really easy way of confirming this

lib/agent/https_proxy_ocsp_agent.js
sfc-gh-dszmolka: ok

lib/agent/https_proxy_ocsp_agent.js
sfc-gh-dszmolka: good point, thanks

test/integration/testHTAP.js
```diff
@@ -80,6 +80,9 @@ if (process.env.CLOUD_PROVIDER === 'AWS') {
               connection.execute({
                 sqlText: sqlTexts[k],
                 complete: function (err) {
+                  if (err) {
+                    console.log(err)
```
sfc-gh-pfus, line 84: I believe linter is right here :)

sfc-gh-dszmolka: fully agree and this is not part of the original PR or my intention and will be removed from the final version
i really just wanted to see why the HTAP tests break and if them breaking is anyhow related to my change. (it is not, 'create hybrid table' fails with Snowflake internal crash, which I don't think is related to this client side logging PR)

sfc-gh-dszmolka: and reverted

lib/agent/https_proxy_ocsp_agent.js
sfc-gh-dprzybysz: should we do anything more than only reading it? eomeone who reads logs will see that maybe there is some difference between those options and will it be enough?

Here we are using agent that is aware of proxy so we can see only differences between driver and env configuration - maybe checking proxy env variables should also be done when we don't have proxy configuration on connection?

sfc-gh-dszmolka: thank you - i'll work on this per our discussion

sfc-gh-dszmolka: re: the env variables: even though it is not documented yet, my tests show that we do support it. I mean the connection breaks as soon as I point HTTP_PROXY / HTTPS_PROXY to a non-existing proxy. It can be confirmed it _does_ go through the proxy as soon as I set up a proxy and set HTTPS_PROXY to point the driver towards it.

Even without proxyHost/proxyPort.

sfc-gh-dszmolka: PROXY envvar now also should be detected outside of the agent

lib/agent/https_proxy_ocsp_agent.js
```diff
@@ -98,7 +99,33 @@ function connect(req, opts, fn)
...
+
+  const prxPort = proxy.port ? ':' + proxy.port : '';
+  const prxUser = proxy.auth ? ' (user: ' + proxy.auth.split(':')[0] + ')' : '';
+  const prxHostWithPort = proxy.host + prxPort;
+  const envProxy = Util.getEnvProxy();
+
+  // using both the envvar and the proxyHost/proxyPort settings for proxy, this might conflict
+  if (envProxy.httpProxy &&
+      Util.removeScheme(envProxy.httpProxy).toLowerCase() != prxHostWithPort.toLowerCase()) {
+    Logger.getInstance().debug('Using both the HTTP_PROXY (%s) and the proxyHost:proxyPort (%s) settings to connect, '
```
sfc-gh-dprzybysz, line 111: when it's different maybe let's log it with WARN, also in other added logs

lib/services/sf.js
```diff
@@ -615,6 +615,9 @@ function StateAbstract(options)
           // if we got an error, wrap it into a network error
           if (err)
           {
+            // if we're running in DEBUG loglevel, probably we want to see the full error instead
+            Logger.getInstance().debug(JSON.stringify(err, Util.getCircularReplacer()))
```
sfc-gh-dprzybysz, line 619: could we add a prefixing message at the begining of this log? to not only log the reasult error as json

lib/services/sf.js
```diff
@@ -1264,6 +1267,13 @@ StateConnecting.prototype.continue = function ()
   // issue a login request
   const sendRequest = function ()
   {
+    // try to detext PROXY envvar even outside of agent
```
sfc-gh-dprzybysz, line 1270: we can detect it once when agent is created - agents are reused now

lib/agent/https_proxy_ocsp_agent.js
sfc-gh-dszmolka: ok

lib/services/sf.js
sfc-gh-dszmolka: ok

sfc-gh-dszmolka:
closing this PR for now until i clear some test locally

PR #699, opened 2023-11-14
SNOW-926008 supportability/observability enhancement for generic and proxy use-cases

sfc-gh-dszmolka:
- generic observability: instead of swallowing the underlying error upon connectivity issue and emit a hard-to-debug generic error message, log that underlying error stack on DEBUG loglevel

- proxy: detecting if HTTP_PROXY/HTTPS_PROXY environmental variable is set, log them on DEBUG loglevel
- proxy: added logging for Connection's proxyPort and proxyUser (if set), had only proxyHost. 
-- this bit of logging now moved to generic agent
- proxy: if both HTTPS_PROXY and proxyHost is set, can lead to not-so-trivial-to-debug situations ; log this on WARN loglevel for more intuitive troubleshooting

### DIFF
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -130,7 +130,6 @@ NodeHttpClient.prototype.getAgent = function (parsedUrl, proxy, mock) {
   } else {
     agent = getFromCacheOrCreate(HttpAgent, agentOptions, parsedUrl);
   }
-
   return agent;
 };
 
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -53,6 +53,16 @@ function getFromCacheOrCreate(agentClass, options, parsedUrl) {
     const agent = agentClass(agentOptions);
     httpsAgentCache.set(agentId, agent);
     Logger.getInstance().trace(`Create and add to cache new agent ${agentId}`);
+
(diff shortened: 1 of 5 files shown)


lib/http/node.js
```diff
@@ -120,6 +126,31 @@ NodeHttpClient.prototype.getAgent = function (parsedUrl, proxy, mock) {
   } else {
     agent = getFromCacheOrCreate(HttpAgent, agentOptions, parsedUrl);
   }
+
+  // try to detect and log PROXY envvar
+  const envProxy = Util.getEnvProxy();
```
sfc-gh-dprzybysz, line 131: is this check still run for each request? maybe we can convert this block of code into a method and run when creating a new agent inside `getFromCacheOrCreate`?

sfc-gh-dszmolka: actually a good idea, wish this approach occurred to me sooner :) 
rewrote the proxy detection function and put the functionality into the new agent creation step so hopefully it only gets executed once, when the new agent is created

sfc-gh-dprzybysz (approved):

sfc-gh-pmotacki (approved):
LGTM

PR #702, opened 2023-11-20
SNOW-926008 observability improvements continued

sfc-gh-dszmolka:
* extra logging for errors and unsuccessful responses in requests when retrieving query result from cloud storage
* '2515 milliseconds' is enough precision in the logs for retries (was: '2515.1356764370476 milliseconds')

### DIFF
--- a/lib/services/large_result_set.js
+++ b/lib/services/large_result_set.js
@@ -71,8 +71,9 @@ function LargeResultSetService(connectionConfig, httpClient)
       if (err || isUnsuccessfulResponse(response))
       {
         // if we're running in DEBUG loglevel, probably we want to see the full error too
-        const logErr = err ? JSON.stringify(err, Util.getCircularReplacer()) 
-        : JSON.stringify(response, Util.getCircularReplacer());
+        const logErr = err ? JSON.stringify(err, Util.getCircularReplacer())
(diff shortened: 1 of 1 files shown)


lib/services/large_result_set.js
```diff
@@ -70,6 +70,10 @@ function LargeResultSetService(connectionConfig, httpClient)
       // err happens on timeouts and response is passed when server responded
       if (err || isUnsuccessfulResponse(response))
       {
+        // if we're running in DEBUG loglevel, probably we want to see the full error too
+        const logErr = err ? JSON.stringify(err, Util.getCircularReplacer()) 
+        : JSON.stringify(response, Util.getCircularReplacer());
```
sfc-gh-pfus, line 75: Not having bigger picture - but this is potentially too much? Are we sure, that no sensitive data is logged?

sfc-gh-dszmolka: we already log the response headers even for a [successful response](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/v1.9.1/lib/services/large_result_set.js#L104-L106) but definitely can use similar approach here too

sfc-gh-dszmolka: done

sfc-gh-pmotacki (approved):
It's good idea to add debug logs for error/unsuccesfull responses.

PR #735, opened 2023-12-15
Reuse storage clients + destroy when no longer needed (fixes #734)

### LINKED ISSUE #734: SNOW-992395: The lib leaks keepalive sockets when uploading files + using S3 client
Hi! I have a service running that uploads lots of files to a Snowflake stage to then insert the data into tables. The problem, however, is that there is a memory leak created by how the Snowflake SDK uses AWS S3 clients. There are many places where the SDK creates temporary clients (to name a few: [here](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/49d0e7bfc1a919334e22b6e9df75a5ff953f3beb/lib/file_transfer_agent/file_transfer_agent.js#L136), [here](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/49d0e7bfc1a919334e22b6e9df75a5ff953f3beb/lib/file_transfer_agent/file_transfer_agent.js#L299) and [here](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/49d0e7bfc1a919334e22b6e9df75a5ff953f3beb/lib/file_transfer_agent/s3_util.js#L117)) but the SDK never calls `client.destroy()` on them (defined in the AWS SDK [here](https://github.com/aws/aws-sdk-js-v3/blob/ad9fb3c116e684a00e4526eb65869939c99c5c4d/clients/client-s3/src/S3Client.ts#L808)).

Each S3 client creates a http agent with keepalive sockets and those are just never destroyed, creating increasing memory usage over the time. While [Node docs](https://nodejs.org/api/http.html#agentdestroy) state that destroying an agent is usually not needed, for keepalive sockets (this case), it's best to do it explicitly and, in our experience, keepalive sockets pointing at AWS infrastructure stay alive for a really long time, enough to cause increasing memory usage.

I also noticed the SDK creates clients (for instance, [here](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/49d0e7bfc1a919334e22b6e9df75a5ff953f3beb/lib/file_transfer_agent/file_transfer_agent.js#L139)) and set them in `meta` but they're never reused in the respective util files when they could. [PR to fix this](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/735) has been created.

1. What version of NodeJS driver are you using?
v1.9.0
   
2. What operating system and processor architecture are you using?
Ubuntu, x86_64
   
3. What version of NodeJS are you using?
v14.19.0

4. What are the component versions in the environment (`npm list`)?
NA

5. Server version:* E.g. 1.90.1
NA

6. What did you do?
My service uploads lots of data files to Snowflake stage which is then inserted into tables.

7. What did you expect to see?
Controlled memory usage over a greater period of time.

9. What is your Snowflake account identifier, if any? (Optional)
NA

GabrielLomba:
This change reuses the storage clients created in file_transfer_agent.js and destroys them when no longer needed to avoid increasing memory usage in the S3 use case (fixes #734 ).

### DIFF
--- a/lib/file_transfer_agent/file_transfer_agent.js
+++ b/lib/file_transfer_agent/file_transfer_agent.js
@@ -150,7 +150,13 @@ function file_transfer_agent(context) {
       meta['dstFileName'] = meta['srcFileName'];
 
       var storageClient = getStorageClient(meta['stageLocationType']);
-      await storageClient.uploadOneFileStream(meta);
+      try {
+        await storageClient.uploadOneFileStream(meta);
+      } finally {
+        if (client.destroy) {
+          client.destroy();
+        }
+      }
     } else {
       parseCommand();
       initFileMetadata();
@@ -312,6 +318,9 @@ function file_transfer_agent(context) {
     if (largeFileMetas.length > 0) {
       await uploadFilesinSequential(largeFileMetas);
     }
+    if (client.destroy) {
+      client.destroy();
+    }
   }
 
   /**
@@ -426,6 +435,9 @@ function file_transfer_agent(context) {
     if (largeFileMetas.length > 0) {
       await downloadFilesinSequential(largeFileMetas);
     }
+    if (client.destroy) {
+      client.destroy();
+    }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


GabrielLomba:

GabrielLomba:
recheck

lib/file_transfer_agent/file_transfer_agent.js
```diff
@@ -312,6 +318,9 @@ function file_transfer_agent(context) {
     if (largeFileMetas.length > 0) {
       await uploadFilesinSequential(largeFileMetas);
     }
+    if (client.destroy) {
```
sfc-gh-dheyman-1, line 321: Is there a reason why you didn't want to use a `try {} finally {}` construct as well? This relates also to line 438.

GabrielLomba: No reason. I missed those two. Added a commit to fix this behavior

sfc-gh-pmotacki (approved):
LGTM

sfc-gh-dheyman-1 (approved):

sfc-gh-dheyman-1:
Hi @GabrielLomba, your PR looks fine. Please fix linter errors and we'll merge it.

GabrielLomba:
Lint errors fixed

sfc-gh-dheyman-1:
In behalf of Snowflake, thank you for your contribution @GabrielLomba! Much appreciated!

PR #742, opened 2023-12-27
Remove tmp package from the driver

sfc-gh-ext-simba-jy:
### Description
- [x] replaced the tmp package with fs module. 

### Checklist


lib/file_transfer_agent/encrypt_util.js
```diff
@@ -32,6 +35,62 @@ function EncryptionMetadata(key, iv, matDesc) {
 
 exports.EncryptionMetadata = EncryptionMetadata;
 
+function TempFileGenerator() {
+
+  this.fileSync = function (option = { dir: os.tmpdir(), prefix: '', postfix: '', extension: '' }) {
+    const randomName = crypto.randomUUID();
+    const fileName = `${option.prefix || ''}${randomName}${option.postfix || ''}${option.extension || ''}`;
```
sfc-gh-ext-simba-hx, line 42: We need dot (.) before extension when its not empty.

lib/file_transfer_agent/encrypt_util.js
```diff
@@ -32,6 +35,62 @@ function EncryptionMetadata(key, iv, matDesc) {
...
+    fs.writeFileSync(fullpath, '');
+
+    const fileDescripter = fs.openSync(fullpath);
+    return { name: fullpath, fd: fileDescripter };
+  };
+  
+  this.file = function (option = { dir: os.tmpdir(), prefix: '', postfix: '', extension: '' }, callback) {
+    let error;
+    const randomName = crypto.randomUUID();
+    const fileName = `${option.prefix || ''}${randomName}${option.postfix || ''}${option.extension || ''}`;
```
sfc-gh-ext-simba-hx, line 58: same here. dot(.) before extension.

lib/file_transfer_agent/encrypt_util.js
```diff
@@ -32,6 +35,62 @@ function EncryptionMetadata(key, iv, matDesc) {
...
+
+    if (!this.checkDirInTemp(option.dir)) {
+      option.dir = os.tmpdir();
+    }
+
+    const fullpath = path.join( option.dir, fileName);
+    
+    try {
+      fs.writeFileSync(fullpath, '');
```
sfc-gh-ext-simba-hx, line 67: I'm not familiar with node.js and always get confused with sync/async functions. Is it OK to call sync functions in async function? Would it be better to return a promise?

sfc-gh-ext-simba-lf: It's ok to call sync functions in an async function. However, it would be preferred to use async functions

But this function doesn't have an async modifier though `this.file = function`

sfc-gh-ext-simba-hx: we have both this.fileSync and this.file so I think this.file supposed to be async function?

lib/file_transfer_agent/encrypt_util.js
```diff
@@ -32,6 +35,62 @@ function EncryptionMetadata(key, iv, matDesc) {
...
+
+    const fullpath = path.join( option.dir, fileName);
+    fs.writeFileSync(fullpath, '');
+
+    const fileDescripter = fs.openSync(fullpath);
+    return { name: fullpath, fd: fileDescripter };
+  };
+  
+  this.file = function (option = { dir: os.tmpdir(), prefix: '', postfix: '', extension: '' }, callback) {
+    let error;
```
sfc-gh-ext-simba-lf, line 56: Does this variable ever get assigned a value?

lib/file_transfer_agent/encrypt_util.js
sfc-gh-ext-simba-lf: It's the callback version of `fileSync`

sfc-gh-ext-simba-hx: sounds good.

lib/file_transfer_agent/encrypt_util.js
sfc-gh-ext-simba-jy: Added.

lib/file_transfer_agent/encrypt_util.js
sfc-gh-ext-simba-jy: Added.

lib/file_transfer_agent/encrypt_util.js
sfc-gh-ext-simba-jy: Removed and put null value instead of the 'error' variable in the callback.

lib/file_transfer_agent/encrypt_util.js
```diff
@@ -32,6 +35,61 @@ function EncryptionMetadata(key, iv, matDesc) {
...
+  
+  this.file = function (option = { dir: os.tmpdir(), prefix: '', postfix: '', extension: '' }, callback) {
+    const randomName = crypto.randomUUID();
+    const fileName = `${option.prefix || ''}${randomName}${option.postfix || ''}${'.' + option.extension || ''}`;
+
+    if (!this.checkDirInTemp(option.dir)) {
+      option.dir = os.tmpdir();
+    }
+
+    const fullpath = path.join( option.dir, fileName);
```
sfc-gh-pmotacki, line 63: The part of creating fullpath is duplicated, what do you think about exctracting?

sfc-gh-ext-simba-jy: I made another method to get the full path.

package.json
```diff
@@ -78,4 +77,4 @@
     "url": "https://www.snowflake.com/"
   },
   "license": "Apache-2.0"
-}
+}
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: This is because there was a blank line at the last line of this file. In my changes, I removed it because I thought it was unnecessary. I made a space again

lib/file_transfer_agent/encrypt_util.js
```diff
@@ -32,6 +35,61 @@ function EncryptionMetadata(key, iv, matDesc) {
...
+    
+    try {
+      fs.writeFileSync(fullpath, '');
+      const fileDescripter = fs.openSync(fullpath);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Fixed.

lib/file_transfer_agent/encrypt_util.js
sfc-gh-pmotacki: I think about even more duplicated lines :

sfc-gh-ext-simba-jy: You are right. I read my code again and realized that the filesync and the file were using the same code until we read the descriptor. As a result, I called the filesync in the file function, and return a callback. Please have a look.

lib/file_transfer_agent/encrypt_util.js
```diff
@@ -32,6 +35,52 @@ function EncryptionMetadata(key, iv, matDesc) {
...
+    const fileDescriptor = fs.openSync(fullpath);
+    return { name: fullpath, fd: fileDescriptor };
+  };
+  
+  this.file = function (option = { dir: os.tmpdir(), prefix: '', postfix: '', extension: '' }, callback) {
+   
+    try {
+      const { name, fd } = this.fileSync(option);
+      callback(null, name, fd);
+
```
sfc-gh-pmotacki, line 60: Remove this empty line please

sfc-gh-ext-simba-jy: removed

PR #745, opened 2024-01-08
SNOW-916946: Fix Native Okta Auth HTTP 429 retry logic for sso/saml endpoint

sfc-gh-ext-simba-jy:
### Description
- [x] refactored AuthOkta steps.
- [x] Added re-authenticate function to renew token
- [x] added testing codes
- [x] applied the retry strategy to the AuthOkta   

### Checklist


lib/authentication/auth_web.js
```diff
@@ -98,12 +99,28 @@ function AuthWeb(connectionConfig, httpClient, webbrowser) {
...
+    while (browserActionTimeout === 0 || Date.now() < currentTime + browserActionTimeout) {
+      try {
+        // Step 3: open browser
+        open(loginUrl);
 
-    // Step 4: get SAML token
-    const tokenData = await withBrowserActionTimeout(browserActionTimeout, receiveData);
-    processGet(tokenData);
+        // Step 4: get SAML token
+        const tokenData = await withBrowserActionTimeout(browserActionTimeout, receiveData);
```
sfc-gh-ext-simba-hx, line 109: I think we need to reduce the browserActionTimeout each time otherwise in the worst case the actual timeout could be doubled.

sfc-gh-ext-simba-jy:
For this implementation, it is hard to write a unit test for this feature because we do not have the option to test connectAsync with the unit test currently. The driver code throws error if I try to test Okta auth with connect function, and mock HTTP client for the unit testing is not supporting connectAsync function.

lib/authentication/auth_okta.js
```diff
@@ -100,7 +91,7 @@ function AuthOkta(password, region, account, clientType, clientVersion, httpClie
   *
   * @returns {Object}
   */
-  function step1(authenticator, serviceName, account, username) {
+  async function step1(authenticator, serviceName, account, username) {
```
sfc-gh-ext-simba-lf, line 94: I know you didn't write this but does the function parameters actually get used in this function? Seems like they can be removed

lib/authentication/auth_okta.js
```diff
@@ -49,47 +49,38 @@ function AuthOkta(password, region, account, clientType, clientVersion, httpClie
...
-      responseHtml = response['data'];
-    });
+    
+    const responseHtml = await step4( await step3(tokenUrl, username, this.password), ssoUrl);
 
     step5(responseHtml);
   };
 
+  this.reauthenticate = async function (body) {
+    await this.authenticate(connectionConfig.getAuthenticator(),
```
sfc-gh-ext-simba-lf, line 77: Is it possible to use the current okta authenticator instead of creating a new one that uses the same credentials?

lib/authentication/auth_okta.js
sfc-gh-ext-simba-jy: Yes, the driver is using them in the step1 function.

`   let header;
    if (serviceName) {
      header = {
        'HTTP_HEADER_SERVICE_NAME': **serviceName**
      };
    }

    // JSON body to send with POST request
    const body = {
      'data': {
        'ACCOUNT_NAME': **account**,
        'LOGIN_NAME': **username**,
        'PORT': port,
        'PROTOCOL': protocol,
        'AUTHENTICATOR': **authenticator**,
        'CLIENT_APP_ID': clientAppId,
        'CLIENT_APP_VERSION': clientAppVersion
      }
    };

    // POST request to get SSO URL and token URL
    return await httpClient.post(url, body, {
      headers: header
    });
  }`

lib/authentication/auth_okta.js
sfc-gh-ext-simba-jy: I checked that we do not need to run step 1 and 2 when the driver reauthenticates. As a result, I fixed the driver to run only step 3 to 5 for renewing the token. Also, I realized that the auth class does not need to have credential info as public, so I refactored some codes. Please have a look.

lib/authentication/auth_okta.js
sfc-gh-ext-simba-lf: My bad, I thought I was looking at the entire function hehe

lib/authentication/auth_okta.js
sfc-gh-ext-simba-lf: Looks good

lib/services/sf.js
```diff
@@ -571,6 +574,12 @@ function StateAbstract(options) {
 
             err = Errors.createNetworkError(
               ErrorCodes.ERR_SF_NETWORK_COULD_NOT_CONNECT, err);
+
+            const auth = getAuthenticator(connectionConfig, httpClient);
+            if (auth instanceof AuthOkta) {
+              Logger.getInstance().debug('OKTA authentication requires token refresh.');
+              await auth.reauthenticate(requestOptions.json);
```
sfc-gh-ext-simba-hx, line 581: Would that approach re-initialize retry timeout and retry count each time and end up infinite retry if the server side keep returning retry-able error code like 429?

sfc-gh-ext-simba-jy: I tested this as the driver throws an Request error with the wrong query to test the retry strategy instead of the network Error. This approach does not reset the retry count because it just creates the error instance. This instance is  passed to the  requestcallback as a parameter(https://github.com/snowflakedb/snowflake-connector-nodejs/pull/745/files#diff-333982e8b49a2955b892e81f109de2f00423a1cdbe166258fb1eded7a6c906e2R1123) and this function will manage the retry.

As a result, I do not think this makes an infinite loop although the server throws 429 error.

Please let me know if I misunderstand your question.

sfc-gh-ext-simba-jy:
I created a new function in authentication to get the current auth method because the function 'getAuthenticator' I have used returns a new instance, so it does not have attributes I saved when I create an authenticator in the connection. As a result, it would be better to create another function to read the current authentication and used it in the different files.

lib/connection/connection_config.js
```diff
@@ -481,6 +481,26 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
+    return numofRetries;
+  };
+
+  this.setNumofRetries = function (retry) {
+    numofRetries = retry;
+  };
+
+  let elapsedTime = 0;
+
+  this.getElapsedTime = function () {
```
sfc-gh-ext-simba-hx, line 496: I don't like this approach since elapsedTime is per request while connection_config is on connection level.
Can we expend reauthenticate() interface to pass elapsed time or maybe even retried attempts there?

sfc-gh-ext-simba-jy: I fixed it and passed the retryOption as a parameter to the re-authenticate function. Please have a look.

lib/authentication/auth_okta.js
```diff
@@ -49,46 +51,61 @@ function AuthOkta(password, region, account, clientType, clientVersion, httpClie
...
+    const remainingTimeout = maxRetryTimeout !== 0 ? (maxRetryTimeout - retryOption.totalElapsedTime) * 1000 : 0;
+    const startTime = Date.now();
     let responseHtml;
-    await step4(oneTimeToken, ssoUrl).then((response) => {
-      responseHtml = response['data'];
-    });
 
+    while (remainingTimeout === 0 || Date.now() < startTime + remainingTimeout) {
+      try { 
+        responseHtml = await step4( await step3(tokenUrl, user, password), ssoUrl);
```
sfc-gh-dheyman-1, line 88: is it possible to rename these methods with more descriptive names? It would make the code much more clean. I mean `step#` methods.

lib/authentication/auth_okta.js
```diff
@@ -49,46 +51,61 @@ function AuthOkta(password, region, account, clientType, clientVersion, httpClie
...
-      } else {
-        oneTimeToken = data['cookieToken'];
-      }
-    });
+    step5(responseHtml);
+  };
 
+  this.reauthenticate = async function (body, retryOption) {
+    const maxRetryTimeout = connectionConfig.getRetryTimeout();
+    const maxRetry = connectionConfig.getRetrySfMaxLoginRetries();
```
sfc-gh-dheyman-1, line 81: `maxRetryCount`

lib/authentication/auth_okta.js
```diff
@@ -49,46 +51,61 @@ function AuthOkta(password, region, account, clientType, clientVersion, httpClie
...
+    const remainingTimeout = maxRetryTimeout !== 0 ? (maxRetryTimeout - retryOption.totalElapsedTime) * 1000 : 0;
```
sfc-gh-dheyman-1, line 82: `: 0` is redundant; can be replaced with _if_ statement

lib/authentication/auth_okta.js
```diff
@@ -49,46 +51,61 @@ function AuthOkta(password, region, account, clientType, clientVersion, httpClie
...
+    const startTime = Date.now();
     let responseHtml;
-    await step4(oneTimeToken, ssoUrl).then((response) => {
-      responseHtml = response['data'];
-    });
 
+    while (remainingTimeout === 0 || Date.now() < startTime + remainingTimeout) {
```
sfc-gh-dheyman-1, line 86: In what cases remainingTimeout could be 0? This seems counterintuitive for me, because if the remainingTimeout reaches 0 shouldn't we stop retrying?

sfc-gh-dheyman-1: Also, shouldn't we have here also a condition that checks numRetries against maxRetryCount?

sfc-gh-ext-simba-jy: Hello @sfc-gh-dheyman-1, Thank you for checking. You are right. We do not need remainingTimeout condition here because we already declared the maxRetryTimeout. As a result, I changed the condition to 
`while (maxRetryTimeout=== 0 || Date.now() < startTime + remainingTimeout)`
Also, I added the maxRetryCount condition.

lib/authentication/auth_okta.js
sfc-gh-ext-simba-jy: I just removed the statement because we do not need to check whether maxRetryTimeout is 0 or not at the remainingTImeout. I used maxRetryTimeout directly.

lib/authentication/auth_okta.js
sfc-gh-ext-simba-jy: Fixed

lib/authentication/auth_okta.js
sfc-gh-ext-simba-jy: I thought all the drivers were using the same function name, but they were not. I changed it and removed the comments which explain what the steps are. Please have a look. Thank you!

lib/authentication/auth_okta.js
```diff
@@ -49,49 +51,64 @@ function AuthOkta(password, region, account, clientType, clientVersion, httpClie
...
 
+    const startTime = Date.now();
     let responseHtml;
-    await step4(oneTimeToken, ssoUrl).then((response) => {
-      responseHtml = response['data'];
-    });
 
-    step5(responseHtml);
-  };
+    while ((maxRetryTimeout === 0 || Date.now() < startTime + remainingTimeout) && retryOption.numRetries <= maxRetryCount) {
```
sfc-gh-dheyman-1, line 87: Now, in case of maxRetryTimeout = 0 & maxRetryCount = 0 we will still perform a retry, right? Overall I think we should have a unit test that checks just this condition in isolation.

lib/authentication/auth_okta.js
sfc-gh-dheyman-1: Great job!

lib/authentication/auth_okta.js
sfc-gh-ext-simba-jy: Hello @sfc-gh-dheyman-1, I understand what you worried about. However, this is a little complicated problem.
If maxRetryTimeout = 0 & maxRetryCount = 0, the driver will not perform the retry because of the point of the retry logic because the minimum value of the numRetries is always 1.

The workflow is

1. obtain the sleep time with the jitter Util function
2. Increased the numRetries value
3. Reauthenticated in Okta.

As a result, if the maxRetryCount = 0, the retryOption.numRetries is always greater than maxRetryCount, the authentication will not be executed. Also, this authentication cannot be executed because of the condition before we start the retry logic in the SF. (https://github.com/snowflakedb/snowflake-connector-nodejs/pull/745/files#diff-333982e8b49a2955b892e81f109de2f00423a1cdbe166258fb1eded7a6c906e2L1101).

If the maxRetryCount = 1, the driver will execute 1 retry because my condition in the reauthenticate is  `retryOption.numRetries **<=** maxRetryCount`.

Therefore, this will not be a problem.

sfc-gh-ext-simba-jy: FYI, I changed one part. In the SF, I replaced the value of numtries 1 to 0. When I implemented the new retry strategy in the past. I just used the same number. However, I found out we should start at 0, not 1 when I answered to your question.
Please have a llok

sfc-gh-ext-simba-jy: It seems like because of the change 1 to 9, the OCSP testing is taking longer than the previous. When the numRetries is 1, it might run 6 retries although the max retry is 7. I extend the testing time for OCSP integration testing because 90 seconds are too short to test the OCSP-fail.

sfc-gh-dheyman-1: I think we can leave it like it is, your explanation makes sense. By unit testing this condition, I meant extracting the condition to a method like `boolean shouldPerformAuthRetry(retryCount, remainingTimeout, etc.)` and testing this method. Apart from that the PR looks fine.

sfc-gh-ext-simba-jy: @sfc-gh-dheyman-1, ah I see. I created a function shouldRetryOktaAuth in the util and made the unit tests. Please have a look.

sfc-gh-dheyman-1: Thanks 👌

PR #754, opened 2024-01-23
SNOW-975781: [snowflake-connector-nodejs][GH#705] storage operations PUT/GET unsuccessful on GCP when GOOGLE_APPLICATION_CREDENTIALS set

sfc-gh-ext-simba-jy:
### Description
- [x] If the environment GOOGLE_APPLICATION_CREDENTIALS is existing in the local machine, the driver will use HTTP request instead of the gcp library.

### Checklist


sfc-gh-pmotacki:
@sfc-gh-ext-simba-jy Basically the PR looks good. I'm wondering that we can test the change some way. WDYT?

lib/file_transfer_agent/gcs_util.js
```diff
@@ -137,7 +137,7 @@ function GCSUtil(httpclient, filestream) {
         let matDescKey;
 
         try {
-          if (accessToken) {
+          if (accessToken && !process.env.GOOGLE_APPLICATION_CREDENTIALS) {
```
pharturs, line 140: I might be wrong (need to check this), but this still will fail if the Google Library is authenticated in other possible ways.

The library can be authenticated using ([doc](https://cloud.google.com/docs/authentication/application-default-credentials)):
- the env variable
- a global config file
- a google meta server

I don't expect the env variable to be set when two other authentication methods are used.

pharturs: confirming, it still fails when GCP is authenticated using the config file or Google's metadata server. 

Perhaps we could have a configuration toggle which enables the use of Axios instead of gcsClient? So that you don't need to make a guesswork if there can be a conflict with the system-wide setup of GCP.

pharturs (changes requested):
This is fixing only 1/3 of the problematic cases with GCP.

lib/file_transfer_agent/gcs_util.js
naira-petrosyan-m, line 140: 1. As stated above this workaround will work only for the cases when the Google authentication is done with env variables. 
And with the other two methods of authentication this fix won't work.

2. Other than that, if the auth is handled, when the axios.head() request is sent, an error is thrown, which looks like this: `AxiosError: Request failed with status code 404.`
Before the latest updates(I have checked with 1.6.3) it seems that the same error is handled [by checking](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/v1.6.3/lib/file_transfer_agent/gcs_util.js#L168C11-L168C42) if the status is 404 and not throwing the error.
But with this new version the `errCode` is [calculated](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/754/files#diff-284c7e4be78ca4c5a17e537e54884c43b1e7f837ec9877eb415078b138ee021eR180) in a way that the same check does not work, because in the error thrown by Axios `err['code'] = ERR_BAD_REQUEST` and the error should be handled with `err.response.status` in addiction to rest of the status checks. 


cc: @pharturs

sfc-gh-ext-simba-jy:
> @sfc-gh-ext-simba-jy Basically the PR looks good. I'm wondering that we can test the change some way. WDYT?

Hello @sfc-gh-pmotacki,  I will try it. It seems like the issue has not been solved yet because I only considered the environment path Google auth for this issue. This implementation can be changed if there are relationships between three different authentications. After everything is fixed, I will try to create a unit test or an integration test for this.

sfc-gh-pmotacki (commented):
I has no comments to code. What about test? There is any idea how to test it?

sfc-gh-ext-simba-jy:
> I has no comments to code. What about test? There is any idea how to test it?

I think we should ask the customer to test this PR first to confirm that this workaround is working. About the testing parts, This is very tricky to test this feature because this issue happens outside of the driver. We can test it with the mock credential.json file or set GOOGLE_APPLICATION_CREDENTIALS, but it is hard to catch all the scenario of this issue.

sfc-gh-dszmolka:
hi @naira-petrosyan-m - once you have a chance, could you please test the https://github.com/snowflakedb/snowflake-connector-nodejs/tree/gcpoperationerror branch from this PR if using the newly introduced `forceGCPUseDownscopedCredential` Connection param helps alleviating the problems in your environment ? Value is `false` by default, and you can set it to `true` to test the scenarios.

Your help in testing is very much appreciated !

naira-petrosyan-m:
@sfc-gh-dszmolka thank you, the issue with GCP auth seems to be fixed, but the other issue I have mentioned in my [comment's 2nd point](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/754#discussion_r1474709153) is not yet fixed, the error response status one. Can you check that issue too? 
cc: @pharturs

naira-petrosyan-m:
Hey @sfc-gh-dszmolka I have created a fork for the error code issue, here is the link, please review this https://github.com/snowflakedb/snowflake-connector-nodejs/pull/764
cc: @pharturs

sfc-gh-dszmolka:
thank you, we'll review!

test/unit/util_test.js
```diff
@@ -1062,4 +1062,41 @@ describe('Util', function () {
...
+        result: true,
+      },
+      {
+        name: 'test - when the disableGCPTokenUplaod is enabled',
+        accessToken: 'Token',
+        forceGCPUseDownscopedCredential: true,
+        result: false,
+      },
+      {
+        name: 'test - when token is emptry but the disableGCPTokenUplaod is enabled',
```
sfc-gh-pmotacki, line 1081: emptry -> empty

test/unit/util_test.js
```diff
@@ -1062,4 +1062,41 @@ describe('Util', function () {
...
+        accessToken: null,
+        forceGCPUseDownscopedCredential: true,
+        result: false,
+      },
+      {
+        name: 'test - test - when token is emptry but the disableGCPTokenUplaod is disabled',
```
sfc-gh-pmotacki, line 1087: emptry -> empty

sfc-gh-pmotacki: Fixed

test/unit/util_test.js
sfc-gh-pmotacki: Fixed

sfc-gh-pmotacki (approved):
LGTM

lib/file_transfer_agent/gcs_util.js
sfc-gh-pmotacki: It was fixed in pr #764

sfc-gh-pmotacki: Fixed in next steps

PR #782, opened 2024-03-01
Fix cannot set sendRequest on undefined in Next.js apps

ShukantPal:
### Description

It seems like that `this.sendRequest` was being used to set `sendRequest` as a global function. But `this` is undefined when using it in Next.js bundled server-side routes.

This change makes it so that the function can be used in the statement.js file without referencing `this`.

### Checklist


ShukantPal:
recheck

PR #792, opened 2024-03-07
SNOW-1205772 native okta url validation parity

sfc-gh-dszmolka:
### Description
* other Snowflake drivers (Python, JDBC) validate only the prefix (protocol, hostname, port) for Native Okta SSO urls
* node.js is different, and today only works for .okta.com Oktas. It fails validation when Okta is under custom domain, which scenario works well with other drivers
* aim of change is to bring parity to the other drivers

### Checklist


lib/authentication/auth_okta.js
```diff
@@ -165,10 +165,16 @@ function AuthOkta(connectionConfig, httpClient) {
   * @returns {null}
   */
   function validateURLs(authenticator, ssoUrl, tokenUrl) {
-    authenticator = authenticator.toLowerCase();
-    if (!(authenticator.startsWith(ssoUrl.substring(0, authenticator.length)) &&
-      authenticator.startsWith(tokenUrl.substring(0, authenticator.length)))) {
-      throw new Error('The prefix of the SSO/token URL and the specified authenticator do not match.');
+    try {
+      const aUrl = new URL(authenticator.toLowerCase());
```
sfc-gh-pmotacki: It is worth to check because probably for creating  new URL case sensitive isn't important

test/unit/authentication/authentication_test.js
```diff
@@ -541,4 +541,58 @@ describe('okta authentication', function () {
     assert.strictEqual(
       body['data']['AUTHENTICATOR'], undefined, 'No authenticator should be present');
   });
+
+  describe('validateURLs test for Native Okta SSO - prefix must match', () => {
+    // original function is private
+    function mockValidateURLs(authenticator, ssoUrl, tokenUrl) {
```
sfc-gh-pmotacki: The test should be run on real code not on a mock with the same, copied code.

test/unit/authentication/authentication_test.js
```diff
@@ -541,4 +541,58 @@ describe('okta authentication', function () {
...
+        tokenurl: 'https://mycustom.okta.com:8443/api/v1/sessions' },
+      { name: 'custom okta format, ssourl and tokenurl match, port doesnt match', authenticator: 'HTTPS://MYAPPS.MYDOMAIN.COM/SNOWFLAKE/OKTA', ssourl: 'https://MYAPPS.MYDOMAIN.COM:8443/app/snowflake/mytokenmytoken/sso/saml',
+        tokenurl: 'https://MYAPPS.MYDOMAIN.COM:8443/api/v1/authn' },
+      { name: '.okta.com format, authenticator port substring of ssourl port', authenticator: 'https://mycustom.okta.com:3030', ssourl: 'https://mycustom.okta.com:30303/app/snowflake/mytokenmytokenmytoken',
+        tokenurl: 'https://mycustom.okta.com/api/v1/sessions' },
+      { name: 'custom okta format, ssourl/tokenurl port substring of authenticator port', authenticator: 'HTTPS://MYAPPS.MYDOMAIN.COM:8443/SNOWFLAKE/OKTA', ssourl: 'https://MYAPPS.MYDOMAIN.COM:443/app/snowflake/mytokenmytoken/sso/saml',
+        tokenurl: 'https://MYAPPS.MYDOMAIN.COM:443/api/v1/authn' }
+    ].forEach(({ name, authenticator, ssourl, tokenurl }) => {
+      it(`${name}`, () => {
+        assert.equal(mockValidateURLs(authenticator, ssourl, tokenurl), true);
```
sfc-gh-pmotacki: We should assert that error was thrown. You can use ``assert.throws()``method

lib/authentication/auth_okta.js
```diff
@@ -164,13 +164,20 @@ function AuthOkta(connectionConfig, httpClient) {
...
+      if (!(compareUrlsByProtocolAndHost(aUrl, sUrl) && compareUrlsByProtocolAndHost(aUrl, tUrl))) {
+        throw new Error('The prefix of the SSO/token URL and the specified authenticator do not match.');
+      }
+    } catch (err) {
+      throw new Error('Authenticator, SSO, or token URL is invalid.');
     }
-  }
+  };
+
+  const compareUrlsByProtocolAndHost =  (firstUrl, secondUrl) => firstUrl.protocol === secondUrl.protocol && firstUrl.host === secondUrl.host;
```
sfc-gh-dprzybysz, line 180: could we make this lambda a local variable in `validateURLs `?

lib/authentication/auth_okta.js
```diff
@@ -164,13 +164,20 @@ function AuthOkta(connectionConfig, httpClient) {
...
-      throw new Error('The prefix of the SSO/token URL and the specified authenticator do not match.');
+  this.validateURLs = function (authenticator, ssoUrl, tokenUrl) {
+    try {
+      const aUrl = new URL(authenticator);
+      const sUrl = new URL(ssoUrl);
+      const tUrl = new URL(tokenUrl);
+      if (!(compareUrlsByProtocolAndHost(aUrl, sUrl) && compareUrlsByProtocolAndHost(aUrl, tUrl))) {
+        throw new Error('The prefix of the SSO/token URL and the specified authenticator do not match.');
+      }
+    } catch (err) {
```
sfc-gh-dprzybysz, line 175: by catching here and throwing a different exception we are loosing the exception from error above? should we pass cause with error?

sfc-gh-dszmolka: a very fair point, my bad. 
problem is that `error.cause` is convenient but unfortunately [not available until node16](https://nodejs.org/docs/latest-v16.x/api/errors.html#errorcause) and while we need to maintain compatibility with node14, had to do something else.

lib/authentication/auth_okta.js
sfc-gh-dszmolka: ok, moved it inside `validateURLs`

lib/authentication/auth_okta.js
sfc-gh-pmotacki: Looks and working correct.

sfc-gh-pmotacki (approved):

PR #799, opened 2024-03-19
SNOW-1230848 Fix timestamp discrepancy related to rounding issue in sf_timestamp

sfc-gh-lmonteromarin:
#### What was the issue?
There were specific timestamps that were displayed incorrectly.
In Snowsight (the data is correct):
image

Using the Node Connector (look specifically at the "TO JSON" log):
image

#### What caused the issue?
When a SfTimestamp is created, it receives a number of epochSeconds and a number of incremental nanoSeconds. The SfTimestamp object has a property for the nanoseconds and a **moment** property.
The number of milliseconds was sometimes rounded up (from 253402300799999.999999 to 253402300800000, for example), but the property nanoseconds was not changed accordingly to reflect this rounding. This meant that, if we had 253402300799999.999999 epoch milliseconds, that would create a SfTimestamp with a moment that represents 253402300800000 epoch milliseconds, and with a nanosecond property of 999999999. That is actually 253402300800 epoch seconds and 999999999 nanoseconds (an additional second). When converting the SfTimestamp to a Date, and then converting the Date to a string, the moment and the nanosecond properties were both used, which resulted in this specific timestamp being printed as 10000-01-01 00:00:00.999 instead of 9999-12-31 23:59:59.999.
 
#### How was the issue fixed?

The fix avoids using rounding when creating the moment object. Instead of creating the moment from a number of milliseconds, it creates it using the epoch seconds and the nanoseconds, in a way that is similar to what is proposed here: https://github.com/moment/moment/issues/2022#issuecomment-114319265.

A unit test was added in order to check that this issue is fixed with the change in `sf_timestamp`

### Checklist

### DIFF
--- a/lib/connection/result/sf_timestamp.js
+++ b/lib/connection/result/sf_timestamp.js
@@ -60,8 +60,9 @@ function SfTimestamp(epochSeconds, nanoSeconds, scale, timezone, format) {
   this.timezone = timezone;
   this.format = format;
 
-  // compute the epoch milliseconds and create a moment object from them
-  let moment = Moment((epochSeconds * 1000) + (nanoSeconds / 1000000));
+  // create a moment object that includes the epoch seconds and the incremental nano seconds
+  // X corresponds to UNIX epoch, SSSSSSSSS corresponds to nano seconds
+  let moment = Moment(`${epochSeconds}.${nanoSeconds}`, 'X.SSSSSSSSS');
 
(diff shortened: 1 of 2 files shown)


sfc-gh-lmonteromarin:

sfc-gh-lmonteromarin:
recheck

sfc-gh-pmotacki (approved):
LGTM

PR #807, opened 2024-03-27
SNOW-1278568: Fix externalbrowser auth for Chrome browser

sfc-gh-pbulawa:
### Description
Starting with Chrome v123 the browser sends the `OPTIONS` request first, instead of `GET` thus the `externalbrowser` auth was failing.

### Checklist

### DIFF
--- a/lib/authentication/auth_web.js
+++ b/lib/authentication/auth_web.js
@@ -102,11 +102,11 @@ function AuthWeb(connectionConfig, httpClient, webbrowser) {
     open(loginUrl);
 
     // Step 4: get SAML token
-    const tokenData = await withBrowserActionTimeout(browserActionTimeout, receiveData).catch((rejected) => {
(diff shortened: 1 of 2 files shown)


lib/authentication/auth_web.js
```diff
@@ -135,6 +135,11 @@ function AuthWeb(connectionConfig, httpClient, webbrowser) {
         // Receive the data and split by line
         const data = chunk.toString().split('\r\n');
 
+        // Make sure that we do not close the socket on other types of requests like OPTIONS or favicon
+        if (!data[0].startsWith('GET /?token=')) {
+          return;
+        }
+
         // Stop accepting connections and close
         socket.destroy();
```
sfc-gh-tkojima, line 144: we should still close the socket, otherwise the requests will hang (we just can't close the server).

sfc-gh-pbulawa: Right, thanks for spotting. Fixed.

sfc-gh-pmotacki (approved):
LGTM

PR #820, opened 2024-04-03
SNOW-824882 SSO token cache

sfc-gh-ext-simba-jy:
### Description
- [x] Added id_token authentication
- [x] Added secureStorage feature
- [x] Added reauthentication  

### Checklist


lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,124 @@
...
+    if (!Util.exists(tokenDir)) {
+      Logger.getInstance().error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location(home). 
+      Please assign the enviroment variable value SF_TEMPORARY_CREDENTIAL_CACHE_DIR to enable the default credential manager.`);
+      return null; 
+    }
+    const tokenCacheFile = path.join(tokenDir, 'temporary_credential.json');
+
+    if (fs.existsSync(tokenCacheFile)) {
+      try {
+        fs.accessSync(tokenCacheFile,  fs.constants.R_OK | fs.constants.W_OK);
```
sfc-gh-pmotacki, line 33: Our goal here was to verify that only user has permission to this file. What about using fs.constants.S_IRUSR and fs.constants.S_IWUSR

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,124 @@
...
+  
+    if (fs.existsSync(this.tokenDir)){
+      jsonCredential = this.readJsonCredentialFile() || {};
+    } else {
+      jsonCredential = {}; 
+    }
+    jsonCredential[key] = token;
+   
+    try {
+      fs.writeFileSync(this.tokenDir, JSON.stringify(jsonCredential), { mode: 0o600 });
```
sfc-gh-pmotacki, line 74: if file exists after 'writeFileSync' there will be updated data and access 600 even if didn't before?

lib/core.js
```diff
@@ -206,6 +206,14 @@ function Core(options) {
 
           GlobalConfig.setKeepAlive(keepAlive);
         }
+
+        const customCredentialManager = options.customCredentialManager;
+        if (Util.exists(customCredentialManager)) {
+          Errors.checkArgumentValid(Util.isObject(customCredentialManager),
+            ErrorCodes.ERR_GLOBAL_CONFIGURE_INVALID_CUSTOM_CREDENTIAL_MANAGER);
+
```
sfc-gh-pmotacki, line 214: Do we create a log that customer will be use customCredentialManager?

sfc-gh-pmotacki: Ok. I see, we have below

test/integration/testManualConnection.js
```diff
@@ -70,6 +72,40 @@ if (process.env.RUN_MANUAL_TESTS_ONLY === 'true') {
...
+        GlobalConfig.getCredentialManager().remove(key);
+        
+        const connectionOption = connOption.externalBrowser;
+
+        it('test - obtain the id token from the server and save it on the local storage', async function () {
+          const connection = snowflake.createConnection(connectionOption);
+          await connection.connectAsync(function (err) {
+            assert.ok(!err);
+            const idToken = GlobalConfig.getCredentialManager().read(key);
+            assert.ok( idToken !== null);
```
sfc-gh-pmotacki, line 87: Could we verify that file contains expected key and some token?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,124 @@
...
+const fs = require('fs');
+const os = require('os');
+const Util = require('../../util');
+
+function JsonCredentialManager() {
+
+  this.userinfo = os.userInfo();
+
+  this.getTokenDir = function () {
+    let tokenDir = process.env.SF_TEMPORARY_CREDENTIAL_CACHE_DIR;
```
sfc-gh-dprzybysz, line 16: should it be an environment variable or can we make it parameter configured via `snowflake.configure(...)`?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,124 @@
...
+    if (!Util.exists(tokenDir)) {
+      tokenDir = os.homedir();
+    } else {
+      Logger.getInstance().info(`The credential cache directory is configured by the user. The token will be saved at ${tokenDir}`);
+    }
+
+    if (!Util.exists(tokenDir)) {
+      Logger.getInstance().error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location(home). 
+      Please assign the enviroment variable value SF_TEMPORARY_CREDENTIAL_CACHE_DIR to enable the default credential manager.`);
+      return null;
```
sfc-gh-dprzybysz, line 27: should we return null or throw an error?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    if (!Util.exists(tokenDir)) {
+      Logger.getInstance().error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location(home). 
+      Please assign the enviroment variable value SF_TEMPORARY_CREDENTIAL_CACHE_DIR to enable the default credential manager.`);
+      return null; 
+    }
+    const tokenCacheFile = path.join(tokenDir, 'temporary_credential.json');
+
+    if (fs.existsSync(tokenCacheFile)) {
```
sfc-gh-dprzybysz, line 31: should we use sync methods? can we switch to async ones?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,124 @@
...
+      try {
+        fs.accessSync(tokenCacheFile,  fs.constants.R_OK | fs.constants.W_OK);
+      } catch (err) {
+        Logger.getInstance().info(`You do not have permission to the credential file: ${tokenCacheFile}.`);
+        return null;
```
sfc-gh-dprzybysz, line 36: should we only log on info level and return null? should we throw the exception?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+    Logger.getInstance().info('Cache directory: ', tokenCacheFile);
+    return tokenCacheFile;
+  };
+
+  this.tokenDir = this.getTokenDir();
+
+  this.readJsonCredentialFile = function () {
+    if (this.tokenDir === null) {
+      return null;
```
sfc-gh-dprzybysz, line 47: should we log before returning null? should we throw the exception? my question about null vs exception is also for following code

lib/connection/connection_config.js
```diff
@@ -474,6 +475,14 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
     includeRetryReason = options.includeRetryReason;
   }
 
+  let clientStoreTemporaryCredential = false;
```
sfc-gh-dprzybysz, line 478: is the default correct? don't we want to use the temporal storage always? @sfc-gh-pmotacki

lib/global_config.js
```diff
@@ -256,4 +257,21 @@ exports.getKeepAlive = function () {
...
+  Logger.getInstance().info('Custom credential manager is set by a user.');
+};
+
+exports.getCredentialManager = function () { 
+  return credentialManager;  
+};  
+
+function createDefaultCredentialManager() {
+  const JsonCrednetialManager = require('./authentication/secure_storage/json_credential_manager');
+  return new JsonCrednetialManager();
```
sfc-gh-dprzybysz, line 276: for MFA tokens we will use the same class? should we pass the directory here?

lib/services/sf.js
```diff
@@ -623,6 +624,18 @@ function StateAbstract(options) {
             // success flag is false, the operation we tried to perform failed
             if (body && !body.success) {
               const data = body.data;
+              const auth = getCurrentAuth();
+
+              if (body.code === GSErrors.code.ID_TOKEN_INVALID && data.authnMethod === 'TOKEN') {
+                Logger.getInstance().debug('ID Token being used has expired. Reauthenticating');
+                const key = Util.buildCredentialCacheKey(connectionConfig.account, 
+                  connectionConfig.username, authenticationTypes.ID_TOKEN_AUTHENTICATOR);
+                await GlobalConfig.getCredentialManager().remove(key);
```
sfc-gh-dprzybysz, line 633: are the called method async? I think I saw it without async and with `fs.xyzSync` methods used?

test/integration/testManualConnection.js
```diff
@@ -70,6 +72,40 @@ if (process.env.RUN_MANUAL_TESTS_ONLY === 'true') {
...
+        const key = Util.buildCredentialCacheKey(connectionOption.host, connectionOption.username, 'ID_TOKEN');
+        GlobalConfig.getCredentialManager().remove(key);
+        
+        const connectionOption = connOption.externalBrowser;
+
+        it('test - obtain the id token from the server and save it on the local storage', async function () {
+          const connection = snowflake.createConnection(connectionOption);
+          await connection.connectAsync(function (err) {
+            assert.ok(!err);
+            const idToken = GlobalConfig.getCredentialManager().read(key);
```
sfc-gh-dprzybysz, line 86: should we call the functions from `GlobalConfig.getCredentialManager()` with await?

test/integration/testManualConnection.js
```diff
@@ -70,6 +72,40 @@ if (process.env.RUN_MANUAL_TESTS_ONLY === 'true') {
...
+            assert.ok( idToken !== null);
+          });
+          await testUtil.destroyConnectionAsync(connection);
+        });
+        
+        it('test - id token reauthentication', async function () {
+          const idTokenConnection = snowflake.createConnection(connectionOption);
+          idTokenConnection.connectAsync(function (err) {
```
sfc-gh-dprzybysz, line 94: should we await here?

test/unit/authentication/custom_credential_manager_test.js
```diff
@@ -0,0 +1,101 @@
...
+const user = 'mock_user';
+const credType = 'mock_cred';
+const key = Util.buildCredentialCacheKey(host, user, credType);
+const randomPassword = randomUUID();
+const defaultCredentialManager = new JsonCredentialManager();
+const mockCustomCrednetialManager = {
+  read: function () {
+    return 'mock_token';
+  },
+  write: function () {
```
sfc-gh-dprzybysz, line 20: don't we expect these functions to return promise? (or being async?)

sfc-gh-ext-simba-jy: There are two test case scenarios. 1. Sync 2. Async.
You can find the async testing case at the bottom: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/820/files#diff-c92bed55d478910c2b3d0fa5930295b793b9d866ee42e7411ad6ac4fc15531adR66

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: I think that will be better for customers. The reason why I used it is the JDBC is using process.env.SF_TEMPORARY_CREDENTIAL_CACHE_DIR parameter. I will fix it.

lib/global_config.js
sfc-gh-ext-simba-jy: Yes, the original idea is to use the same file. I am not sure whether we need to create a different credential file or not.
Do you think that separating MFA tokens and ID Tokens is better? I agree with this.

lib/services/sf.js
sfc-gh-ext-simba-jy: Yes, I created the JSON credential manager with sync methods. However, we need to consider a scenario when users use the async functions for the custom credential manager. That is why I added await or then keywords.

lib/connection/connection_config.js
sfc-gh-ext-simba-jy: On the risk assessment, they mentioned that this option is disabled by default. The customer should read the descripiton first, and need to know what the risk is.

test/integration/testManualConnection.js
sfc-gh-ext-simba-jy: This testing is the manual testing. The purpose of this testing is that we need to verify that the ID token connection is working with its own credentials. As a result, we cannot know the expected key and tokens in this testing. I created the unit testing for the JSON credential manager in a different file.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: Users can set the location where they do not have permission. In this case, I am not sure if the driver throws an error. I think disabling the credential cache and going to the next step is a better idea. What do you think?

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: If we should use the async functions, I think the sync functions are okay for the condition check. Is there any other reason why we need to use async functions?

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: The same idea on above. Please let me know if we need to throw an error in this case or if we just need to disable the credential cache.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: We do not need to log some messages at this point because other error messages are logged already when the driver validates the path on this.getTokenDir. if the tokenDir is null  Also, I am confused about whether we need to throw an exception or disable the credential cache and keep running the driver. Please let me know :)

test/integration/testManualConnection.js
sfc-gh-ext-simba-jy: I fixed it

test/integration/testManualConnection.js
sfc-gh-ext-simba-jy: Fixed.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: In fs.access(Sync), fs.constants.S_IRUSR and fs.constants.S_IWUSR cannot be used because It is out of scope. This function only checks the permission, However, I changed the mode with fs.constants.S_IRUSR and fs.constants.S_IWUSR  when the driver writes or modifies the file.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: Sorry, It is not clear to me. Could you please elaborate on this?

sfc-gh-pmotacki: It is connected with fs.accessSync(tokenCacheFile, fs.constants.R_OK | fs.constants.W_OK). It should be checked before writing token to file as well.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-pmotacki: If this is a consequence of an incorrect client configuration, we should report the error.

lib/global_config.js
sfc-gh-pmotacki: I think it could be the same file. If you see possibility to any conflict we can add type to generated key.

sfc-gh-ext-simba-jy: No, there is no conflict because the key is generated based on the combination of the host, username, and token name (ID_TOKEN or MFA_TOKEN). The same file is also okay.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: @sfc-gh-dprzybysz @sfc-gh-pmotacki, I raised a question about this problem at https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/383. Please have a look.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: fs.constants.R_OK | fs.constants.W_OK means the users should have both the read and write permission. Although we already checked before creating the instance, do we need to validate the permission before writing the cache? Or, do you mean we need to check that the users have the permission of fs.constants.S_IRUSR and fs.constants.S_IWUSR?

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-pmotacki: I will discuss it with security team.

lib/core.js
```diff
@@ -206,6 +207,19 @@ function Core(options) {
...
+          GlobalConfig.setCustomCredentialManager(customCredentialManager);
+        } 
+        // else {
+        //   GlobalConfig.setCustomCredentialManager(new JsonCrednetialManager(credentialCacheDir));
+        // }
```
sfc-gh-pmotacki, line 222: Please remove unused code

sfc-gh-ext-simba-jy: Sorry, I was waiting for the confirmation. I wasn't sure whether I could move the cachedir option to the connection config or not because the conversation above was not resolved yet.  I will remove this.

sfc-gh-ext-simba-jy: I removed it.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-dprzybysz: in the nodejs driver we would like to use as much as possible the nonblocking code - let's avoid the sync methods

But if we really really need to have it then let it be cc @sfc-gh-pmotacki

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-dprzybysz: let's keep one thread for discussing it - I am closing this comment

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-pmotacki: What about: 'fs.statSync('./yourfile.txt').mode'?

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-pmotacki: What about: 'fs.statSync('./yourfile.txt').mode'?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  this.tokenDir = this.getTokenDir();
+
+  this.readJsonCredentialFile = function () {
+    try {
+      const cred = fs.readFileSync(this.tokenDir, 'utf8');
+      return JSON.parse(cred);
+    } catch (err) {
+      Logger.getInstance().info(`Cannot read token data from the file in ${this.tokenDir}. Please check the permission or the file format of the token.`);
```
sfc-gh-dprzybysz, line 49: should we warn here about the missing file?

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-dprzybysz: We are using this function e.g. 
```
const jsonCredential = this.readJsonCredentialFile();
return jsonCredential[key] || null;
```
so here there will be the error when json credentials cannot be read - it's better to throw the exception and handle it in upper layers (or we should return an empty object)

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,119 @@
...
+
+    try {
+      if (!fs.existsSync(this.tokenDir)){
+        Logger.getInstance().info('Cannot find the credential file from the directory');
+        return null;
+      }
+      const jsonCredential = this.readJsonCredentialFile();
+      return jsonCredential[key] || null;
+    } catch (err){
+      Logger.getInstance().error('Failed to read Credential or Permission is denied: ', err.message);
```
sfc-gh-dprzybysz, line 89: there is a missing return in the catch - it's ok to return undefined here?

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-pmotacki: I agree. When we create new piece of code the general rule is to write unblocking code. Please replace to async methods where it is possible.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: I added it.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: I replaced all the sync functions to the asnyc functions. Please have a look.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: As I replaced the sync functions to async functions, the error will be handled on the upper layer. Please have a look. Thank you.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: I replaced.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: I refactored this code. Please have a look

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-ext-simba-jy: Done.

test/integration/testManualConnection.js
```diff
@@ -72,6 +74,41 @@ if (process.env.RUN_MANUAL_TESTS_ONLY === 'true') {
       });
     });
 
+    describe('Connection - ID Token authenticator', async function (done) {
+      const connectionOption = connOption.externalBrowser;
+      connectionOption.clientStoreTemporaryCredential = true;
```
sfc-gh-pmotacki, line 79: It shouldn't be extended this way. It affected more tests. Please fix it by 
'const connectionOption = {
        ...connOption.externalBrowser,
        clientStoreTemporaryCredential: true
      }'

sfc-gh-ext-simba-jy: Fixed.

lib/core.js
```diff
@@ -13,6 +13,7 @@ const Logger = require('./logger');
 const LoggerCore = require('./logger/core');
 const DataTypes = require('./connection/result/data_types');
 const GlobalConfig = require('./global_config');
+// const JsonCrednetialManager = require('./authentication/secure_storage/json_credential_manager');
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Done.

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,109 @@
...
+      return null;
+    }
+  
+    const jsonCredential = this.readJsonCredentialFile() || {};
+    jsonCredential[key] = token;
+   
+    try {
+      await fs.writeFile(await this.getTokenDir(), JSON.stringify(jsonCredential), { mode: 0o600 });
+    } catch (err) {
+      throw new Error(`Failed to write token data from the file. Please check the permission or the file format of the token. ${err.message}`);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: removed.

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -0,0 +1,109 @@
...
+    let tokenDir = credentialCacheDir;
+    if (!Util.exists(tokenDir)) {
+      tokenDir = os.homedir();
+    } else {
+      Logger.getInstance().info(`The credential cache directory is configured by the user. The token will be saved at ${tokenDir}`);
+    }
+
+    if (!Util.exists(tokenDir)) {
+      throw new Error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location(home). 
+      Please assign the enviroment variable value SF_TEMPORARY_CREDENTIAL_CACHE_DIR to enable the default credential manager.`);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Fixed.

sfc-gh-pmotacki:
@sfc-gh-ext-simba-jy please update branch with rebase and update security validation links in prepared in doc as well.

sfc-gh-pmotacki (changes requested):
I found some problem  during running manual test. It must be investigated.

sfc-gh-pmotacki (approved):
LGTM

PR #825, opened 2024-04-16
SNOW-1230848: Fix error related to milliseconds precision

sfc-gh-lmonteromarin:
### Description
Previously, some changes had been done to address the bug SNOW-1230848. However, these changes resulted in a loss of precision in some cases in which milliseconds were used. After thinking more about this, the solution I propose is the following: instead of the original `Moment((epochSeconds * 1000) +(nanoSeconds / 1000000))`, we can use `Moment((epochSeconds * 1000) + Math.trunc(nanoSeconds / 1000000))`. This prevents the bug SNOW-1230848 (which was caused by rounding, and is described in detail [here](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/799), while also preserving milliseconds and nanoseconds precision (some unit tests were added to validate this).

### Checklist

### DIFF
--- a/test/unit/connection/result/result_test_timestamp.js
+++ b/test/unit/connection/result/result_test_timestamp.js
@@ -6,6 +6,65 @@ const Util = require('./../../../../lib/util');
 const assert = require('assert');
 const ResultTestCommon = require('./result_test_common');
 
+function checkSingleTimestamp(dateRowset, timestampOutputFormat, scale, expectedJson, done, additionalValidations = undefined) {
+  const response =
+    {
+      'data': {
+        'parameters': [
+          { 'name': 'TIMEZONE', 'value': 'America/Los_Angeles' },
+          { 'name': 'TIMESTAMP_OUTPUT_FORMAT', 'value': timestampOutputFormat },
+          { 'name': 'TIMESTAMP_NTZ_OUTPUT_FORMAT', 'value': '' },
+          { 'name': 'TIMESTAMP_LTZ_OUTPUT_FORMAT', 'value': '' },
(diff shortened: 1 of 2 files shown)


test/unit/connection/result/result_test_timestamp.js
```diff
@@ -140,5 +141,238 @@ describe('Result: test timestamp', function () {
           done();
         }
       );
-    });
+    }
+  );
+
+  it('select dateadd(ms,999, to_timestamp_ntz(\'2024-04-16T14:57:58\', \'YYYY-MM-DD"T"HH24:MI:SS\')) AS C1;',
```
sfc-gh-mpaluchowski, line 147: Can you also add a test with millsecond precision? Ie. `2024-04-16T14:57:58.999`?

sfc-gh-mhofman: And check as well higher precision of nanos

sfc-gh-lmonteromarin: I have added the requested tests, and cleaned the code somewhat.

I also removed a couple tests that were using `DATEADD` and which were redundant. For each test that used `DATEADD`, now there is a test that parses the corresponding timestamp instead (so that the tests are more readable).

For example, instead of `select dateadd(ms,999, to_timestamp_ntz(\'2024-04-16T14:57:58\', \'YYYY-MM-DD"T"HH24:MI:SS\')) AS C1;` the following expression is used `select  to_timestamp_ntz(\'2024-04-16T14:57:58:999\', \'YYYY-MM-DD"T"HH24:MI:SS:FF3\') AS C1;`

sfc-gh-pmotacki (approved):
LGTM

PR #828, opened 2024-04-18
Fix Util.constructHostname bug with us-west-2 region

jdorn:
### Description
There is a bug in the `Util.constructHostname` method, specifically when the region `us-west-2` is used.  This was a breaking change in either 1.9.x or 1.10.x (not sure which).  It worked fine in version 1.8.3.

To reproduce:

```js
createConnection({
  account: "abc123.us-west-2",
  username: "foo",
  password: "foo",
  database: "foo",
  schema: "foo",
  warehouse: "foo",
  role: "foo"
});
```

Throws an error:

> Cannot read properties of undefined (reading 'endsWith')

Now, it works as expected.

### Checklist


jdorn:

sfc-gh-dszmolka:
hi - per the Snowflake documentation for [account identifiers](https://docs.snowflake.com/en/user-guide/admin-account-identifier#non-vps-account-locator-formats-by-cloud-platform-and-region), the correct notation for an account in us-west-2 is:
1. locator version: 'abc123' (no region, nothing)
2. regionless version: 'myorg-myaccount' (same for every region)

as a mitigation, could you please try `account: "abc123"` ?
however of course, i agree that we could be more resilient in treating the configuration errors thus thank you for your contribution !

jdorn:
Yes, using just "abc123" fixes it, but we have a multi-tenant product and some customers had previously entered "abc123.us-west-2" and it worked fine in version 1.8.0.

We updated the SDK to the latest 1.10.1 and this broke our customers dashboards. And worse, it didn't just fail to connect, it threw a syntax error (because of a typo in the code), which happened outside of our try/catch and crashed the entire Node server.

This PR doesn't introduce any new logic. It just fixes the typo in the existing code and adds unit test coverage to the function.

sfc-gh-pmotacki:
Hi @jdorn Thank you for commit with implementations to fix the problem. I've run a tests and one test fails, can you verify and fix it.

thedae:
Thanks for this fix @jdorn , when are you planning to release it?

sfc-gh-rsavenok:
Hi! First, I want to apologize for the long delay in responding to this.

I am sorry that we introduced a breaking change in 1.x that impacted your workflow. However, the current behavior is actually the intended and documented way for the driver to work. Users in us-west-2 are expected to provide the account name without including the region.

Our upcoming 3.x release will move to a shared codebase across all languages, and it will follow this same logic. Because of that, we cannot merge this fix in the Node driver only to have it reverted when the next major version arrives.

I recommend adding a small check in your own code to format the account name before passing it to the driver. Thank you for your patience and for contributing.

PR #835, opened 2024-05-01
[APPS-40212] Execute could take a cwd option used for put statement context

sfc-gh-klin:
### Description
On the VS Code extension for Snowflake's side, when the user tries to execute a PUT statement that uploads a file path specified by relative path, the cwd used by the extension and node connector is different from the cwd of the user in their workspace, so an error like 
```
Error: ENOENT: no such file or directory, stat 'test.csv'
``` 

will occur, which appears as `Network error` from the connector. 

To help users be able to upload using a relative path, we could allow a "cwd" option to be passed in, and use it if a PUT or GET statement contains a relative path. 

https://snowflakecomputing.atlassian.net/browse/APPS-40212

I am interested in feedback about this suggestion. 

Video of working PUT:

https://github.com/snowflakedb/snowflake-connector-nodejs/assets/104793401/de2d50d9-2430-48af-901a-56cc474c0b39

Video of working GET now:

https://github.com/snowflakedb/snowflake-connector-nodejs/assets/104793401/86b7c796-8b22-45b2-8f49-f3c03b2c10f4

### Checklist


lib/file_transfer_agent/file_transfer_agent.js
```diff
@@ -600,7 +601,14 @@ function FileTransferAgent(context) {
       const src = data['src_locations'][0];
 
       // Get root directory of file path
-      const root = path.dirname(src);
+      let root = path.dirname(src);
+
+      // If root is relative . then replace with context's cwd
+      // Used for VS Code extension where extension cwd differs from user workspace dir
+      if (root === '.') {
```
sfc-gh-klin, line 608: I think I have to replace the relative with the cwd. This check isn't sufficient since if the user is trying to PUT something relative into a folder within the cwd it doesn't cover it. 
ex:
put file://./wow/test.csv @~ overwrite=True;

Also need to fix this for GET commands.

lib/connection/statement.js
```diff
@@ -359,6 +359,7 @@ function createContextPreExec(
   statementContext.multiCurId = statementOptions.multiCurId;
   statementContext.rowMode = statementOptions.rowMode;
   statementContext.asyncExec = statementOptions.asyncExec;
+  statementContext.cwd = statementOptions.cwd;
```
sfc-gh-pmotacki, line 362: Can we validate cwd argument before adding to statementContext?

sfc-gh-pmotacki, line 362: What about assigning cwd to context in function 'createRowStatementPreExec'? It isn't needed?

sfc-gh-klin: It isn't needed (I've applied this fix in the VS Code extension side as a patch), but are you recommending that I do for the sake of consistency?

lib/file_transfer_agent/file_transfer_agent.js
```diff
@@ -600,7 +601,15 @@ function FileTransferAgent(context) {
       const src = data['src_locations'][0];
 
       // Get root directory of file path
-      const root = path.dirname(src);
+      let root = path.dirname(src);
+
+      // If cwd exists and root is relative . then replace with context's cwd
+      // Used for VS Code extension where extension cwd differs from user workspace dir      
+      if (cwd && !path.isAbsolute(src)) {
```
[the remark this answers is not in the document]
sfc-gh-klin: Okay, it looks like I had to set account, not host.

lib/connection/statement.js
sfc-gh-klin: Okay!

sfc-gh-pmotacki: My recommendation is to add in `createRowStatementPreExec` as well.

sfc-gh-klin: If it is performed in `createContextPreExec` isn't it already present in the context object? I can add it again but I'm a little confused on why it is needed. Let me know if I have misunderstood where to put it. I added it in `createRowStatementPreExec`.

sfc-gh-pmotacki: Yes, of course. You are right @sfc-gh-klin. It is a mistake in my comment. My worries was about 'createStatementPostExec'. This is method use for async queries and there the cwd parameter should be added to context as well. WDYT?

sfc-gh-pmotacki:
@sfc-gh-klin I've realised that the new parameter should be added to new types declaration which were merged to master branch. It is needed to rebase the master branch and then add new field declaration to ' interface StatementOption ' in file index.d.ts. It should be declared as optional string type parameter. Thank you.

sfc-gh-pmotacki (approved):
LGTM

PR #851, opened 2024-06-03
SNOW-1454926 statement.cancel isn't available anymore with Typescript (v1.11.0 regression) - missing types

sfc-gh-ext-simba-jy:
### Description
- [x] Added the cancel method in the statement and some new connection config options.
- [x] removed the redundant option in the connection config file. 

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -618,7 +618,7 @@ declare module 'snowflake-sdk' {
          *  Cancels this statement if possible.
          *  @param {Function} [callback]
          */
-        cancel(): void;
+        cancel(callback: StatementCallback): void;
 
         /**
          * Streams the rows in this statement's result. If start and end values are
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -396,21 +432,65 @@ declare module 'snowflake-sdk' {
...
         includeRetryReason?: boolean;
 
         /**
-         * The max login timeout value. This value is either 0 or over 300.
+         * Number of retries for the login request.
          */
-        retryTimeout?: number;
+        sfRetryMaxLoginRetries?: number,
+
+        forceStageBindError
```
sfc-gh-dszmolka, line 467: this does not seem to have a type?

sfc-gh-ext-simba-jy: I did not finish the implementation because I am not sure these options should be in the d.ts file. I finished it, and let me know if we need these options.

lib/connection/connection_config.js
```diff
@@ -57,7 +57,6 @@ const DEFAULT_PARAMS =
   'forceGCPUseDownscopedCredential',
   'disableSamlUrlCheck',
   'representNullAsStringNull',
-  'disableSamlURLCheck'
```
sfc-gh-pmotacki, line 60: Correct property name is 'disableSamlURLCheck', so the property above should be removed.

sfc-gh-pmotacki (changes requested):

sfc-gh-pmotacki (approved):

PR #852, opened 2024-06-04
SNOW-1454180: const enums in TypeScript declaration file means enums cannot be accessed when isolatedModules TS compiler option is set

sfc-gh-ext-simba-jy:
### Description

### Checklist


index.d.ts
```diff
@@ -7,21 +7,21 @@
  * @see [source] {@link https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver}
  */
 declare module 'snowflake-sdk' {
-    export const enum RowMode {
+    export enum RowMode {
```
sfc-gh-pmotacki, line 10: Is it fix the problem? It works for isolatedModules? whether the fix was tested?

sfc-gh-ext-simba-jy: I applied the union type instead of the enum.

sfc-gh-pmotacki (approved):
LGTM

PR #873, opened 2024-07-19
SNOW-1532645: getResultsFromQueryId with complete callback does not allow for proper error handling

sfc-gh-ext-simba-jy:
### Description
- [x] Added one more condition in getQueryStatusThrowIfError function to check whether there is a data or not.

### Checklist


lib/connection/connection.js
```diff
@@ -471,9 +471,13 @@ function Connection(context) {
   this.getQueryStatusThrowIfError = async function (queryId) {
     const status = await this.getQueryStatus(queryId);
 
+    if (status === 'NO_DATA') {
```
sfc-gh-dheyman-1, line 474: is this status really returned only if queryId is invalid? I think it's safer throw error with message like `Cannot retrieve data on the status of this query. No information returned from server for queryID={}.` cc @sfc-gh-pmotacki

sfc-gh-ext-simba-jy: Hello @sfc-gh-dheyman-1. Thank you for your reviewing. Actually, my approach was wrong, and I still need your opinion about the workaround. I posted a comment on the ticket. Please have a look at it when you get a chance. Thank you in advance.

PR #890, opened 2024-08-19
SNOW-1532809: [Node.js] Authenticate with passing passcode

sfc-gh-ext-simba-jy:
### Description
- [x] Implemented MFA authentication with Passcode
- [x] Added new Parameters passcodeInPassword, passcode for connection config. 

### Checklist


lib/authentication/auth_default.js
```diff
@@ -10,19 +10,39 @@
  * @returns {Object}
  * @constructor
  */
-function AuthDefault(password) {
+function AuthDefault(connectionConfig) {
+  const password = connectionConfig.password;
+  const mfaToken = connectionConfig.mfaToken;
+  const passcode = connectionConfig.getPasscode();
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: The mfatoken is configured on the connection level (Because we need to check that the cleintMFARequest option is on and the token exists). That is why it does not have the getter for mfaToken in the connection config for the mfaToken field. I have no clue why the connector has used the field for the password. Might be a security reason? If you want to make a getter for those two fields, I will have a look. Please share your idea.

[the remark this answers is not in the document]
sfc-gh-pmotacki: It was only question about why some fields have getters but another don't have. It isn't incorrect.

lib/authentication/auth_mfatoken.js
```diff
@@ -1,41 +0,0 @@
...
-  
-  /**
-     * Update JSON body with token.
-     *
-     * @param {JSON} body
-     *
-     * @returns {null}
-     */
-  this.updateBody = function (body) {
-    body['data']['AUTHENTICATOR'] = 'USERNAME_PASSWORD_MFA';
```
sfc-gh-pmotacki, line 26: The mfaToken authentication file was removed and there is now code which set authenticator as USERNAME_PASSWORD_MFA. Whether this is a deliberate action?

sfc-gh-ext-simba-jy: The reason why I removed this file is that this authenticator is the same as the default authenticator. On the document, it mentions that "set the authenticator as 'USERNAME_PASSWORD_MFA', so I made the connector use the default authenticator when the user sets the authenticator to 'USERNAME_PASSWORD_MFA'. This implementation is the same as ODBC and JDBC, and I found this when I implemented the libsnowflakeclient passcode authenticator.

sfc-gh-pmotacki: In my opinionthere is a lack of setting this authenticator type directly. Firstly I see that requested generated for authentication with passcode still contains authenticator = 'SNOWFLAKE':

sfc-gh-ext-simba-jy: Thank you for your feedback @sfc-gh-pmotacki  @sfc-gh-pbulawa. I refactored the authentication. Please have a look.

sfc-gh-pbulawa (approved):

lib/authentication/auth_default.js
```diff
@@ -10,19 +10,53 @@
...
     body['data']['PASSWORD'] = password;
+
+    if (isMFAAuth()) {
+      setMFASessionParams(body);
+    }
   };
 
-  this.authenticate = async function () {};
+  function isMFAAuth() {
+    return ( connectionConfig.getAuthenticator() === 'USERNAME_PASSWORD_MFA' || mfaToken || passcode || isPasscodeInPassword);
```
sfc-gh-pmotacki, line 35: Please use authenticationTypes.MFA_TOKEN_AUTHENTICATOR instead of 'USERNAME_PASSWORD_MFA'

lib/authentication/auth_default.js
```diff
@@ -10,19 +10,53 @@
...
+  }
+
+  function setMFASessionParams(body) {
+    body['data']['TOKEN'] = mfaToken;
+    body['data']['AUTHENTICATOR'] = 'USERNAME_PASSWORD_MFA';
```
sfc-gh-pmotacki, line 40: Please use authenticationTypes.MFA_TOKEN_AUTHENTICATOR instead of 'USERNAME_PASSWORD_MFA'

test/integration/testManualConnection.js
```diff
@@ -231,8 +231,36 @@ if (process.env.RUN_MANUAL_TESTS_ONLY === 'true') {
...
+          } catch (err){
+            done(err);
+          }
+        });
+      });
+    });
+
     describe('Connection - MFA authenticator', function () {
-      const connectionOption = { ...connOption.valid, authenticator: 'USERNAME_PASSWORD_MFA', clientRequestMFAToken: true };
+      const connectionOption = { ...connOption.MFA, passcode: null, clientRequestMFAToken: true };
```
sfc-gh-pmotacki, line 263: connOption.MFA doesn't set expected authenticator -> authenticator: 'USERNAME_PASSWORD_MFA'

sfc-gh-pmotacki (changes requested):

lib/authentication/auth_default.js
sfc-gh-ext-simba-jy: Same as above.

test/integration/testManualConnection.js
sfc-gh-ext-simba-jy: I added it.

lib/authentication/auth_default.js
sfc-gh-ext-simba-jy: FYI: I imported the authentication type in the function(class) instead of global because auth_default loads faster than authentication.js. If I import authenticateType from authentication.js top level, the authenticateType is undefined because of the reason above.

sfc-gh-pmotacki: My advice is to extract authentication types to new file like authentication_types.js or authentication_util.js. WDYT?

sfc-gh-ext-simba-jy: That is good idea for the future. I changed my code. Please have a look @sfc-gh-pmotacki.

sfc-gh-pmotacki (approved):
LGTM

PR #918, opened 2024-09-26
Fix/stale authenticator reference for external SSO connection caching

supermacro:
### Description

https://github.com/snowflakedb/snowflake-connector-nodejs/pull/820 Implemented sso token caching. However 
**token caching does not work in version 1.13.1** of this lib (and possibly other versions). Every new connection fails to use the id token that is written to disk. 

Issues found:

- in order for an ID token to be written to disk, `clientStoreTemporaryCredential` needs to be set to true in the connection config, yet it's not a field defined in `index.d.ts`.
- The `connectionConfig` (in `connection.js`) that the `sf` service seems to have access to does not have a `idToken` set (even though there is one available)
    -  one can prove that one is set by `console.log`'ing `connectionConfig.idToken` inside of `connection.js:295`

*relatedly, [the docs](https://docs.snowflake.com/en/user-guide/admin-security-fed-auth-use#using-connection-caching-to-minimize-the-number-of-prompts-for-authentication-optional) are outdated as they do not specify that connection caching is supported for the NodeJS driver*

#### Testing the fix

Run the following script twice:

```javascript
const sf = require('./supermacro-snowflake-connector-nodejs/lib/snowflake')

const conn = sf.createConnection({
  username: '<redacted>',
  account: '<redacted>',
  authenticator: 'externalbrowser',
  database: '<redacted>',
  schema: '<redacted>',
  clientStoreTemporaryCredential: true,
  credentialCacheDir: '.', // relative to current dir where the app is run
})

conn.connectAsync((err, econnected) => {
  if (err) {
    console.error('Unable to connect: ' + err.message);
    return
  }

  console.log('Connection successfully established!')
})
```

Expected / fixed behaviour:

- First run opens up a new browser tab to initiate external SSO flow
- Second run DOES NOT open a new browser tab and instead `snowflake-sdk` uses id token inside of `temporary_credential.json` file that was written to disk during initial authentication flow

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -479,6 +479,11 @@ declare module 'snowflake-sdk' {
         validateDefaultParameters?: boolean;
 
         /**
+         *  The option to enable connection caching. Should be used along with `credentialCacheDir`
+         */
+        clientStoreTemporaryCredential?: boolean;
+
+        /**
          *  The option to set the location where the token will be saved for the token authentication (MFA and SSO).
          *  The path must include the folder path only.
          */
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -305,7 +305,7 @@ function Connection(context) {
     }
    
     // Get authenticator to use
-    const auth = services.sf.getAuthenticator();
+    const auth = Authenticator.getAuthenticator(connectionConfig, context.getHttpClient());
 
     try {
       await initEasyLogging(connectionConfig.clientConfigFile);


supermacro:

supermacro:
recheck

sfc-gh-dszmolka:
can you please check https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-authenticate#authentication-token-caching ? the token cache **should not** work without the server side component turned on.

supermacro:
Related: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/919

supermacro:
> can you please check https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-authenticate#authentication-token-caching ? the token cache **should not** work without the server side component turned on. tested it after the feature was released and worked for me (on Linux).
> 
> if you see an issue, please file an Issue.

We eneabled the server side component by running

```
ALTER ACCOUNT SET ALLOW_ID_TOKEN = TRUE;
```

https://github.com/snowflakedb/snowflake-connector-nodejs/issues/921

PR #938, opened 2024-10-23
SNOW-1346341: Consideration of PROXY*/NO_PROXY env variables in creation httpAgent.

sfc-gh-ext-simba-jy:
### Description
- [x] Implemented PROXY NO PROXY env variables
- [X] Added proxy:false option in axios option(This should be off. If not, there will be two proxy configuration. One in axios and another one in HTTPS_PROXY_Agent)
- [x] In snowflake Config, added an option called "proxy", which users can enable or disable the proxy from the environment variables.

### Checklist


lib/connection/connection_config.js
```diff
@@ -260,17 +221,29 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
-      {
-        host: proxyHost,
-        port: proxyPort,
-        protocol: proxyProtocol,
-        noProxy: noProxy
-      };
+    } 
+  
+    if (!proxy && GlobalConfig.isEnvProxyActive()) {
+      let proxyEnv = Util.getEnvVar('http_proxy') || Util.getEnvVar('https_proxy');
```
sfc-gh-dszmolka, line 227: in theory, http_proxy and https_proxy can be different (not likely, though). so we might need to think about handling it separately. `httpsAgent` will want HTTPS_PROXY, i guess. 
maybe we want to favor https_proxy over http_proxy; and put that to the left? in case both are defined, and are different

lib/connection/connection_config.js
```diff
@@ -208,49 +208,10 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
-      // check for missing proxyPassword
-      Errors.checkArgumentExists(Util.exists(proxyPassword),
-        ErrorCodes.ERR_CONN_CREATE_MISSING_PROXY_PASS);
-
-      // check for invalid proxyPassword
-      Errors.checkArgumentValid(Util.isString(proxyPassword),
-        ErrorCodes.ERR_CONN_CREATE_INVALID_PROXY_PASS);
-
+  if (Util.isNode()) {
+    if ((Util.exists(proxyHost) || Util.exists(proxyPort))) {
```
sfc-gh-dszmolka, line 214: in this path, we try to set `user` and `password` for the proxy as well. but that's only for authenticated proxy. not all proxies need authentication. what happens if the proxy is unauthenticated and only `host` and `port` is set ?

lib/connection/connection_config.js
```diff
@@ -260,17 +221,29 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
+  
+    if (!proxy && GlobalConfig.isEnvProxyActive()) {
+      let proxyEnv = Util.getEnvVar('http_proxy') || Util.getEnvVar('https_proxy');
+
+      if (proxyEnv) {
+        proxyEnv = new URL(proxyEnv);
+        proxy = {
+          host: Util.validateEmptyString(proxyEnv.hostname),
+          port: Number(Util.validateEmptyString(proxyEnv.port)),
+          user: Util.validateEmptyString(proxyEnv.username),
```
sfc-gh-dszmolka, line 234: same thing, maybe `user` and `password` won't exist in proxy URL , if proxy is unauthenticated.

lib/connection/connection_config.js
```diff
@@ -260,17 +221,29 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
-        protocol: proxyProtocol,
-        noProxy: noProxy
-      };
+    } 
+  
+    if (!proxy && GlobalConfig.isEnvProxyActive()) {
+      let proxyEnv = Util.getEnvVar('http_proxy') || Util.getEnvVar('https_proxy');
+
+      if (proxyEnv) {
+        proxyEnv = new URL(proxyEnv);
```
sfc-gh-dszmolka, line 230: this will throw `TypeError` if the user specifies the proxy as `hostname:port` instead of `scheme://hostname:port`

lib/connection/connection_config.js
sfc-gh-ext-simba-jy: Yes, in that case, user and password will be deleted. Please take a look at https://github.com/snowflakedb/snowflake-connector-nodejs/pull/938/files#diff-3fb720f70ea34fb2975fd6c38ac4b6b6131373be2abfa4760dfaae8a25daaee2R855.

lib/connection/connection_config.js
sfc-gh-ext-simba-jy: So, when the driver validates each options, the user and the password will be deleted if they are empty.

lib/connection/connection_config.js
```diff
@@ -260,17 +221,29 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
-      proxy =
-      {
-        host: proxyHost,
-        port: proxyPort,
-        protocol: proxyProtocol,
-        noProxy: noProxy
-      };
+    } 
+  
+    if (!proxy && GlobalConfig.isEnvProxyActive()) {
```
sfc-gh-dszmolka, line 226: what happens if the user deliberately configures `proxyHost` + `proxyPort` (optonally user + password) on the connection config, and their environment also has HTTP_PROXY or HTTPS_PROXY ?

sfc-gh-ext-simba-jy: Then, the driver will only use the proxy info in the connection option,

lib/connection/connection_config.js
sfc-gh-ext-simba-jy: I will fix this.

lib/http/node.js
```diff
@@ -100,6 +100,13 @@ function isBypassProxy(proxy, destination) {
  * @inheritDoc
  */
 NodeHttpClient.prototype.getAgent = function (parsedUrl, proxy, mock) {
+  if (!proxy && GlobalConfig.isEnvProxyActive()) {
+    const isHttps = parsedUrl.protocol === 'https:';
+    proxy = Util.getProxyEnv(isHttps);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: @sfc-gh-dszmolka, that is a good question. Actually, the original request for this ticket is that we can use HTTP_PROXY and HTTPS_PROXY env variables like Axios. However, we have one problem. Axios proxy keeps changing the proxy info based on the location of URL (If the location is http, use HTTP_PROXY. If https, use HTTS_PROXY) call. However, our driver only supports one proxy that a user configured in the connectionConfig.

Currently, I am not sure which workflow we should implement. Let's wait for @sfc-gh-pmotacki response.

Thank you for your questions.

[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Hello @sfc-gh-dszmolka, please take a look at https://github.com/snowflakedb/snowflake-connector-nodejs/pull/938/files#diff-f7fc5e0118097ba33301dd904921ab1c12e77854a9493c441aaf79aadb00be69R1398. This is how Axios works. However, we can change the workflow if we want to use the schema 'http' as default in the driver. Let's wait for @sfc-gh-pmotacki response.

lib/util.js
```diff
@@ -804,3 +805,88 @@ exports.validatePath = function (dir) {
...
+  }
+};
+
+exports.validateEmptyString = function (value) {
+  return value !== '' ? value : undefined;
+};
+
+exports.getProxyEnv = function (isHttps) {
+  let proxyFromEnv = isHttps ? this.getEnvVar('https_proxy') : this.getEnvVar('http_proxy');
+  const protocol = isHttps ? 'https' : 'http';
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Same as above. If users want to use HTTP://proxy in HTTPS_PROXY, they should set HTTPS_PROXY with the schema like http://. As far as I know, other drivers and Axios will be the same

lib/http/node.js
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: FYI:
This is the code from proxy-from-env that Axios uses: https://github.com/Rob--W/proxy-from-env/blob/master/index.js#L44

Axios: https://github.com/axios/axios/blob/v1.x/lib/adapters/http.js#L86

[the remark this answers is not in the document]
sfc-gh-dszmolka: thanks for the answers @sfc-gh-ext-simba-jy ! for the 'this is how axios does it' topics - this very PR is necessary because the stuff axios does with proxies, and it doesn't work, and causes problems :)

lib/util.js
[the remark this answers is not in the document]
sfc-gh-dszmolka: the difficult thing is with these environmental variables is that they are **environmental**. They affect the whole environment, all the applications running in the same environment, not just snowflake-sdk.

The main goal is for snowflake-sdk to play nice with the environmental variables, which already work well with any other application. 

I still think it would be good to consider my suggestion above, i.e. prepending http and logging a line, _if_ HTTPS_PROXY and/or HTTP_PROXY is configured, for some reason, as 'my.pro.xy:1234' instead of 'http://my.pro.xy:1234' (or https://my.pro.xy:1234)

lib/connection/connection_config.js
```diff
@@ -212,63 +212,16 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
-      };
-    }
+    proxy =
+    {
+      host: proxyHost,
+      port: proxyPort,
+      user: proxyUser,
+      password: proxyPassword,
+      protocol: proxyProtocol,
+      noProxy: noProxy
```
sfc-gh-dszmolka, line 222: should we put a validator on this as well, similarly to proxy username and proxy password ? (delete if not supplied)

sfc-gh-ext-simba-jy: Here it is: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/938/files#diff-b74ee385f322c90ac5e644528cf9202501632d6350ad75c154e2c989a81e63b2R224

lib/util.js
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Hello @sfc-gh-dszmolka, it is my bad. I have researched this, and I realized that other drivers have used the 'HTTP' schema if there is no schema in the proxy URL ex) python(request), or curl. I updated my PR based on your comment.

lib/util.js
```diff
@@ -804,3 +805,95 @@ exports.validatePath = function (dir) {
...
+  } else {
+    proxyFromEnv = this.getEnvVar('https_proxy');
+    Logger.getInstance().info('Util.getProxyEnv: Using HTTPS_PROXY from the environment variable');
+  }
+  
+  if (!proxyFromEnv) {
+    return null;
+  }
+  if (proxyFromEnv.indexOf('://') === -1) {
+    proxyFromEnv = 'http' + '://' + proxyFromEnv;
```
sfc-gh-dszmolka, line 878: since in this scenario customer specified 'my.pro.xy' and we using something else instead, http://my.pro.xy, can we please log a INFO that protocol was missing from the environment proxy, and `http` is automatically assumed?

sfc-gh-dszmolka (commented):
can we please log when http is auto-appended?

lib/util.js
```diff
@@ -804,3 +805,95 @@ exports.validatePath = function (dir) {
...
+
+exports.validateEmptyString = function (value) {
+  return value !== '' ? value : undefined;
+};
+
+exports.getProxyEnv = function (isHttps) {
+  let proxyFromEnv;
+  if (!isHttps) {
+    proxyFromEnv = this.getEnvVar('http_proxy');
+    Logger.getInstance().info('Util.getProxyEnv: Using HTTP_PROXY from the environment variable');
```
sfc-gh-dszmolka, line 868: maybe at DEBUG as well?

lib/util.js
```diff
@@ -804,3 +805,95 @@ exports.validatePath = function (dir) {
...
+  } else {
+    proxyFromEnv = this.getEnvVar('https_proxy');
+    Logger.getInstance().info('Util.getProxyEnv: Using HTTPS_PROXY from the environment variable');
```
sfc-gh-dszmolka, line 871: maybe at DEBUG as well, instead of INFO?

sfc-gh-dszmolka (changes requested):
left some comments here (minor changes) and on the internal Issue ticket to review

lib/util.js
sfc-gh-ext-simba-jy: Done.

lib/util.js
sfc-gh-ext-simba-jy: Done.

lib/util.js
sfc-gh-ext-simba-jy: I added it.

index.d.ts
```diff
@@ -245,6 +247,12 @@ declare module 'snowflake-sdk' {
          * pass the custom credential manager to this option.
          */
         customCredentialManager?: object;
+
+        /**
+         * The option whether the driver loads the proxy information from the environment variable or not
+         * The default value is true. If false, it does not use the proxy.
+         */
+        proxy?: boolean;
```
sfc-gh-pmotacki, line 255: WDYT about the more appropriate name? This is only a flag, and the name is too general. Maybe: useEnvProxy ?

lib/constants/error_messages.js
```diff
@@ -24,6 +24,7 @@ exports[403004] = 'Invalid custom JSON parser. The specified value must be a fun
 exports[403005] = 'Invalid custom XML parser. The specified value must be a function.';
 exports[403006] = 'Invalid keep alive value. The specified value must be a boolean.';
 exports[403007] = 'Invalid custom credential manager value. The specified value must be an object, and it should have three methods: write, read, remove';
+exports[403008] = 'Invalid proxy value. The specified value must be a boolean.';
```
sfc-gh-pmotacki, line 27: Change the parameter name as a follow-up to the first comment.

sfc-gh-pmotacki (approved):

index.d.ts
sfc-gh-ext-simba-jy: The reason why I use 'proxy' is because Axios uses it. I will replace this with an appropriate name.

lib/constants/error_messages.js
sfc-gh-ext-simba-jy: Replaced it with useEnvProxy.

index.d.ts
sfc-gh-ext-simba-jy: Replaced it with useEnvProxy.

lib/util.js
```diff
@@ -804,3 +805,91 @@ exports.validatePath = function (dir) {
...
+};
+
+exports.getProxyFromEnv = function (isHttps = true) {
+  const protocol = isHttps ? 'https' : 'http';
+  let proxyFromEnv = this.getEnvVar(`${protocol}_proxy`);
+  if (!proxyFromEnv){
+    return null;
+  }
+
+  Logger.getInstance().debug(`Util.getProxyEnv: Using ${protocol}_PROXY from the environment variable`);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Change it to the uppercase.

sfc-gh-dszmolka (approved):

PR #943, opened 2024-10-30
SNOW-1774162: Add null check for resultIds

sfc-gh-dheyman-1:
### Description

Snowflake NodeJs driver eslint rules application changed resultIds check from a loose to a strict value undefined check, which doesn't match null value. 

### Checklist

### DIFF
--- a/lib/connection/statement.js
+++ b/lib/connection/statement.js
@@ -747,7 +747,7 @@ function createOnStatementRequestSuccRow(statement, context) {
         };
         return;
       }
-      if (body.data.resultIds !== undefined && body.data.resultIds.length > 0) {
+      if (body.data.resultIds != null && body.data.resultIds.length > 0) {
         //multi statements
         this._resultIds = body.data.resultIds.split(',');
         context.isMulti = true;


sfc-gh-pbulawa (approved):

lib/connection/statement.js
```diff
@@ -747,7 +747,7 @@ function createOnStatementRequestSuccRow(statement, context) {
         };
         return;
       }
-      if (body.data.resultIds !== undefined && body.data.resultIds.length > 0) {
+      if (body.data.resultIds != null && body.data.resultIds.length > 0) {
```
sfc-gh-dprzybysz, line 750: can we leave only `!body.data.resultIds` instead of `body.data.resultIds != null`?

sfc-gh-dheyman-1: I didn't want to change semantics, but in this case it makes sense, since it would be basically equivalent of  `body.data.resultIds != null && body.data.resultIds != 0`

sfc-gh-dheyman-1: Then this doesn't really make sense anymore unless we invert it, because `!body.data.resultIds` would be more of an equivalent of `body.data.resultIds == null`. I'd leave it like it is.

sfc-gh-dprzybysz (approved):

Issue #936, opened 2024-10-15
SNOW-1737830: Sporadic error seen in AWS lambda after upgrading to 1.14.0 driver


spdaley:
### Info
1. What version of NodeJS driver are you using?

```
"snowflake-sdk": "1.14.0"
```
   
2. What operating system and processor architecture are you using?

```
AWS Lambda Linux Instances
```
   
3. What version of NodeJS are you using?

```
Node 20
```

4. What are the component versions in the environment (`npm list`)?

```
too many to list?
```

5. What did you do?

```
Unfortuantely, I don't have a complete runnable program for you. This was just something that started happening in our logs after ugprading to 1.14.0. Our application runs in AWS Lambda. We generally try to preserve the snowflake connections across lambda invocations. This typically works just fine. After upgrading to 1.14.0 we started noticing very sporadic errors in the log which say:

Encountered an error when sending the request. Details: {}

There were not a large number of these but enough to set off our error monitoring. Also, the ones we found did not seem to be causing any issues.

From the logs it looks like when some of the lambda invocations start, a snowflake heartbeat event will get started and if the lambda finishes quickly we end up seeing an error in a subsequent invocation. See logs below.

Note, one other thing is these lambdas have multiple data sources that they deal with and sometimes do not actually talk to snowflake during an invocation. I believe all of the incidents where I saw this happen were during executions that never used snowflake in the end.
```

6. What did you expect to see?

```
No error
```

7. Can you collect debug logs?

```
We have some logs now. I've included when the lambdas started and finished in the logs but I've pruned out everything else we were doing. 

"2024-10-10T16:10:09.234Z","REPORT RequestId: 8c70ef49-88c6-4792-ad95-4345f126859d	Duration: 87.92 ms	Billed Duration: 88 ms	Memory Size: 1024 MB	Max Memory Used: 468 MB"
"2024-10-10T16:10:09.234Z","END RequestId: 8c70ef49-88c6-4792-ad95-4345f126859d"
"2024-10-10T16:10:09.209Z","[4:10:09.209 PM]: Encountered an error when sending the request. Details: {}"
"2024-10-10T16:10:09.146Z","START RequestId: 8c70ef49-88c6-4792-ad95-4345f126859d Version: 2842"
[...]

PR #944, opened 2024-11-04
SNOW-1737830 log the original error with JSON.stringify differently (default doesn't log non-enumerable properties, thus in Error: none)

sfc-gh-dszmolka:
### Description
During investigating https://github.com/snowflakedb/snowflake-connector-nodejs/issues/936, it turned out that `JSON.stringify`-ing the request's error in `sf.js` unintentionally hides it. Resulting logline:
```js
Encountered an error when sending the request. Details: {}
``` 
We're supposed to log the actual full error stack instead. 
Tested it and even leaving the circular replacer out and just using the `JSON.stringify(err)` is not enough, logged error is still empty. When logging the raw `err` , the object contents get properly logged:
```
[12:18:59.899 AM]: TypeError: conn.getId is not a function
..
    at options.callback (/opt/nodejs/node_modules/snowflake-sdk/lib/services/sf.js:1333:28)
    at Object.callback (/opt/nodejs/node_modules/snowflake-sdk/lib/services/sf.js:676:43)
    at /opt/nodejs/node_modules/snowflake-sdk/lib/http/base.js:38:24
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  request:
```
This is what we expect; logging something which actually helps troubleshooting errors instead of `{}`. However this causes unnecessary splitting of loglines. 

Tested with `JSON.stringify(err, Object.getOwnPropertyNames(err))` , seems to also fulfill the requirement here, and also can be concatenated into the continuation of the single logline.
```
{"level":"DEBUG","message":"[3:23:36.016 PM]: Encountered an error when sending the request. Details: {\"stack\":\"TypeError: conn.getId is not a function\\n    at /test/i936.js:37:85\\n    at options.callback (/test/snowflake-connector-nodejs/lib/services/sf.js:1334:28)\\n    at Object.callback (/test/snowflake-connector-nodejs/lib/services/sf.js:677:43)\\n    at /test/snowflake-connector-nodejs/lib/http/base.js:38:24\\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\",\"message\":\"conn.getId is not a function\"}"}
```

### Checklist


sfc-gh-dszmolka:
recheck

sfc-gh-dszmolka:
Jenkins job was actually successful yesterday, just progress is not reported back into github apparently. It's still 'Waiting to run this check..' here.

sfc-gh-dprzybysz (approved):

lib/services/large_result_set.js
```diff
@@ -64,11 +64,14 @@ function LargeResultSetService(connectionConfig, httpClient) {
...
       if (err || isUnsuccessfulResponse(response)) {
-        // if we're running in DEBUG loglevel, probably we want to see the full error too
-        const logErr = err ? JSON.stringify(err, Util.getCircularReplacer())
-          : `status: ${JSON.stringify(response.status)} ${JSON.stringify(response.statusText)}` 
-            + ` headers: ${JSON.stringify(response.headers)}`;
-        Logger.getInstance().debug('Encountered an error when getting data from cloud storage: ' + logErr);
+        Logger.getInstance().debug('Encountered an error when getting data from cloud storage. Details: ');
+        if (err) {
+          // if we're running in DEBUG loglevel, probably we want to see the full error too
+          Logger.getInstance().debug(err);
```
sfc-gh-pfus, line 70: I'm not sure I like it. It will be logged in separate lines, right? It may be problematic for all tools for logs gathering (like ELK). Can we join a string and prepare a single message before logging?

lib/services/sf.js
```diff
@@ -593,8 +593,8 @@ function StateAbstract(options) {
           // if we got an error, wrap it into a network error
           if (err) {
             // if we're running in DEBUG loglevel, probably we want to see the full error instead
-            Logger.getInstance().debug('Encountered an error when sending the request. Details: '
-                + JSON.stringify(err, Util.getCircularReplacer()));
+            Logger.getInstance().debug('Encountered an error when sending the request. Details: ');
+            Logger.getInstance().debug(err);
```
sfc-gh-pfus, line 597: Same here.

lib/services/large_result_set.js
sfc-gh-dszmolka: That's what we were doing up until now, and resulted in `{}` as the error, hence this PR :(

lib/services/sf.js
sfc-gh-dszmolka: See above

lib/services/large_result_set.js
sfc-gh-pfus: But can't we do something like this?
```
errMsg = `Encountered blabla...`
if (err) {
   errMsg += err
} else {
   errMsg += `status: ...
}
Logger.getInstance().debug(err)
```

sfc-gh-dszmolka: resolved

lib/services/sf.js
sfc-gh-dszmolka: resolved, too

sfc-gh-pfus (approved):

sfc-gh-pmotacki (approved):

PR #945, opened 2024-11-04
SNOW-1769573 Add support for `describeOnly`

### LINKED ISSUE #940: SNOW-1769573: Support describeOnly
## What is the current behavior?

The NodeJS client does not currently support the `describeOnly` setting to run a describe only query.

## What is the desired behavior?

`connection.execute({ /* other options not shown */, describeOnly: true })`

## How would this improve `snowflake-connector-nodejs`?

Bring the NodeJS client up to parity with other clients like `gosnowflake`, Java, and .NET that support `describeOnly`.

## References, Other Background

https://pkg.go.dev/github.com/snowflakedb/gosnowflake#WithDescribeOnly

https://github.com/snowflakedb/snowflake-jdbc/blob/83e5849103907f6b48a96c023233f2ca3e024a30/src/main/java/net/snowflake/client/core/SFStatement.java#L116

https://github.com/snowflakedb/snowflake-connector-net/blob/81995633744df22936f82e1d0f135b9b76f03945/Snowflake.Data/Core/SFStatement.cs#L223

sfc-gh-pmotacki:
### Description
SNOW-1769573 - Add support for describeOnly.

### Checklist


sfc-gh-pbulawa (commented):
Maybe we should also check a success flow?

test/unit/connection/statement_test.js
```diff
@@ -205,6 +205,16 @@ describe('Statement.execute()', function () {
...
+      {
+        name: 'execute() invalid describeOnly',
+        options: {
+          statementOptions: {
+            sqlText: '',
+            describeOnly: 1,
+          },
+        },
+        errorCode: ErrorCodes.ERR_CONN_EXEC_STMT_INVALID_DESCRIBE_ONLY
       }
```
sfc-gh-fpawlowski, line 218: Can we add a happy path scenario as well? E.g. calling "DESCRIBE SCHEMA TESTSCHEMA;" with describeOnly set to true?

sfc-gh-fpawlowski (changes requested):

sfc-gh-dprzybysz (changes requested):
1. Let's add SNOW-XXX to the PR name
2. Let's add more tests when describe only is passed

test/unit/connection/statement_test.js
sfc-gh-pmotacki: Tests added.

test/integration/testExecute.js
```diff
@@ -137,6 +137,25 @@ describe('Execute test', function () {
       done
     );
   });
+
```
sfc-gh-fpawlowski, line 140: The provided integration test shows, that for describeOnly === undefined we have 1 row returned (no argument in the function call - line 157).
Do we check what happens when it is actually set to true?

test/integration/testExecute.js
```diff
@@ -137,6 +137,25 @@ describe('Execute test', function () {
...
+  it('testDescribeOnly', function (done) {
+    const selectWithDescribeOnly = 'SELECT 1.0::NUMBER(30,2) as C1, 2::NUMBER(38,0) AS C2, \'t3\' AS C3, 4.2::DOUBLE AS C4, \'abcd\'::BINARY(8388608) AS C5, true AS C6';
+
+    function executeQueryAndVerifyResultDependOnDescribeOnly(describeOnly) {
```
sfc-gh-dprzybysz, line 144: why do we need to have a function? cannot we inline it?

test/integration/testExecute.js
```diff
@@ -137,6 +137,25 @@ describe('Execute test', function () {
...
+      connection.execute({
+        sqlText: selectWithDescribeOnly,
+        describeOnly: describeOnly,
```
sfc-gh-dprzybysz, line 147: it's always undefined in the tests?

test/integration/testExecute.js
```diff
@@ -137,6 +137,25 @@ describe('Execute test', function () {
...
+        complete: (err, stmt, rows) => {
+          assert.strictEqual(stmt.getColumns().length, 6);
+          //Empty rowset in response
+          assert.strictEqual(rows.length, describeOnly ? 0 : 1);
```
sfc-gh-dprzybysz, line 151: let's make the `expectedReturnedRows` the parameter for the test and let's check that we have metadata for returned columns

sfc-gh-fpawlowski, line 151: How can the user access data returned by GS for DESCRIBE, since there are no rows returned? Should we have test for that as well?

test/integration/testExecute.js
sfc-gh-pmotacki: My understanding is that setting describeOnly expects only metadata to be returned. The rowset should be empty.

test/integration/testExecute.js
sfc-gh-pmotacki: DESCRIBE also returns only metadata, rows are empty.

sfc-gh-fpawlowski: As we found out on site - the DESCRIBE statement returns rows, regardless of a value set to the describeOnly parameter (e.g. when sql txt of the above test is changed to: "DESCRIBE DATABASE TESTDB").
This happens when selecting data from a real table as well (e.g. "SELECT ... FROM REAL_TABLE_NAME").

Knowing that, we should probably investigate further what is the expected behaviour of this flag, to be able to determine wether our tests are covering such cases.

sfc-gh-pmotacki: Added.

sfc-gh-pbulawa (approved):

test/integration/testExecute.js
```diff
@@ -138,23 +138,33 @@ describe('Execute test', function () {
...
+              assert.equal(JSON.stringify(rows), JSON.stringify(expectedReturnedRows));
+            }
+            return err ? reject(err) : resolve(rows);
+          }
+        });
+      }
+      );
+    };
+    const expectedRows = [{ 'C1': 1, 'C2': 2, 'C3': 't3', 'C4': 4.2, 'C5': { 'type': 'Buffer', 'data': [171, 205] }, 'C6': true }];
+    await executeQueryAndVerifyResultDependOnDescribeOnly(true, []);
```
sfc-gh-dprzybysz, line 165: can we use paramaterized test? I mean `testData.forEach((describeOnly, expectedRows) => ...)`

sfc-gh-pmotacki: Sure. It will be better presented in reports this way.

sfc-gh-dprzybysz (approved):

sfc-gh-fpawlowski (approved):

PR #946, opened 2024-11-05
SNOW-1346341 adding destination to httpsAgent cache key to be able to apply separate settings for separate destinations

sfc-gh-ext-simba-jy:
### Description
- [x] Added the destination when the driver creates the agentID for httpsAgent cache.
- [x] Added the testing for HTTP agent cache. 

### Checklist

### DIFF
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -157,9 +157,4 @@ function getAgentCacheSize() {
   return httpsAgentCache.size;
 }
 
-//This is for the testing purpose.
-function clearAgentCache() {
-  return httpsAgentCache.clear();
-}
-
-module.exports = { NodeHttpClient, getProxyAgent, getAgentCacheSize, clearAgentCache };
+module.exports = { NodeHttpClient, getProxyAgent, getAgentCacheSize };
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -127,9 +127,11 @@ function getProxyAgent(proxyOptions, parsedUrl, destination, mock) {
   try {
     destHost = new URL(destination).hostname;
   } catch (err) {
+    Logger.getInstance().error(`Failed to parse the destination to URL with the error: ${err}. Use the full destination as the host: ${destHost}`);
     destHost = destination;
   }
-  Logger.getInstance().debug(`Destination is ${destHost}`);
+  Logger.getInstance().debug(`The destination host is: ${destHost}`);
+
   const agentId = createAgentId(agentOptions.protocol, agentOptions.hostname, destHost, agentOptions.keepAlive);
   const bypassProxy = isBypassProxy(proxyOptions, destination);
   let agent;
@@ -155,4 +157,9 @@ function createAgentId(protocol, hostname, destination, keepAlive) {
   return `${protocol}//${hostname}-${destination}-${keepAlive ? 'keepAlive' : 'noKeepAlive'}`;
 }
 
-module.exports = { NodeHttpClient, getProxyAgent };
+//This is for the testing purpose.
+function getAgentCacheSize() {
+  return httpsAgentCache.size;
+}
+
+module.exports = { NodeHttpClient, getProxyAgent, getAgentCacheSize };
--- a/lib/http/node.js
+++ b/lib/http/node.js
@@ -123,8 +123,14 @@ function getProxyAgent(proxyOptions, parsedUrl, destination, mock) {
       return mockAgent;
     }
   }
-
-  const agentId = createAgentId(agentOptions.protocol, agentOptions.hostname, agentOptions.keepAlive);
+  let destHost;
+  try {
+    destHost = new URL(destination).hostname;
+  } catch (err) {
+    destHost = destination;
+  }
+  Logger.getInstance().debug(`Destination is ${destHost}`);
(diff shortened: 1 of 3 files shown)


lib/http/node.js
```diff
@@ -123,8 +123,14 @@ function getProxyAgent(proxyOptions, parsedUrl, destination, mock) {
       return mockAgent;
     }
   }
-
-  const agentId = createAgentId(agentOptions.protocol, agentOptions.hostname, agentOptions.keepAlive);
+  let destHost;
+  try {
+    destHost = new URL(destination).hostname;
+  } catch (err) {
+    destHost = destination;
```
sfc-gh-dprzybysz, line 130: should we log the error here?

lib/http/node.js
```diff
@@ -123,8 +123,14 @@ function getProxyAgent(proxyOptions, parsedUrl, destination, mock) {
...
+  }
+  Logger.getInstance().debug(`Destination is ${destHost}`);
+  const agentId = createAgentId(agentOptions.protocol, agentOptions.hostname, destHost, agentOptions.keepAlive);
```
sfc-gh-dprzybysz, line 133: it means that we will always have hostname in the key but it only affects the cases when proxy is used - do we want to have a property in driver to control it?

lib/http/node.js
```diff
@@ -123,8 +123,14 @@ function getProxyAgent(proxyOptions, parsedUrl, destination, mock) {
       return mockAgent;
     }
   }
-
-  const agentId = createAgentId(agentOptions.protocol, agentOptions.hostname, agentOptions.keepAlive);
+  let destHost;
+  try {
+    destHost = new URL(destination).hostname;
+  } catch (err) {
```
sfc-gh-dprzybysz, line 129: can we add a test for both cases?

lib/http/node.js
```diff
@@ -123,8 +123,14 @@ function getProxyAgent(proxyOptions, parsedUrl, destination, mock) {
       return mockAgent;
     }
   }
-
-  const agentId = createAgentId(agentOptions.protocol, agentOptions.hostname, agentOptions.keepAlive);
+  let destHost;
```
sfc-gh-fpawlowski, line 126: Since we are introducing changes in getProxyAgent method, would it be possible to add test cases covering agent caching bahaviour?

lib/http/node.js
sfc-gh-dszmolka: i personally would vote for always having the same structure of the key and not have a separate setting. also would help with not needing to rewrite the applications built on top of snowflake-sdk if we would happen to need to expose it. if no such setting, driver always uses the same cache key structure, then this change can help transparently already

lib/http/node.js
sfc-gh-dszmolka: See some test cases in TW947

lib/http/node.js
sfc-gh-ext-simba-jy: Hello @sfc-gh-dprzybysz, I added an agent cache test. https://github.com/snowflakedb/snowflake-connector-nodejs/pull/946/files#diff-e251ed9eb74ad284dfcd0fcc9a3d2e34d6b18d4b5cba06f21f040beeb0573013R10. This testing includes parsing URL

lib/http/node.js
sfc-gh-ext-simba-jy: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/946/files#diff-e251ed9eb74ad284dfcd0fcc9a3d2e34d6b18d4b5cba06f21f040beeb0573013R10

lib/http/node.js
sfc-gh-ext-simba-jy: I added it: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/946/files#diff-3fb720f70ea34fb2975fd6c38ac4b6b6131373be2abfa4760dfaae8a25daaee2R905

lib/http/node.js
sfc-gh-ext-simba-jy: Please let me know if I need to add an option for this. I agree with @sfc-gh-dszmolka's opinion.

lib/http/node.js
sfc-gh-ext-simba-jy: Hello @sfc-gh-fpawlowski. I added testing cases here: https://github.com/snowflakedb/snowflake-connector-nodejs/pull/946/files#diff-e251ed9eb74ad284dfcd0fcc9a3d2e34d6b18d4b5cba06f21f040beeb0573013R10. Please take a look when you have a chance. Thank you.

sfc-gh-pmotacki (approved):

lib/http/node.js
sfc-gh-dszmolka: @sfc-gh-dprzybysz since this is _the_ core of this whole change; to always add a destination to agent cache key, for being able to utilize separate settings for separate destination, do you agree to proceed like this?

sfc-gh-fpawlowski (approved):

sfc-gh-dprzybysz (approved):

lib/http/node.js
sfc-gh-fpawlowski: Thank You, all looks great now 👍

PR #963, opened 2024-11-18
SNOW-1631790 transport layer

sfc-gh-fpawlowski:
### Description

### Checklist


lib/http/base.js
```diff
@@ -29,42 +32,56 @@ function HttpClient(connectionConfig) {
...
+  let requestPromise;
   const requestOptions = prepareRequestOptions.call(this, options);
   let sendRequest = async function sendRequest() {
-    request = axios.request(requestOptions).then(response => {
+    Logger.getInstance().trace('Request%s - sending.', requestUtil.describeRequestFromOptions(requestOptions));
+    const timer = new ExecutionTimer().start();
+    requestPromise = axios.request(requestOptions).then(response => {
+      const httpResponseTime = timer.getDuration();
+      Logger.getInstance().debug('Request%s - response received after %s milliseconds with status %s.', requestUtil.describeRequestFromOptions(requestOptions), httpResponseTime, response.status);
       sanitizeAxiosResponse(response);
```
sfc-gh-fpawlowski, line 45: I believe that to contain parity with the 'requestAsync' method, next to the 'sanitizeAxiosResponse' method, there should be a 'parseResponseData(response)' statement.

However at this point, it may cause a BCR, since JSON data was not parsed previously in this Synch method.

lib/http/browser.js
```diff
@@ -13,6 +14,7 @@ const Base = require('./base');
  * @constructor
  */
 function BrowserHttpClient(connectionConfig) {
+  Logger.getInstance().trace('Initializing BrowserHttpClient with connection config');
```
sfc-gh-pmotacki, line 17: Is it worth adding more details from connectionConfig?

sfc-gh-fpawlowski: Thanks, I added them for all HttpClients with new describeIdentityAttributes() method of ConnectionConfig objects. 

Important note tho - I believe it is not possible to unambiguously identify the connection for which the HttpClient / ConnectionConfig is used, by looking only at those attributes - they can be reused for multiple connections. Nevertheless they increase the traceability of events represented in logs - what costs us quite increased verbosity in those situations.

lib/http/node.js
```diff
@@ -100,17 +106,19 @@ function isBypassProxy(proxy, destination) {
  * @inheritDoc
  */
 NodeHttpClient.prototype.getAgent = function (parsedUrl, proxy, mock) {
+  Logger.getInstance().trace('Agent[url: %s] - getting an agent instance.', parsedUrl.href);
```
sfc-gh-fpawlowski, line 109: During additional testing in turned out that 'href' of parsedUrl can contain tokens as query params under path like '...queries/download-result?token=ver:1-hint:1036-...'.
Therefore the approached should be changed either to using only host and protocol info or to obfuscating the href on the fly to avoid exposing confidential data.

sfc-gh-fpawlowski: Add token to mask secrets in logger.

lib/http/browser.js
sfc-gh-pmotacki: Yes, you are right, @sfc-gh-fpawlowski. However, we can not log all connection config data.

sfc-gh-dheyman-1 (approved):

PR #967, opened 2024-11-19
SNOW-1814745: Fix aborting requests in HTTP client. Normalization of responses unified.

sfc-gh-fpawlowski:
### Description

### Checklist


test/integration/testHttpClient.js
```diff
@@ -0,0 +1,113 @@
...
+      // Abort the request after some time
+      await testUtil.sleepAsync(TIME_IN_MS_TO_WAIT_BEFORE_ABORT);
+      requestObject.abort();
+      await requestObject.requestPromise;
+    });
+  });
+
+  describe('Normalizing Response', () => {
+    it('should return a normalized response with statusCode and body for requestAsync', async () => {
+      const testUrl = hangWebServerUrl + '/json';
```
sfc-gh-dprzybysz, line 75: can we extract common variable above the test to describe?

ci/container/hang_webserver.py
```diff
@@ -11,36 +11,41 @@ class HTTPRequestHandler(BaseHTTPRequestHandler):
     # counts specific calls to change behaviour after some calls
     counter = 0
 
-    def __respond(self, http_code, content_type='text/plain', body=None, ):
-        if body:
-            self.send_response(http_code, body)
+    def __respond(self, http_code, content_type='text/plain', message=None, body=None, ):
+        if message:
```
sfc-gh-dprzybysz, line 15: why do we need this change?

test/integration/testHttpClient.js
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  beforeEach(() => {
+    httpClientInstance = new NodeHttpClient(httpConnectionOptions);
+  });
+
+  describe('Aborting requests', () => {
+    it('should allow aborting any request immediately', async () => {
+      const requestObject = httpClientInstance.request({
+        url: hangWebServerUrl + '/hang',
```
sfc-gh-dprzybysz, line 28: let's have single const for url - it's the same for many tests

test/integration/testHttpClient.js
```diff
@@ -0,0 +1,113 @@
...
+        method: 'GET',
+        callback: (err) => {
+          // We expect an error due to aborting the request.
+          if (err) {
+            testUtil.isRequestCancelledError(err);
+            assert.equal(err.message, 'canceled');
+            assert.equal(err.name, 'CanceledError');
+            assert.equal(err.code, 'ERR_CANCELED');
+          } else {
+            throw Error('Expected an error from aborted request, but got success.');
```
sfc-gh-dprzybysz, line 38: throwing error in callback will fail the test? questio nalso in other calls with callback

test/integration/testHttpClient.js
```diff
@@ -0,0 +1,113 @@
...
+const { hangWebServerUrl } = require('../hangWebserver');
+const assert = require('assert');
+const testUtil = require('./testUtil');
+
+describe('HttpClient Tests', () => {
+  let httpClientInstance;
+
+  const connectionOptions = {
+    ...(connOptions.valid),
+    timeout: 100000
```
sfc-gh-dprzybysz, line 13: why 10 seconds? not too much?

test/integration/testHttpClient.js
```diff
@@ -0,0 +1,113 @@
...
+            assert.ok(response, 'Response should be defined');
+            assert.ok(response.statusCode, 'Normalized response should have statusCode');
+            assert.ok(response.body, 'Normalized response should have body');
+          } catch (err) {
+            errorRaisedInCallback = err;
+          }
+        }
+      });
+      //Due to usage of 'nextTick' in the httpClient requestPromise may be undefined for some time, only to be set in when scheduled sending took place.
+      while (!requestObject.requestPromise) {
```
sfc-gh-dprzybysz, line 105: let's have max wait time defined somewhere

test/integration/test_utils/httpInterceptorUtils.js
```diff
@@ -82,7 +82,7 @@ function HttpClientWithInterceptors(connectionConfig, initialInterceptors) {
 
 Util.inherits(HttpClientWithInterceptors, NodeHttpClient);
 
-
+//To add new methods to be intercepted wrap them here with appropriate interceptors calls
```
sfc-gh-dprzybysz, line 85: do we need this comment? what does it change in the context of the whole PR?

ci/container/hang_webserver.py
sfc-gh-fpawlowski: Because there was no option to return the actual body from our hang_webserver. What was previously called 'body' was actually returned as message in request with empty body.

To test proper parsing of a request body we need it to contain some value.

test/integration/testHttpClient.js
sfc-gh-fpawlowski: That's right - now I incorporated everywhere the approach from the other test (with assignment of an error from callback to the variable from the outer scope).

test/integration/testHttpClient.js
sfc-gh-fpawlowski: You are right, changed to 2 seconds - should be enough.

test/integration/test_utils/httpInterceptorUtils.js
sfc-gh-fpawlowski: I believe it is important manual fur future usage of the HttPClientWithInterceptors - it is only required to override the original method in this place (surrounding its execution with interceptors calls) and then we have all set up.
I believe I missed this comment in the previous PR.

test/integration/testHttpClient.js
sfc-gh-fpawlowski: Moved to testUtil and added there.

test/integration/testHttpClient.js
sfc-gh-fpawlowski: Done 👍

sfc-gh-dprzybysz (approved):

sfc-gh-pmotacki (approved):
LGTM

PR #972, opened 2024-11-25
SNOW-1825719 - setting proxy port if default was skipped during parsing

sfc-gh-pmotacki:
### Description

### Checklist

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -862,9 +862,10 @@ exports.validateEmptyString = function (value) {
 };
 
 exports.getProxyFromEnv = function (isHttps = true) {
-  const getDefaultPortIfNotSet = (proxyFromEnv, isHttps) => {
+  const getDefaultPortIfNotSet = (proxyFromEnv) => {
+    const isProxyProtocolHttps = proxyFromEnv.protocol === 'https:';
     if (!proxyFromEnv.port) {
-      return isHttps ? 443 : 80;
+      return isProxyProtocolHttps ? 443 : 80;
     } else {
       return proxyFromEnv.port;
     }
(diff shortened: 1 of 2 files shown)


sfc-gh-dprzybysz (approved):

lib/util.js
```diff
@@ -872,11 +879,13 @@ exports.getProxyFromEnv = function (isHttps = true) {
...
     Logger.getInstance().info('Util.getProxyEnv: the protocol was missing from the environment proxy. Use the HTTP protocol.');
     proxyFromEnv = 'http' + '://' + proxyFromEnv;
+    isHttps = false;
   }
   proxyFromEnv = new URL(proxyFromEnv);
+  const port = getDefaultPortIfNotSet(proxyFromEnv, isHttps);
   const proxy = {
     host: this.validateEmptyString(proxyFromEnv.hostname),
-    port: Number(this.validateEmptyString(proxyFromEnv.port)),
+    port: Number(this.validateEmptyString(port)),
```
sfc-gh-astachowski, line 888: I think this validate is no longer needed

sfc-gh-pmotacki: Ok, fixed

lib/util.js
```diff
@@ -871,12 +878,13 @@ exports.getProxyFromEnv = function (isHttps = true) {
   Logger.getInstance().debug(`Util.getProxyEnv: Using ${protocol.toUpperCase()}_PROXY from the environment variable`);
   if (proxyFromEnv.indexOf('://') === -1) {
     Logger.getInstance().info('Util.getProxyEnv: the protocol was missing from the environment proxy. Use the HTTP protocol.');
-    proxyFromEnv = 'http' + '://' + proxyFromEnv;
+    proxyFromEnv = protocol + '://' + proxyFromEnv;
```
sfc-gh-astachowski, line 881: shouldn't we check if we need to change isHttps here?

lib/util.js
```diff
@@ -871,12 +878,13 @@ exports.getProxyFromEnv = function (isHttps = true) {
   Logger.getInstance().debug(`Util.getProxyEnv: Using ${protocol.toUpperCase()}_PROXY from the environment variable`);
   if (proxyFromEnv.indexOf('://') === -1) {
     Logger.getInstance().info('Util.getProxyEnv: the protocol was missing from the environment proxy. Use the HTTP protocol.');
```
sfc-gh-astachowski, line 880: We should change this to log proper protocol

sfc-gh-dszmolka: this line is not part of this change; but the earlier implementation with #938 . 
It was agreed, that we follow the practice set in the industry and the other drivers, and will use the HTTP protocol if the user does not set it in the envvar.

lib/util.js
sfc-gh-dszmolka, line 881: please, let's use HTTP if the user doesn't define the protocol. See above comment.

sfc-gh-dszmolka (changes requested):

sfc-gh-dheyman-1 (approved):

Issue #932, opened 2024-10-08
`1.3.6.1.5.5.7.48.1 not found in AuthorityInfoAccess` and OCSP warning due to missing OCSP Responder in Google cert


archiewood:
We are getting warning messages about OCSP Responses using the latest driver version.
These are very noisy for our clients.

```
WARNING!!! using fail-open to connect. Driver is connecting to an HTTPS endpoint without OCSP based Certificated Revocation checking as it could not obtain a valid OCSP Response to use from the CA OCSP responder. Details: Error: 1.3.6.1.5.5.7.48.1 not found in AuthorityInfoAccess
```

This happens in particular with larger queries, see my repro

## Info

1. What version of NodeJS driver are you using?

```
"snowflake-sdk": "^1.14.0"
```
   
3. What operating system and processor architecture are you using?

```
Mac Silicon
```

4. What version of NodeJS are you using?
(`node --version` and `npm --version`)

```
node --version
v20.15.0
npm --version
10.7.0
```

5. What are the component versions in the environment (`npm list`)?

```
npm list
repro-snowflake-ocsp@1.0.0 /Users/archie/Projects/repro-snowflake-ocsp
└── snowflake-sdk@1.14.0
```

6.Server version:* E.g. 1.90.1

```
'CURRENT_VERSION()': '8.38.2'`
```

7. What did you do?

See minimal repro in this repository: https://github.com/archiewood/repro-snowflake-ocsp

```javascript
const snowflake = require('snowflake-sdk');

// Connection configuration
const config = {
    account: '<your_account>',
    username: '<your_username>',
    password: '<your_password>',
    warehouse: '<your_warehouse>',
    database: '<your_database>',
    schema: '<your_schema>'
};

// Create connection
const connection = snowflake.createConnection(config);

// Connect to Snowflake
connection.connect((err, conn) => {
    if (err) {
        console.error('Unable to connect: ' + err.message);
    } else {
        console.log('Successfully connected to Snowflake.');
        // Execute a sample query
        const query = `
        SELECT CURRENT_VERSION();
        WITH RECURSIVE generate_series(n) AS (
            SELECT 1
            UNION ALL
            SELECT n + 1
            FROM generate_series
            WHERE n < 3000000
        )
        SELECT 
            n AS id,
            RANDSTR(10, RANDOM()) AS random_string,
            UNIFORM(1, 100, RANDOM()) AS random_number,
[...]

PR #973, opened 2024-11-25
SNOW-1825607 Initial OCSP deprecation plan steps

sfc-gh-ext-simba-jy:
### Description
- [x] FAIL_OPEN log changed (message + log level)?
- [x] FAIL_OPEN passing request along if responder URL is missing?
- [x] FAIL_CLOSED - no NPE if responder URL is missing? Please check the message below about this.
Updated: the driver will log the debug message and return callback with null if the OCSP status is FAIL_OPEN. If the status is FAIL_CLOSED, the driver will log the error message and return callback with the error.
- [x] Retries and timeouts confirmed (both to responder and cache server)?
- [x] `disableOCSPChecks` introduced?
  - [x] Deprecated `insecureMode`?
  - [x] `disableOCSPChecks` has precedence?
  
### Checklist


sfc-gh-ext-simba-jy:
The node.js driver does not have the case that the responder URL does not exist according to the code: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/bada542bbb8b46026a903d45b622562e65348e08/lib/agent/check.js#L176. As we have the default OCSP_RESPONSE_CACHE_SERVER_URL, this will not throw the error like JDBC.

lib/agent/socket_util.js
```diff
@@ -120,7 +115,7 @@ exports.secureSocket = function (socket, host, agent, mock) {
 function isOcspValidationDisabled(host) {
   // ocsp is disabled if insecure-connect is enabled, or if we've disabled ocsp
   // for non-snowflake endpoints and the host is a non-snowflake endpoint
-  return GlobalConfig.isInsecureConnect() ||
+  return GlobalConfig.isOCSPChecksDisabled() ||
```
sfc-gh-pfus, line 118: Does this cover also a case when someone sets `insecureMode`?

lib/agent/socket_util.js
```diff
@@ -158,7 +153,7 @@ function canEarlyExitForOCSP(errors) {
       const err = errors[errorIndex];
       if (err && !isValidOCSPError(err)) {
         // any of the errors is NOT good/revoked/unknown
-        Logger.getInstance().warn(ocspFailOpenWarning + err);
+        Logger.getInstance().debug('OCSP responder didn\'t respond correctly. Assuming certificate is not revoked');
```
sfc-gh-pfus, line 156: Can we leave details as it was before?

lib/core.js
```diff
@@ -196,14 +196,18 @@ function Core(options) {
           Logger.getInstance().info('Configuring logger with level: %s, filePath: %s, additionalLogToConsole: %s', logLevel, logFilePath, additionalLogToConsole);
         }
 
-        const insecureConnect = options.insecureConnect;
-        if (Util.exists(insecureConnect)) {
+        if (Util.exists(options.insecureConnect)) {
+          Logger.getInstance().warn('Warning! The option insecureConnect will be deprecated. Please use the disableOCSPChecks.');
```
sfc-gh-pfus, line 200: Actually it is deprecated, not will be :)

lib/core.js
```diff
@@ -196,14 +196,18 @@ function Core(options) {
...
+        }
+
+        const disableOCSPChecks = options.insecureConnect || options.disableOCSPChecks;
```
sfc-gh-pfus, line 203: Are this values nullable? If yes and both are not null, `disableOCSPChecks` should have precedence, not the alternative.

test/unit/ocsp/test_unit_ocsp_mode.js
```diff
@@ -9,13 +9,13 @@ const assert = require('assert');
 describe('OCSP mode', function () {
   it('getOcspMode', function (done) {
     // insecure mode
-    GlobalConfig.setInsecureConnect(true);
+    GlobalConfig.setDisableOCSPChecks(true);
```
sfc-gh-pfus, line 12: Can we have also a test confirming that `insecureMode` still works?

test/unit/snowflake_config_test.js
```diff
@@ -15,7 +15,7 @@ describe('Snowflake Configure Tests', function () {
   before(function () {
     originalConfig = {
       logLevel: Logger.getInstance().getLevelTag(),
-      insecureConnect: GlobalConfig.isInsecureConnect(),
+      disableOCSPChecks: GlobalConfig.isOCSPChecksDisabled(),
```
sfc-gh-pfus, line 18: Same here + test for setting both values at once.

sfc-gh-pfus:
> It seems like the node.js driver does not have the use case that the responder URL does not exist according to the code:
> 
> https://github.com/snowflakedb/snowflake-connector-nodejs/blob/bada542bbb8b46026a903d45b622562e65348e08/lib/agent/check.js#L176
> 
> . As we have the default OCSP_RESPONSE_CACHE_SERVER_URL and there will be no error for the responder URL, this will not throw the error like JDBC.

What you mentioned is not a responder URL - it is a cache server URL. Responder URL is provided in each certificate.

sfc-gh-dszmolka:
> It seems like the node.js driver does not have the use case that the responder URL does not exist according to the code:
> 
> https://github.com/snowflakedb/snowflake-connector-nodejs/blob/bada542bbb8b46026a903d45b622562e65348e08/lib/agent/check.js#L176
> 
> . As we have the default OCSP_RESPONSE_CACHE_SERVER_URL and there will be no error for the responder URL, this will not throw the error like JDBC.


OCSP Responder url != OCSP Cache URL.  The (default) OCSP Cache URL is always the same, by default, `ocsp.snowflakecomputing.com` (gracefully ignoring privatelink situations here)

However OCSP Responder URL is not from Snowflake. It is always hardcoded into the particular certificate, which is under verification. See this random cert as an example: https://crt.sh/?sha256=9c3f2fd11c57d7c649ad5a0932c0f0d29756f6a0a1c74c43e1e89a62d64cd320
```
            Authority Information Access: 
                CA Issuers - URI:http://i.pki.goog/r4.crt
```

It _does not_ have OCSP endpoint. The driver today will complain if it encounters such certs which doesn't have the OCSP endpoint, because the driver expects the cert to have it. See this issue what exactly happens: https://github.com/snowflakedb/snowflake-connector-nodejs/issues/932

As a comparison, here's a cert which _does_ have OCSP endpoint: https://crt.sh/?id=12092745633
Observe:
```
            Authority Information Access: 
>>>>            OCSP - URI:http://ocsp.r2m03.amazontrust.com
                CA Issuers - URI:http://crt.r2m03.amazontrust.com/r2m03.cer
```
(this section is missing from the first cert)

The goal here, is for the driver if encountering a cert _without_ the said OCSP endpoint, then it should just log the debug-level message and gracefully continue as if the verification was successful. 
Implementing this change will also fix 932.

lib/agent/socket_util.js
sfc-gh-ext-simba-jy: Done.

lib/core.js
sfc-gh-ext-simba-jy: Changed.

lib/agent/socket_util.js
sfc-gh-ext-simba-jy: Suppose you mean ` insecureConnect`, Yes. In the core.js, the driver will take both insecureConnect and disableOCSPchecks. The disableOCSPChecks option is prioritized.

test/unit/ocsp/test_unit_ocsp_mode.js
sfc-gh-ext-simba-jy: I made testing for insecureConnect. Thank you.

test/unit/snowflake_config_test.js
sfc-gh-ext-simba-jy: Done.

sfc-gh-pfus (approved):

index.d.ts
```diff
@@ -219,10 +219,16 @@ declare module 'snowflake-sdk' {
         additionalLogToConsole?: boolean | null;
 
         /**
-         * Check the ocsp checking is off.
+         * @deprecated
+         * This option will be deprecated. Use disableOCSPChecks
```
sfc-gh-pfus, line 223: `is deprecated` instead of `will be`

sfc-gh-pfus (approved):

index.d.ts
sfc-gh-ext-simba-jy: I fixed it. Thank you.

sfc-gh-pmotacki:
Hi @sfc-gh-ext-simba-jy.  Regarding the plan for BCR release in the next version, we have an option to remove the old parameter at all, not only set it deprecated.

sfc-gh-ext-simba-jy:
@sfc-gh-pmotacki, I removed the old option (insecureConnect). Please have a look.

sfc-gh-pmotacki (approved):
LGTM

PR #982, opened 2024-12-07
SNOW-1820372: (from 947) when proxy is set in Connection, driver does send traffic through the proxy to S3, but not to Azure blob / GCS bucket (only Snowflake). Works with proxy envvar (GCP)

sfc-gh-ext-simba-jy:
### Description

### Checklist


index.d.ts
```diff
@@ -512,6 +512,11 @@ declare module 'snowflake-sdk' {
          *  The option to pass passcode from DUO.
          */
         passcode?: string;
+
+        /**
+         *  The option to override the environment proxy value (HTTPS_PROXY) if the connection proxy is configured in GCS
+         */
+        overrideEnvProxy?: string;
```
sfc-gh-dszmolka, line 519: probably i'm missing something but since we anyways [validate whether the configured value is a boolean](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/982/files#diff-549a4dcefafa5648ba3e29b9b56eb7fb0de5ee9ef37f56344f30d929d04f3319R87) , i don't understand why we expecting a string here

lib/proxy_util.js
```diff
@@ -55,20 +55,20 @@ exports.getCompareAndLogEnvAndAgentProxies = function (agentOptions) {
       ' protocol=' + agentOptions.protocol + ' proxy=' + proxyHostAndPort
       : ' proxy=' + proxyHostAndPort;
     const proxyUsername = agentOptions.user ? ' user=' + agentOptions.user : '';
+    const proxyString = `${Util.exists(proxyUsername) ? `${agentOptions.user}:${agentOptions.password}@` : ''}${proxyHostAndPort}`.toLowerCase();
```
sfc-gh-dszmolka, line 58: using `proxyHost: 192.168.100.1` and `proxyPort: 80` this ends up being `undefined:undefined@192.168.100.1:80` and i'm unsure if it's expected or not

index.d.ts
sfc-gh-ext-simba-jy: My bad. It must be the boolean type.

lib/proxy_util.js
sfc-gh-ext-simba-jy: I fixed it. Thank you for letting me know.

lib/file_transfer_agent/gcs_util.js
```diff
@@ -37,15 +43,19 @@ function GCSLocation(bucketName, path) {
...
  * @constructor
  */
-function GCSUtil(httpclient, filestream) {
-  const axios = typeof httpclient !== 'undefined' ? httpclient : require('axios');
-  const fs = typeof filestream !== 'undefined' ? filestream : require('fs');
+function GCSUtil(connectionConfig, httpClient, fileStream) {
+  let axios = httpClient;
+  const fs = typeof fileStream !== 'undefined' ? fileStream : require('fs');
+  const connectionProxy = connectionConfig.getProxy();
+  const originalHttpsProxy = process.env[HTTPS_PROXY];
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I thought the storage used the HTTPS_PROXY, but it is not. In storage, HTTP_PROXY is prioritized. I need to fix my code.

index.d.ts
```diff
@@ -512,6 +512,11 @@ declare module 'snowflake-sdk' {
          *  The option to pass passcode from DUO.
          */
         passcode?: string;
+
+        /**
+         *  The option to override the environment proxy value (HTTPS_PROXY) if the connection proxy is configured in GCS
+         */
+        overrideEnvProxy?: boolean;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Changed to overrideGCSEnvProxy.

lib/file_transfer_agent/gcs_util.js
```diff
@@ -360,28 +422,41 @@ function GCSUtil(httpclient, filestream) {
...
-          .bucket(gcsLocation.bucketName)
-          .file(gcsLocation.path + meta['srcFileName'])
-          .getMetadata();
+          metadata = await meta['client'].gcsClient
+            .bucket(gcsLocation.bucketName)
+            .file(gcsLocation.path + meta['srcFileName'])
+            .getMetadata();
+          Util.restoreEnvVar(HTTPS_PROXY, originalHttpsProxy);
+        } else {
+          await meta['client'].gcsClient
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I refactored it.

lib/file_transfer_agent/gcs_util.js
```diff
@@ -287,20 +332,37 @@ function GCSUtil(httpclient, filestream) {
...
+                  [MATDESC_KEY]: gcsHeaders[GCS_METADATA_MATDESC_KEY],
+                  [SFC_DIGEST]: gcsHeaders[GCS_METADATA_SFC_DIGEST]
+                }
               }
-            }
-          });
+            });
+          Util.restoreEnvVar(HTTPS_PROXY, originalHttpsProxy);
+        } else {
+          await meta['client'].gcsClient
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I refactored it.

lib/file_transfer_agent/gcs_util.js
```diff
@@ -37,15 +41,24 @@ function GCSLocation(bucketName, path) {
 /**
  * Creates an GCS utility object.
  *
- * @param {module} httpclient
- * @param {module} filestream
+ * @param {module} httpClient
+ * @param {module} fileStream
```
sfc-gh-dprzybysz, line 45: the first parameter is missing in documentation

lib/file_transfer_agent/gcs_util.js
```diff
@@ -37,15 +41,24 @@ function GCSLocation(bucketName, path) {
...
  *
  * @returns {Object}
  * @constructor
  */
-function GCSUtil(httpclient, filestream) {
-  const axios = typeof httpclient !== 'undefined' ? httpclient : require('axios');
-  const fs = typeof filestream !== 'undefined' ? filestream : require('fs');
+function GCSUtil(connectionConfig, httpClient, fileStream) {
```
sfc-gh-dprzybysz, line 50: @sfc-gh-pmotacki do we consider it as a breaking change?

lib/file_transfer_agent/gcs_util.js
```diff
@@ -37,15 +41,24 @@ function GCSLocation(bucketName, path) {
...
+  let axios = httpClient;
```
sfc-gh-dprzybysz, line 51: do we need this variable?

lib/file_transfer_agent/gcs_util.js
```diff
@@ -57,7 +70,15 @@ function GCSUtil(httpclient, filestream) {
   this.createClient = function (stageInfo) {
     const stageCredentials = stageInfo['creds'];
     const gcsToken = stageCredentials['GCS_ACCESS_TOKEN'];
-
+    //TODO: SNOW-1789759 the value is hardcoded now, but it should be server driven
+    const isRegionalUrlEnabled = (stageInfo.region).toLowerCase() === 'me-central2' || stageInfo.useRegionalUrl;
```
sfc-gh-dprzybysz, line 74: this change should be already included in the master branch and released, I don't think it should be part of this PR

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -41,15 +43,17 @@ function RemoteStorageUtil(connectionConfig) {
   * @returns {Object}
   */
   this.getForStorageType = function (type) {
+    if (client) {
+      return client;
+    }
     if (type === 'S3') {
-      return new SnowflakeS3Util(connectionConfig);
+      client = new SnowflakeS3Util(connectionConfig);
```
sfc-gh-dprzybysz, line 50: why is this change needed? cannot we return as previously?

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -41,15 +43,17 @@ function RemoteStorageUtil(connectionConfig) {
...
     } else if (type === 'AZURE') {
-      return new SnowflakeAzureUtil(connectionConfig);
+      client = new SnowflakeAzureUtil(connectionConfig);
     } else if (type === 'GCS') {
-      return new SnowflakeGCSUtil();
-    } else {
-      return null;
+      client = new SnowflakeGCSUtil(connectionConfig);
     }
+    return client;
```
sfc-gh-dprzybysz, line 56: when null - should we log sth?

lib/proxy_util.js
```diff
@@ -77,6 +77,12 @@ exports.getCompareAndLogEnvAndAgentProxies = function (agentOptions) {
   return logMessages;
 };
 
+/**
+ * Validate whether the proxy object has the appropriate information
+ *
+ * @param proxy
+ * @returns {object}
```
sfc-gh-dprzybysz, line 84: could we describe this object? the same also in other documentati on enrichments

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: If we return the client as before, a new instance will be created each time. In remote_storage_util, every method calls getForStorageType, which results in each function having a different Util instance. I believe this is not the behavior we want. WDYT @sfc-gh-dprzybysz?

lib/file_transfer_agent/gcs_util.js
sfc-gh-ext-simba-jy: My answer to this question is yes. The reason is that we need to control the proxy, similar to how we manage connections in Axios. This means we need to configure https_proxy_agents. To achieve this, we require the destination (endPoint) for the agent, which prevents us from initializing the Axios instance during module loading since the destination information is unavailable at that time. Additionally, the httpClient is necessary for unit testing purposes.

lib/file_transfer_agent/gcs_util.js
sfc-gh-ext-simba-jy: Previously, this implementation was within the if statement that checks for the existence of a GCS token. I moved it to the beginning of the method because the endpoint value is needed to initialize the Axios instance.

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: We have returned null if the type does not match any of the available Utils. As far as I understand, this implies that the driver will default to using local storage. Do you think adding a log for this scenario would be necessary?

lib/file_transfer_agent/gcs_util.js
sfc-gh-ext-simba-jy: Fixed.

lib/proxy_util.js
sfc-gh-ext-simba-jy: I defined the Proxy object with typedef and we can see the details.

lib/file_transfer_agent/gcs_util.js
sfc-gh-pmotacki: According to information in the release notes for version 2.0.0, whether the API isn't included in the typing file, we treated it as internal.

lib/file_transfer_agent/gcs_util.js
```diff
@@ -68,20 +79,27 @@ function GCSUtil(httpclient, filestream) {
...
-        endPoint = `storage.${stageInfo.region.toLowerCase()}.rep.googleapis.com`;
-      }
-      const storage = endPoint ? new Storage({ interceptors_: interceptors, apiEndpoint: endPoint }) : new Storage({ interceptors_: interceptors });
+      const storage = Util.exists(endPoint) ? new Storage({ interceptors_: interceptors, apiEndpoint: endPoint }) : new Storage({ interceptors_: interceptors });
       client = { gcsToken: gcsToken, gcsClient: storage };
     } else {
       client = null;
     }
+  
+    if (typeof httpClient === 'undefined') {
```
sfc-gh-pmotacki, line 88: The code of setting up axios could be extracted to the separated method.

lib/util.js
```diff
@@ -740,6 +740,10 @@ exports.getEnvVar = function (variable) {
   return process.env[variable.toLowerCase()] || process.env[variable.toUpperCase()];
 };
 
+exports.restoreEnvVar = function (envVar, value) {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: This function is for the Azure proxy and I made testing cases in the Azure PR. Because of the 'useEnvProxy' option, we still need this function. Here are the testing case: https://github.com/snowflakedb/snowflake-connector-nodejs/blob/08d802e696a7c3a594dc0e4935279e014d9ff5f1/test/unit/proxy_util_test.js#L301

lib/file_transfer_agent/gcs_util.js
sfc-gh-ext-simba-jy: Done.

lib/file_transfer_agent/gcs_util.js
```diff
@@ -466,6 +476,25 @@ function GCSUtil(httpclient, filestream) {
     }
     return endPoint;
   };
+
+  this.setupHttpClient = function (endPoint) {
+    if (typeof httpClient === 'undefined') {
+      const proxy = ProxyUtil.getProxy(connectionConfig.getProxy(), 'GCS Util');
+
+      //When http_proxy is enabled, the driver should use Axios for HTTPS requests to avoid relying on HTTP_PROXY in GCS.
+      if (proxy || Util.getEnvVar('http_proxy')) {
```
sfc-gh-pmotacki, line 485: Why do we need the second part in the condition?

lib/file_transfer_agent/gcs_util.js
```diff
@@ -142,14 +155,14 @@ function GCSUtil(httpclient, filestream) {
         let matDescKey;
 
         try {
-          if (shouldPerformGCPBucket(accessToken)) {
+          if (shouldPerformGCPBucket(accessToken) && !isProxyEnabled) {
             const gcsLocation = this.extractBucketNameAndPath(meta['stageInfo']['location']);
-
-            const metadata = await meta['client'].gcsClient
+            const  metadata = await meta['client'].gcsClient
```
sfc-gh-dprzybysz, line 160: it's strange that lint is not complaining about such whitespace changes - close after reading

sfc-gh-dprzybysz (approved):

lib/file_transfer_agent/gcs_util.js
sfc-gh-ext-simba-jy: Yes, it is weird. I removed the whitespace.

PR #1012, opened 2025-02-18
SNOW-1825789 Secure token cache

sfc-gh-astachowski:


lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,82 +1,233 @@
...
+        await fs.rmdir(file);
+      } catch (err) {
+        Logger.getInstance().warn('Failed to remove stale file. Error: %s', err.message);
+      }
+    }
+
+  };
+
+
+  this.withFileLocked = async function (fun) {
```
sfc-gh-pmotacki, line 132: Should we treat fun as a callback?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,82 +1,233 @@
...
 const fs = require('node:fs/promises');
-const os = require('os');
 const Util = require('../../util');
-const { validateOnlyUserReadWritePermissionAndOwner } = require('../../file_util');
+const os = require('os');
+const crypto = require('crypto');
+const { getSecureHandle } = require('../../file_util');
+
+function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
+  const topLevelKey = 'tokens';
```
sfc-gh-pmotacki, line 10: It would be more readable for me if the name indicated its relation to the token.

lib/file_util.js
```diff
@@ -164,7 +164,9 @@ exports.validateOnlyUserReadWritePermissionAndOwner = async function (filePath,
     if (octalPermissions === '600') {
       Logger.getInstance().debug(`Validated that the user has only read and write permission for file: ${filePath}, Permission: ${permission}`);
     } else {
-      throw new Error(`Invalid file permissions (${octalPermissions} for file ${filePath}). Make sure you have read and write permissions and other users do not have access to it. Please remove the file and re-run the driver.`);
+      await fsp.chmod(filePath, 0o600).catch(() => {
```
sfc-gh-pmotacki, line 167: Why are we changing permission for existing file? Potentially, it could be modified by an attacker.

lib/file_util.js
```diff
@@ -183,6 +185,51 @@ exports.validateOnlyUserReadWritePermissionAndOwner = async function (filePath,
...
+    const fileHandle = await fsp.open(filePath, flags | fsp.constants.O_NOFOLLOW, 0o600);
+    if (os.platform() === 'win32') {
+      return fileHandle;
+    }
+    const stats = await fileHandle.stat();
+    const mode = stats.mode;
+    const permission = mode & 0o777;
+
+    //This should be 600 permission, which means the file permission has not been changed by others.
+    const octalPermissions = permission.toString(8);
```
sfc-gh-pmotacki, line 207: Isn't it duplication of validateOnlyUserReadWritePermissionAndOwner method?

test/integration/testCache.js
```diff
@@ -21,16 +21,6 @@ describe('Validate cache permissions test', async function () {
       await fs.unlink(validPermissionsFilePath);
     });
 
-    it('should return error on insecure permissions', async function () {
```
sfc-gh-pmotacki, line 24: Should we remove this test? why?

test/unit/authentication/json_credential_manager_test.js
```diff
@@ -1,23 +1,47 @@
...
 const key = Util.buildCredentialCacheKey(host, user, credType);
+const key2 = Util.buildCredentialCacheKey(host, user, credType2);
 const randomPassword = randomUUID();
-const os = require('os');
+const randomPassword2 = randomUUID();
+
+const pathFromHome = function () {
+  switch (process.platform) {
+  case 'win32':
+    return ['AppData', 'Local', 'Snowflake', 'Caches'];
```
sfc-gh-pmotacki, line 20: We use the paths in JsonCredentialManager. What du you thin about extracting subfolders to some variables?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,82 +1,233 @@
...
+        await fs.mkdir(cacheDir, options);
+        return true;
+      } else {
+        if (process.platform === 'win32') {
+          return true;
+        }
+        if ((cacheStat.mode & 0o777) === 0o700) {
+          return true;
+        }
+        await fs.chmod(cacheDir, 0o700);
```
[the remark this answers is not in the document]
sfc-gh-astachowski: This is in accordance with ["Design of Authentication Token Caching"], section "Location"(https://docs.google.com/document/d/1taFHFZNaKr3D5kPwOJw-0PDGoFKB2TFaMtx6gSZ5iY4/edit?usp=sharing)

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,82 +1,233 @@
...
+  this.tryTokenDir = async function (dir, subDirs) {
+    const cacheDir = path.join(dir, ...subDirs);
+    try {
+      const stat = await fs.stat(dir);
+      if (!stat.isDirectory()) {
+        Logger.getInstance().info(`Path ${dir} is not a directory`);
+        return false;
+      }
+      const cacheStat = await fs.lstat(cacheDir).catch(() => {});
+      if (!Util.exists(cacheStat)) {
```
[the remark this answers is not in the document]
sfc-gh-astachowski: Looking back at it, I should probably check if the error is ENOENT and only ignore that one.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Not exactly, we have some cleanup to do after (removing lock)

sfc-gh-astachowski: Well, I'm not sure how you define callback. It is called within the function, but not at the end of the function, as we need to cleanup after

lib/file_util.js
sfc-gh-astachowski: It is very similar, but operates on file handles in order to mitigate any manipulations between our checks and operations. `validateOnlyUserReadWritePermissionAndOwner` should probably be replaced with this later on, but I didn't want to do it in this PR. See https://snowflakecomputing.atlassian.net/browse/SNOW-1944224

test/integration/testCache.js
sfc-gh-astachowski: This was removed because of the change to the tested function that attempted to change the permissions. I'll restore it together with reverting the mentioned change

test/unit/authentication/json_credential_manager_test.js
sfc-gh-astachowski: Good call, done

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+      }
+      const cacheStat = await fs.lstat(cacheDir).catch((err) => {
+        if (err.code !== 'ENOENT') {
+          throw err;
+        }
+      });
+      if (!Util.exists(cacheStat)) {
+        const options = { recursive: true };
+        if (process.platform !== 'win32') {
+          options.mode = 0o700;
```
sfc-gh-jszczerbinski, line 69: We are using here 0o700 permissions and recursive. This is problematic since for example:
- `0o700` is too narrow for $HOME/.cache directory (this directory has usually 0755 on mac and linux systems)

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+        }
+        await fs.mkdir(cacheDir, options);
+        return true;
+      } else {
+        if (!stat.isDirectory()) {
```
sfc-gh-jszczerbinski, line 74: stat or cacheStat?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+          return false;
+        }
+        if (process.platform === 'win32') {
+          return true;
+        }
+        if ((cacheStat.mode & 0o777) === 0o700) {
```
sfc-gh-jszczerbinski, line 80: We should also check owner

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+    return crypto.createHash('sha256').update(key).digest('hex');
+  };
+
+  this.getTokenDirCandidates = function () {
+    const candidates = [];
+    if (Util.exists(credentialCacheDir)) {
+      candidates.push({ folder: credentialCacheDir, subfolders: [] });
+    }
+    const sfTemp = process.env.SF_TEMPORARY_CREDENTIAL_CACHE_DIR;
+    if (Util.exists(sfTemp)) {
```
sfc-gh-jszczerbinski, line 28: Checking if this `exists` seems like redundant code. Can we do it in `tryTokenDir`?

lib/file_util.js
```diff
@@ -183,6 +183,49 @@ exports.validateOnlyUserReadWritePermissionAndOwner = async function (filePath,
...
+    const fileHandle = await fsp.open(filePath, flags, 0o600);
+    if (os.platform() === 'win32') {
+      return fileHandle;
+    }
+    const stats = await fileHandle.stat();
+    const mode = stats.mode;
+    const permission = mode & 0o777;
+
+    //This should be 600 permission, which means the file permission has not been changed by others.
+    const octalPermissions = permission.toString(8);
```
sfc-gh-jszczerbinski, line 205: Why convert it to string? just compare with octal value?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
-      throw new Error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location(home). 
+      throw new Error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location. 
       Please set 'credentialCacheDir' connection configuration option to enable the default credential manager.`);
     }
 
-    const tokenCacheFile = path.join(tokenDir, 'temporary_credential.json');
-    await validateOnlyUserReadWritePermissionAndOwner(tokenCacheFile);
-    return tokenCacheFile;
+    const tokenCacheFile = path.join(tokenDir, 'credential_cache_v1.json');
+    return [await getSecureHandle(tokenCacheFile, fs.constants.O_RDWR | fs.constants.O_CREAT, fs), tokenCacheFile];
```
sfc-gh-jszczerbinski, line 114: I have an issue with how this function is used and O_CREAT flag. It seems to me that it will create cache file (empty one) during read-only operations (retrieving key), and creating file lock

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+      await new Promise(resolve => setTimeout(resolve, 100));
+    }
+    if (!locked) {
+      if (Util.exists(fileHandle)) {
+        await fileHandle.close();
+      }
+      Logger.getInstance().warn('Could not acquire lock on cache file %s', file);
+      return null;
+    }
+    const res = await fun(fileHandle, file);
```
sfc-gh-jszczerbinski, line 173: Nit: Can you more the file locking functionality to a seperate object/functions? I would like this function to look similar to this:
```
const lck = lockFile()
const res = func(fileHandle, file)
lck.unlock()
```

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+    const sfTemp = process.env.SF_TEMPORARY_CREDENTIAL_CACHE_DIR;
+    if (Util.exists(sfTemp)) {
+      candidates.push({ folder: sfTemp, subfolders: [] });
+    }
+    const xdgCache = process.env.XDG_CACHE_HOME;
+    if (Util.exists(xdgCache) && process.platform === 'linux') {
+      candidates.push({ folder: xdgCache, subfolders: ['snowflake'] });
+    }
+    const home = process.env.HOME;
+    switch (process.platform) {
```
sfc-gh-jszczerbinski, line 36: Nit: it's hard to see what's returned for each system. Can you do it like this
```
switch(proces.platform) {
  case 'win32':
    return [ ... ]
  case 'linux':
    return [ ... ]
  case 'darwin':
    return [ ... ]
}
```

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+  this.tryTokenDir = async function (dir, subDirs) {
+    const cacheDir = path.join(dir, ...subDirs);
+    try {
+      const stat = await fs.stat(dir);
+      if (!stat.isDirectory()) {
+        Logger.getInstance().info(`Path ${dir} is not a directory`);
+        return false;
+      }
+      const cacheStat = await fs.lstat(cacheDir).catch((err) => {
+        if (err.code !== 'ENOENT') {
```
sfc-gh-jszczerbinski, line 62: Maybe create dir here when ENOENT? Then stat it again and recover from error?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,252 @@
...
+    }
+    const res = await fun(fileHandle, file);
+    if (Util.exists(fileHandle)) {
+      await fileHandle.close();
+    }
+    await fs.rmdir(lckFile);
+    return res;
+  };
   
   this.write = async function (key, token) {
```
sfc-gh-jszczerbinski, line 181: It seems like we don't check directory permissions again here

sfc-gh-jszczerbinski: Similar in other cache operations

sfc-gh-jszczerbinski (changes requested):

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Fair, changed to creating recursively with `0o755`, then changing the final directory's permission to `0o700`

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Good catch, fixed

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Added

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Done

lib/file_util.js
sfc-gh-astachowski: This was copied from an analogous function, fixed

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Good point, changed

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Changed it a bit, is this better?

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Didn't do exactly that, but without the null/undefined checks the function is much more readable now

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: Good point, changed that

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-astachowski: I don't see what you mean. We check directory permissions in try token dir, and they aren't breaking either way - worst case scenario we won't log a warning if the permissions change in between then and here.

lib/authentication/secure_storage/json_credential_manager.js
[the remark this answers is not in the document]
sfc-gh-astachowski: Removed this after discussion

lib/file_util.js
sfc-gh-astachowski: No longer attempting to change file permissions

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -21,68 +21,62 @@ function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
...
+      const cacheStat = await fs.stat(cacheDir).catch(async (err) => {
         if (err.code !== 'ENOENT') {
           throw err;
         }
-      });
-      if (!Util.exists(cacheStat)) {
         const options = { recursive: true };
         if (process.platform !== 'win32') {
-          options.mode = 0o700;
+          options.mode = 0o755;
```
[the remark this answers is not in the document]
sfc-gh-astachowski: The idea is that the folders along the way should have 755 permissions, while the folder directly containing the cache should have 700. I think this is the best way to do it - could alternatively create all but the last one with 755 permissions, then create the last one separately with 700

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -21,68 +21,62 @@ function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
...
         return false;
       }
-      const cacheStat = await fs.lstat(cacheDir).catch((err) => {
+      const cacheStat = await fs.stat(cacheDir).catch(async (err) => {
         if (err.code !== 'ENOENT') {
           throw err;
         }
-      });
-      if (!Util.exists(cacheStat)) {
         const options = { recursive: true };
```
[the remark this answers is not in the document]
sfc-gh-astachowski: Good point, done

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -95,26 +89,27 @@ function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
...
 
     if (!Util.exists(tokenDir)) {
       throw new Error(`Temporary credential cache directory is invalid, and the driver is unable to use the default location. 
       Please set 'credentialCacheDir' connection configuration option to enable the default credential manager.`);
     }
 
-    const tokenCacheFile = path.join(tokenDir, 'credential_cache_v1.json');
-    return [await getSecureHandle(tokenCacheFile, fs.constants.O_RDWR | fs.constants.O_CREAT, fs), tokenCacheFile];
+    return path.join(tokenDir, 'credential_cache_v1.json');
+    //return [await getSecureHandle(tokenCacheFile, fs.constants.O_RDWR, fs), tokenCacheFile];
```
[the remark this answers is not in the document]
sfc-gh-astachowski: Oops 😄 Removed

lib/file_util.js
sfc-gh-pmotacki: The issue you attached doesn't address the migration of control for configuration files. Do we need to create a new issue?

sfc-gh-pmotacki (approved):
One new comment to create follow-up issue. 
Code LGTM.

lib/authentication/secure_storage/json_credential_manager.js
sfc-gh-jszczerbinski: Looks good

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -184,18 +186,20 @@ function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
...
       const jsonCredential = await this.readJsonCredentialFile(fileHandle) || {};
+      await closeHandle(fileHandle);
       if (!Util.exists(jsonCredential[tokenMapKey])) {
         jsonCredential[tokenMapKey] = {};
       }
       jsonCredential[tokenMapKey][keyHash] = token;
 
       try {
-        const flag = Util.exists(fileHandle) ? fs.constants.O_RDWR | fs.constants.O_CREAT : fs.constants.O_WRONLY;
+        const flag = Util.exists(fileHandle) ? fs.constants.O_WRONLY : fs.constants.O_RDWR | fs.constants.O_CREAT;
```
sfc-gh-jszczerbinski, line 199: Why ternary? isn't it always  O_RDWR | O_CREAT?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -209,8 +213,10 @@ function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
 
     const keyHash = this.hashKey(key);
 
-    return await this.withFileLocked(async (fileHandle) => {
+    return await this.withFileLocked(async (filename) => {
+      const fileHandle = await getSecureHandle(filename, fs.constants.O_RDWR, fs);
```
sfc-gh-jszczerbinski, line 217: O_WRONLY?

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -226,16 +232,18 @@ function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
 
     const keyHash = this.hashKey(key);
 
-    await this.withFileLocked(async (fileHandle, filename) => {
+    await this.withFileLocked(async (filename) => {
+      const fileHandle = await getSecureHandle(filename, fs.constants.O_RDWR, fs);
```
sfc-gh-jszczerbinski, line 236: O_WRONLY

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -226,16 +232,18 @@ function JsonCredentialManager(credentialCacheDir, timeoutMs = 60000) {
...
       const jsonCredential = await this.readJsonCredentialFile(fileHandle);
+      await closeHandle(fileHandle);
 
       if (jsonCredential && jsonCredential[tokenMapKey] && jsonCredential[tokenMapKey][keyHash]) {
         try {
           jsonCredential[tokenMapKey][keyHash] = null;
-          const flag = Util.exists(fileHandle) ? fs.constants.O_RDWR | fs.constants.O_CREAT : fs.constants.O_WRONLY;
+          const flag = Util.exists(fileHandle) ? fs.constants.O_WRONLY : fs.constants.O_RDWR | fs.constants.O_CREAT;
```
sfc-gh-jszczerbinski, line 243: fileHandle never exists here?

sfc-gh-jszczerbinski (approved):
Looks good, but please look at the comments before merging

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,260 @@
...
+    while (attempts <= 10) {
+      Logger.getInstance().debug('Attempting to get a lock on file %s, attempt: %d', filename, attempts);
+      attempts++;
+      await fs.mkdir(lckFile, options).then(() => {
+        locked = true;
+      }, () => {});
+      if (locked) {
+        break;
+      }
+      await new Promise(resolve => setTimeout(resolve, 100));
```
sfc-gh-pmotacki, line 160: Should we extract the timeout value to const?

sfc-gh-astachowski: Done

sfc-gh-ext-simba-jy (commented):
In Node.js, using await with .then() is usually unnecessary. Also, if mkdir is rejected, an error will be thrown because there is no try...catch block in my opinion. @sfc-gh-pmotacki, Could you please confirm this? I think this will throw an error if the first mkdir attempt fails.

lib/authentication/secure_storage/json_credential_manager.js
```diff
@@ -1,87 +1,261 @@
...
+    let attempts = 1;
+    let locked = false;
+    const options = {};
+    if (process.platform !== 'win32') {
+      options.mode = 0o600;
+    }
+    while (attempts <= 10) {
+      Logger.getInstance().debug('Attempting to get a lock on file %s, attempt: %d', filename, attempts);
+      attempts++;
+      await fs.mkdir(lckFile, options).then(() => {
```
sfc-gh-ext-simba-jy, line 155: In Node.js, using await with .then() is usually unnecessary. Also, if mkdir is rejected, an error will be thrown because there is no try...catch block. @sfc-gh-pmotacki, could you please confirm this? In my opinion.  this will throw an error if the first mkdir attempt fails and no retry.

sfc-gh-astachowski: Notice this is then with two arguments - the second one is onReject (so effectively catch)

PR #1055, opened 2025-04-16
SNOW-1791642: Util.getProxyFromEnv is always isHttps, causing HTTP_PROXY value to be ignored for traffic destined to HTTP / port 80

sfc-gh-ext-simba-jy:
### Description
- [x] Added a new connectionconfig option called useProxyForOCSP (default is false).
- [x] If connection proxy is not set for OCSP (either useProxyForOCSP is false or no proxy info is provided), the agent will be created with HTTP_PROXY env (if available) unless isEnvProxyEnable is false.
- [x] Edited the testing cases.  

### Checklist


sfc-gh-pbulawa (changes requested):

lib/proxy_util.js
```diff
@@ -307,6 +308,22 @@ exports.restoreEnvironmentProxy = function () {
...
+ * @param proxy
+ * @param destination {string | Regex } 
+ */
+exports.isByPassProxy = function (proxy, destination) {
+  if (proxy && proxy.noProxy) {
+    const bypassList = proxy.noProxy.split('|');
+    return bypassList.find((noProxy) => {
+      let host = noProxy.trim();
+      host = host.replace('*', '.*?');
+      return destination instanceof RegExp ? destination.test(host) : destination.match(host);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I refactored it. On the top of your opinion, I make the each noProxy values to Regex and test it with the destination. Please have a look. Thank you!

lib/agent/https_proxy_agent.js
```diff
@@ -22,7 +23,8 @@ class SnowflakeHttpsProxyAgent extends HttpsProxyAgent {
     Logger.getInstance().debug('Using proxy=%s for host %s', this.proxy.hostname, opts.host);
     const socket = await super.connect(req, opts);
     if (socket instanceof tls.TLSSocket) {
-      SocketUtil.secureSocket(socket, opts.host, this);
+      const isProxyRequiredForOCSP = this.proxy.useForOCSP && !ProxyUtil.isByPassProxy(this.proxy, /.snowflakecomputing./);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I changed it.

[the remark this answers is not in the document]
sfc-gh-pbulawa: We now have the const duplicated here and in `socket_util.js`. Can you import the const from the latter one?

lib/proxy_util.js
```diff
@@ -307,6 +308,26 @@ exports.restoreEnvironmentProxy = function () {
...
+ * Provide the details of the proxy info.
+ * @param proxy
+ * @param destination {string | Regex } 
+ */
+exports.isByPassProxy = function (proxy, destination) {
+  if (proxy && proxy.noProxy) {
+    const bypassList = proxy.noProxy.split('|');
+    return bypassList.find((noProxy) => {
+      let host = noProxy.trim();
+      host = host.replace('*', '.*?');
```
sfc-gh-pbulawa, line 321: Could you please proceed with fixing the pattern as per CodeQL

lib/proxy_util.js
```diff
@@ -307,6 +308,26 @@ exports.restoreEnvironmentProxy = function () {
...
+      if (destination instanceof RegExp) {
+        return destination.test(host);
+      }
+      const noProxyRegex = new RegExp(`^${host}`);
```
sfc-gh-pbulawa, line 325: Can you fix this part as per CodeQL.

lib/proxy_util.js
sfc-gh-ext-simba-jy: Actually, this code is from my implementation, not the original one. In my opinion, the original intention was to replace only the first wildcard in the URL. For example, if the user sets *.snowflakecomputing.com, it should block all subdomains. That's why we only need to replace the first *. I do not think the value like '*.snowflakecomputing.*' is not an appropriate value for noProxy. What do you think?

lib/proxy_util.js
sfc-gh-ext-simba-jy: This is the tricky part. It seems like the CodeQL worries that the host can be varied because we get the host value from the environment variable. However, we should get the values of the HTTPS_PROXY, HTTP_PROXY, and NO_PROXY from the environment variable unless the proxy information is directly provided in the connection. What do you think?

lib/proxy_util.js
sfc-gh-pbulawa: That makes sense. Thanks for the explanation.

lib/proxy_util.js
sfc-gh-pbulawa: That makes sense, thanks.

sfc-gh-pbulawa (approved):

index.d.ts
```diff
@@ -332,6 +332,11 @@ declare module 'snowflake-sdk' {
          */
         proxyProtocol?: string;
 
+        /**
+         * Specifies whether the OCSP request is also sent to the proxy specified. 
+         */
+        useProxyForOCSP?: boolean;
```
sfc-gh-dszmolka, line 338: as discussed on Slack; can we please consider using `useConnectionConfigProxyForOCSP` or similar name for the flag; to better describe what it actually does?

sfc-gh-ext-simba-jy: Ohhh, I missed that part. I will replace the name soon. Thank you!

lib/agent/https_proxy_agent.js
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Done. Sorry, I missed this part.

PR #1060, opened 2025-04-25
SNOW-1825499 - added experimental flag for new oauth authentication f…

sfc-gh-pmotacki:
…lows

### Description

### Checklist


lib/authentication/authentication.js
```diff
@@ -49,6 +49,14 @@ exports.formAuthJSON = function formAuthJSON(
   return body;
 };
 
+function throwErrorIfExperimentalAuthnNotEnabled(connectionConfig) {
```
sfc-gh-mkubik, line 52: I don't like the naming since it won't throw for any other authenticator except PAT regardless of experimental flag right? Maybe something like throw if authenticator not allowed would be more indicative? Or for me this logic doesn't necessarily need to be extracted to a separate function.

lib/connection/connection_config.js
```diff
@@ -493,6 +493,15 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
     disableSamlURLCheck = options.disableSamlURLCheck;
   }
 
+
+  let enableExperimentalAuthentication = false;
+  if (Util.exists(options.enableExperimentalAuthentication)) {
+    Errors.checkArgumentValid(Util.isBoolean(options.enableExperimentalAuthentication),
+      ErrorCodes.ERR_CONN_CREATE_INVALID_DISABLE_SAML_URL_CHECK);
```
sfc-gh-mkubik, line 500: I guess this error code is a leftover from the function above

lib/authentication/authentication.js
sfc-gh-pmotacki: Agree, Fixed

lib/connection/connection_config.js
sfc-gh-pmotacki: Fixed

sfc-gh-mkubik (approved):

sfc-gh-dheyman (approved):

PR #1073, opened 2025-05-13
Add prettier and format the repo

sfc-gh-rsavenok:
### Description
Adds `prettier` and formats all files in the repo.

#### Why? (Pros)
* 0 config code formatter for all our files
* it's formatting/checking is faster than eslint
* eslint deprecated all code formatting rules inthe  newest version

#### Cons of using prettier?
* Its highly opionated about how the code should be formatted, so unlike eslint, rules can't be disabled in specific places

### Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,6 @@
 Please explain the changes you made here.
 
 ### Checklist
-- [ ] Format code according to the existing code style (run `npm run lint:check -- CHANGED_FILES` and fix problems in changed code)
 - [ ] Create tests which fail without the change (if possible)
 - [ ] Make all tests (unit and integration) pass (`npm run test:unit` and `npm run test:integration`)
 - [ ] Extend the types in index.d.ts file (if necessary)
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  singleQuote: true,
+  tabWidth: 2,
+  endOfLine: 'lf',
+  printWidth: 100,
+};
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "mocha": "^11.1.0",
     "mock-require": "^3.0.3",
     "nyc": "^15.1.0",
+    "prettier": "^3.5.3",
     "test-console": "^2.0.0",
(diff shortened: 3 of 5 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,7 +2,6 @@
 Please explain the changes you made here.
 
 ### Checklist
-- [ ] Format code according to the existing code style (run `npm run lint:check -- CHANGED_FILES` and fix problems in changed code)
```
sfc-gh-rsavenok, line 5: Removed this one as our GH actions will fail anyway if the code is not linted

.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  singleQuote: true,
+  tabWidth: 2,
+  endOfLine: 'lf',
+  printWidth: 100,
```
sfc-gh-rsavenok, line 5: The default prettier is 80, but since we have lots of log messages, I think it's reasonable to increase it to 100

.eslintrc.js
```diff
@@ -1,38 +1,38 @@
...
+      },
+    },
   ],
-  'parserOptions': {
-    'ecmaVersion': 'latest'
+  parserOptions: {
+    ecmaVersion: 'latest',
   },
-  'rules': {
+  rules: {
```
sfc-gh-rsavenok, line 21: I removed all rules that are defaults in prettier

package.json
```diff
@@ -58,11 +61,15 @@
     "url": "https://github.com/snowflakedb/snowflake-connector-nodejs"
   },
   "typings": "./index.d.ts",
+  "lint-staged": {
+    "*": "npm run prettier:format",
+    "*.{js,ts}": "npm run lint:check"
+  },
   "scripts": {
-    "lint:check": "eslint && check-dts index.d.ts",
-    "lint:check:all": "eslint lib samples system_test test && check-dts index.d.ts",
```
sfc-gh-rsavenok, line 63: I reduced the number of lint scripts to only those that run on the CI/Precommit hook, because developers shouldn't need to run them manually. 

In rare cases where a developer needs to run something specific, node supports that using `npx eslint ..arguments..` and `npx prettier ..arguments..`.

sfc-gh-dheyman (approved):

PR #1076, opened 2025-05-13
Retire ci/container/package.json

sfc-gh-rsavenok:
### Description
Our current test pipeline does the following:
* copy source code to a `WORKDIR`
* replace `package.json` from the  `WORKDIR` with the one in `WORKDIR/ci/container/package.json`
* `npm install`
* `npm install snowflake-sdk` from the artifact to ensure that our package doesn't crush on instalation
* run tests on our `WORKDIR`, not the `snowflake-sdk` artifact

The step of replacing `package.json` with the one `ci/container/package.json` doesn't do anything, and the only reason it doesn't break our test is because `npm install snowflake-sdk` would extend our `node_modules` with all dependencies that weren't installed because of the replacement of `package.json`

#### Cons of keeping `ci/container/package.json`
* dependencies are out of sync (like mocha 10 vs 11)
* if we'll have post-install hooks, they will need to be copied to this file as well

### Checklist


ci/test_windows.bat
```diff
@@ -67,7 +66,7 @@ start /b python hang_webserver.py 12345 > hang_webserver.out 2>&1
 popd
 
 echo [INFO] Testing
-cmd /c node_modules\.bin\mocha --exit --timeout %TIMEOUT% --recursive --full-trace --color --reporter spec \"test/{unit,integration}/**/*.js\"
+cmd /c node_modules\.bin\mocha --exit --timeout %TIMEOUT% --recursive --full-trace --color --reporter spec test/{unit,integration}/**/*.js
```
sfc-gh-rsavenok, line 69: "" for test files were a hack for a bug in mocha dependency - https://github.com/mochajs/mocha/issues/5341

Mocha 11 doesn't work with this hack

ci/container/test_component.sh
```diff
@@ -24,10 +24,10 @@ if [[ "$LOCAL_USER_NAME" == "jenkins" ]]; then
 else
     export PATH=$WORKSPACE/node_modules/.bin:$PATH
 fi
-cp $SOURCE_ROOT/ci/container/package.json .
 npm install
 
 PACKAGE_NAME=$(cd $WORKSPACE && ls snowflake-sdk*.tgz)
+echo "[INFO] Test $PACKAGE_NAME installation"
```
sfc-gh-rsavenok, line 30: Added this message to indicate that the only reason we install the snowflake-sdk package is to test that it installs (e.g. post install hooks doesnt crush). We don't actually run any tests on this package in this pipeline

sfc-gh-pbulawa (approved):

sfc-gh-akolodziejczyk (approved):
LGTM

PR #1077, opened 2025-05-14
SNOW-2111082 Add sinon for mocking dependencies 

sfc-gh-rsavenok:
### Description
Introduce sinon for mocking dependencies for 2 reasons:
* cleaner api
* We don't need to have function arguments only to be able to mock dependencies in tests

#### Security review
Resulted in LOW https://github.com/snowflakedb/security_reviews/pull/6847

#### Why sinon and not other library?
Sinon is a go-to solution when working with Mocha, it has 7+ million weekly downloads (other libraries has 10x less downloads). It has a rich api allowing to mock both npm dependencies and local dependencies and control the mock lifecycle.

#### Sinon migration
This PR doesn't update every file, because not all of them are straightforward; Would suggest to work on following things only when somebody would need to touch related code:
* s3 - sinon with cjs struggles to mock them because of locked objects. This is fixable with aws-sdk-mock package or typescript migration
* mock_file.js - the whole util needs to be refactored for sinon
* encrypt_util.js - temp file creation dependency needs to be refactored

### North star is to get rid `mock-require` completely in favor of sinon and maybe use proxyrequire (for rare cases where sinon wouldn't be enough)
`mock-require` doesn't do anything for us already. Since we pass dependencies as function arguments, we can just pass plain objects/functions there, but it's useful to keep in the code to find places that need to be migrated to sinon :P 

### Checklist


sfc-gh-akolodziejczyk (approved):
LGTM! is there a plan to migrate other tests to use sinon?

sfc-gh-rsavenok:
> LGTM! is there a plan to migrate other tests to use sinon?
I think it makes sense to migrate other parts only when somebody would touch that part of the code :)

package.json
```diff
@@ -44,6 +44,7 @@
     "mocha": "^11.1.0",
     "mock-require": "^3.0.3",
     "nyc": "^15.1.0",
+    "sinon": "^20.0.0",
```
sfc-gh-rsavenok, line 47: Risk assessment PR in the jira ticket

test/unit/authentication/authentication_test.js
```diff
@@ -362,8 +307,7 @@ describe('key-pair authentication', function () {
   });
 
   it('key-pair - get token with private key path without passphrase', function () {
-    const auth = new AuthKeypair(connectionOptionsKeyPairPath,
-      cryptomod, jwtmod, filesystem);
+    const auth = new AuthKeypair(connectionOptionsKeyPairPath);
```
sfc-gh-pmotacki, line 310: Wasn't it a BCR for customers? I know it's not treated as a public API because it is not in the types file. wdyt?

sfc-gh-rsavenok: I don't think we should treat it as BCR because customers cannot import it into their codebase except through a "hackish" way to access `/lib/anyFile` directly. 


The only entry point we provide to customers is `index.js`, and it doesn't export anything except `core.js`. If we treat it as BCR, basically every internal function change would be a BCR
Screenshot 2025-05-20 at 12 15 18

sfc-gh-dheyman:
Great improvement! LGTM

sfc-gh-pmotacki (approved):

PR #1080, opened 2025-05-19
POC TS support

sfc-gh-rsavenok:
### Description

### Checklist


tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "sourceMap": true,
+    "allowJs": true,
+    "outDir": "./dist",
+    "target": "ES2018",
```
sfc-gh-rsavenok, line 7: TODO: check if we can go higher

lib/util.ts
```diff
@@ -66,6 +66,7 @@ exports.isFunction = function (value) {
   return !!value && typeof value === 'function';
 };
 
+// @ts-ignore TS2451: Cannot redeclare block-scoped variable 'toString'.
```
sfc-gh-rsavenok, line 69: Example of fixing ts errors

lib/__dummyTsFile.ts
```diff
@@ -0,0 +1,8 @@
+import { testJsFn } from './__dummyJsFile';
```
sfc-gh-rsavenok, line 1: TODO: remove before merge

lib/snowflake.ts
```diff
@@ -1,7 +1,9 @@
 const core = require('./core');
 const Util = require('./util');
-const Errors = require('./errors');
-const ErrorCodes = Errors.codes;
+import { codes as ErrorCodes } from './errors';
+const testFn = require('./__dummyTsFile').default;
```
sfc-gh-rsavenok, line 4: TODO: remove before merge

test/dummy_test2.js
```diff
@@ -0,0 +1,7 @@
+const assert = require('assert');
+
+describe('Dummy test2', () => {
```
sfc-gh-rsavenok, line 3: TODO: remove before merge

test_dist.js
```diff
@@ -0,0 +1,3 @@
+const distLib = require('./dist/index.js');
```
sfc-gh-rsavenok, line 1: TODO: remove before merge

.npmignore
```diff
@@ -1,22 +1,5 @@
...
-system_test/
-scripts/
-samples/
-ci/
-.github/
-.eslintrc.js
-.pre-commit-config.yaml
-Jenkinsfile
-whitesource.config
+*
```
sfc-gh-rsavenok, line 1: Since this migration introduces new files, reversion ignore pattern to ignore everything except 1 dir and 3 doc files

ci/test_windows.bat
```diff
@@ -48,6 +48,7 @@ popd
 
 echo [INFO] Installing Test package
 cmd /c npm install
+REM test run script
```
sfc-gh-rsavenok, line 51: We can add a test script here to check that snowflake-sdk node dependency actually runs

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "ES2018",
```
sfc-gh-rsavenok, line 4: TODO: test es2022

lib/util.ts
```diff
@@ -484,11 +487,9 @@ exports.validateClientSessionKeepAliveHeartbeatFrequency = function (input, mast
 };
 
 // driver name
-const driverName = require('./../package.json').name;
```
sfc-gh-rsavenok, line 487: Typescript doesn't validate require() paths, and since in dist this path is different, the compiled js will crush.

This issue won't happen for relative imports in /lib. I think to solve the issue, it would be enough just to add 1 test to check if requiring snowflake-sdk works with no crushes

tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+    "lib": ["ES2018"],
+    "outDir": "./dist",
+    "sourceMap": true,
```
sfc-gh-rsavenok, line 7: See how bad strict mode is. And see if we can use strict mode only on specific files

lib/snowflake.ts
```diff
@@ -1,14 +1,16 @@
...
+import { createClientError, codes as ErrorCodes } from './errors';
+import testFn from './__dummyTsFile';
+
+testFn();
 
 const clientEnvironment = process.versions;
 
 // if we're not using the minimum supported version of node.js, raise an error
-const minimumNodeVersion = '6.0.0';
+const minimumNodeVersion = '18.0.0';
```
sfc-gh-rsavenok, line 11: [!!!!] Our docs say that we support node 18+, es2022 set in typescript targets works on node 16+

https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver

tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "module": "commonjs",
+    "target": "ES2022",
+    "lib": ["ES2022"],
+    "outDir": "./dist",
+    "sourceMap": true,
+    "allowJs": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
+    // NOTE: Ideally, we want strict mode, but it will cause too much @ts-ignore in the codebase
+    // "strict": true
```
sfc-gh-rsavenok, line 12: Let's discuss this when @sfc-gh-dheyman is back. For example, renaming `util.js` to `util.ts` results in 95 strict violations, because we have variables that got casted to "any" type.

ci/container/test_npm_package.js
```diff
@@ -0,0 +1,12 @@
+// NOTE:
```
sfc-gh-rsavenok, line 1: @sfc-gh-akolodziejczyk could you pls take a look at this? It's related to our previous discussion that we don't run any tests against snowflake-sdk package when installing it.

Here's a commit example that would be caught by this script https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1080/commits/2b100701cb275678e42c8eb2a5c646c87c3a788e

tsconfig.json
sfc-gh-rsavenok: Consider reversing and disabling scrict for every existing file

lib/util.ts
sfc-gh-dszmolka: don't we send the driverName (alongside with version) in the application field ; which is then visible in Snowflake SESSION data ? (its a valuable data, the driver name + version, would be great not to lose it)

sfc-gh-rsavenok: I'm not removing this var, I'm just moving the import to top level as `import` instead of `require`.

Reason is, that typescript doesn't adjust path of the `require()` blocks in compile code. Compiled code is located in `/dist` folder, so the `require('./../package.json')` will break. Using import instead of require will copy `package.json` to the `/dist` folder.

A test is added `ci/container/test_npm_package.js` to ensure that this works

lib/util.ts
```diff
@@ -1,10 +1,19 @@
...
+import os from 'os';
+import * as Logger from './logger';
+import fs from 'fs';
+// NOTE: keeping require as it's a circular dependency so * as Errors doesn't work
 const Errors = require('./errors');
-const net = require('net');
+import net from 'net';
+import { name as driverName, version as driverVersion } from '../package.json';
+
+const nodeJSVersion = process.version?.startsWith('v')
```
sfc-gh-rsavenok, line 11: This diff is large, so summing up what I did:
1) replaced require() with imports, allowing to parse correct types for them
2) move CONSTants to top of the file so it's clear that we have 3 of them
3) replaced all exports.name with export const
4) added typings to every function
5) 3 typings were ignored with @ts-ignore
6) remove unneeded bits in JSDoc as typescript types already contain that information

lib/connection/statement.js
```diff
@@ -42,21 +42,14 @@ exports.createContext = function (
 
   context.type = statementTypes.FILE_PRE_EXEC;
 
-  createStatement(options, context, services, connectionConfig);
```
sfc-gh-rsavenok, line 45: #### Explaining this change
When the typescript is in strict mode, it adds JavaScript "use strict" to every file. 

In sloppy mode, `createStatement->this` references to nodejs `global`. In strict mode, it's undefined.

This means that the current `master` branch code **unintentionally leaks BaseStatement methods to the global namespace**:
```
console.log(global.getSqlText);   -> function from BaseStatement
console.log(global.getColumns); -> function from BaseStatement
```

Based on my understanding of the code, its original intention was to extend context with a bunch of data. But since BaseStatement requires `this` for its own internals, the author used `.apply(this` without understanding the consequences. A better way to handle this would be `new BaseStatement` = we still have hackish way to extend context, but at least were not leaking anything.

PR #1096, opened 2025-06-11
[SNOW-1902247] WIF Auth for AWS

sfc-gh-rsavenok:
### Description
Implements experimental WIF auth for AWS.

Besides AWS-SDK dependencies, `rewiremock` was added to mock npm imports. We already have `mock-require` in the project, but it wasn't updated in the past 6 years and doesn't work correctly if target imports land in the require cache before the mocking.

In the future, I'd like to explore migration to a different test runner (jest, vite, or something else) where module mocking is a test runner feature.

The following test was performed on an EC2 instance:
```
const connection = snowflake.createConnection({
  authenticator: 'WORKLOAD_IDENTITY',
  enableExperimentalWorkloadIdentityAuth: true,
  workloadIdentityProvider: 'AWS',
  account: '...',
  accessUrl: '...',

});

(async () => {
  await setupConnection();
  const result = await executeQuery({
    sqlText: 'SELECT 1',
  });
  console.log(result);
})();
```

### Checklist

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -571,13 +571,18 @@ function Connection(context) {
    * @returns {String} the query status.
    */
   function extractQueryStatus(queryResponse) {
-    const queries = queryResponse['data']['queries'];
     let status = QueryStatus.code.NO_QUERY_DATA; // default status
-    if ( queries.length > 0) {
-      status =  queries[0]['status'];
+    if (!queryResponse || !queryResponse['data'] || !queryResponse['data']['queries']) {
+      Logger.getInstance().trace('Connection[id: %s] - No query data found, returning NO_QUERY_DATA status', id);
+      return status;
+    }
+    
+    const queries = queryResponse['data']['queries'];
+    if (queries.length > 0) {
+      status = queries[0]['status'];
     }
 
-    Logger.getInstance().trace('Connection[id: %s] - Extracted query status: %s', status);
+    Logger.getInstance().trace('Connection[id: %s] - Extracted query status: %s', id, status);
     return status;
   }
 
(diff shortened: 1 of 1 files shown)


lib/authentication/auth_workload_identity/types.ts
```diff
@@ -0,0 +1,7 @@
+export enum WorkloadIdentityProvider {
+  AWS = 'AWS',
```
sfc-gh-pmotacki, line 2: Is it only AWS part of WIF?

sfc-gh-pmotacki (approved):

lib/authentication/auth_workload_identity/types.ts
sfc-gh-rsavenok: Other 3 are commented as TODO in this file, they will be implemented in next PRs :)

lib/authentication/auth_workload_identity/auth_workload_identity.ts
```diff
@@ -0,0 +1,56 @@
...
+    } else {
+      // NOTE:
+      // "as WorkloadIdentityProvider" is temporary while no auto-detection is implemented
+      this.tokenProvider = provider as WorkloadIdentityProvider;
+      this.token = token;
+    }
+  }
+
+  async reauthenticate(_body: AuthRequestBody) {
+    throw new Error('TODO: Not implemented');
```
sfc-gh-dheyman, line 52: when is the plan to implement it?

lib/authentication/auth_workload_identity/attestation_aws.ts
```diff
@@ -0,0 +1,75 @@
...
+      return null;
+    }
+  }
+}
+
+/**
+ * Tries to create a workload identity attestation for AWS.
+ * If the application isn't running on AWS or no credentials were found, returns null.
+ */
+export async function getAwsAttestationToken() {
```
sfc-gh-dheyman, line 34: you can think whether it would be useful to add some logs to the flow.

lib/authentication/auth_workload_identity/auth_workload_identity.ts
sfc-gh-rsavenok: In the next few days

lib/authentication/auth_workload_identity/attestation_aws.ts
sfc-gh-rsavenok: The are logs when credentials or region aren't found, but the rest of the code will either crash on TypeScript compilation when wrong arguments are passed or will initialize and crush only when we'll hit auth endpoint.

PR #1097, opened 2025-06-11
SNOW-2113974 Enable single use of refresh token

sfc-gh-pmotacki:
### Description

### Checklist


lib/connection/connection_config.js
```diff
@@ -957,7 +958,16 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
...
-    return oauthHttpAllowed;
+    return oauthHttpAllowed || false;
+  };
+
+  /**
+   * Returns whether single use refresh tokens are enabled.
+   *
+   * @returns {Boolean}
+   */
+  this.getEnableSingleUseRefreshTokens = function () {
```
sfc-gh-rsavenok: General question: why do we create getters for simple options that could be done with:
```
this.enableSingleUseRefreshTokens = options.enableSingleUseRefreshTokens || false;
```

If in the future, the value would be more complex, it could be changed to a getter using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get under the same key (without get* prefix)

sfc-gh-pbulawa (approved):

PR #1102, opened 2025-06-16
 SNOW-2041980 Kerberos Proxy Auth - NodeJS

sfc-gh-ext-simba-jy:
### Description
- [x] Added the customizer Header connection configs
- [x] In request Util, create some functions to validate customizer headers and get headers
- [x] Added 'isRetry' options in snowflake connection requests and GCS and S3 utils (It can be in remote_util, but difficult to test)
- [x] Added testings.   

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -84,7 +84,7 @@ declare module 'snowflake-sdk' {
     useEnvProxy?: boolean;
   }
 
-  export interface ConnectionOptions extends WIP_ConnectionConfig {
+  export type ConnectionOptions = WIP_ConnectionConfig & {
     //Detail information: https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-options
 
     /**


package.json
```diff
@@ -79,16 +79,16 @@
     "lint:check:all": "eslint lib samples system_test test && check-dts index.d.ts",
     "lint:check:all:errorsOnly": "npm run lint:check:all -- --quiet",
     "lint:fix": "eslint --fix",
-    "test": "mocha 'test/unit/**/*.{js,ts}'",
-    "test:authentication": "mocha 'test/authentication/**/*.{js,ts}'",
-    "test:integration": "mocha 'test/integration/**/*.{js,ts}'",
+    "test": "mocha test/unit/**/*.{js,ts}",
```
sfc-gh-rsavenok: The quotes were added recently in https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1079

Mocha's behaviour is poorly documented: when a test pattern is passed without quotes, it uses the default shell pattern matching mechanism (different on every OS/shell), and if a string is passed, it uses glob from the npm package.

You can confirm this by running `npm run test:unit` on a command with quotes and without, and you'll see that it runs 300 fewer tests if quotes aren't set 🙈

lib/constants/error_messages.js
```diff
@@ -86,6 +86,8 @@ exports[404059] = 'Invalid oauth client id. The specified value must not be an e
 exports[404060] = 'Invalid oauth client secret. The specified value must not be an empty string';
 exports[404061] = 'Invalid oauth token request URL. The specified value must be a valid URL starting with the https or http protocol.';
 exports[404062] = 'No workload identity credentials were found. Provider: %s';
+exports[404063] = 'Invalid http header customizers. The specified value must contain the following functions: \'applices\', \'newHeaders\' and \'invokeOnce\'';
```
sfc-gh-rsavenok: Would suggest to adjust eslint config: `'quotes': ['error', 'single', { 'avoidEscape': true }],`

So it lets us use `""` instead of `''` when a string contains `'` :) 

It will let us make this line:
```
exports[404063] = "Invalid http header customizers. The specified value must contain the following functions: 'applices', 'newHeaders' and 'invokeOnce'";
```

lib/util.ts
```diff
@@ -678,3 +683,19 @@ export function escapeHTML(value: string) {
 export async function dynamicImportESMInTypescriptWithCommonJS(moduleName: string) {
   return Function(`return import("${moduleName}")`)()
 }
+
+export function isValidHTTPHeaderCustomizers(customizers: HttpHeadersCustomizer[]) : boolean {
```
sfc-gh-rsavenok: In one of the recent PRs, there was a discussion that util file is a mess and we should try to reduce it's size by moving helpers next to business logic that needs them.

Would suggest to move both this helper function and HttpHeadersCustomizer type http folder:
* add types.ts
* helper in request_util.js

package.json
sfc-gh-ext-simba-jy: Hmmm, interesting. On my machine(Windows), I cannot execute the npm command with the quote, and the 791 unit testing were executed without the quote. Anyway, I will have a look and let you know if I find any other solutions. I will revert this back when I need to update the PR based on the feedbacks. Thank you.

sfc-gh-rsavenok: [Mocha Docs](https://mochajs.org/#the-test-directory)

Maybe it should be double quotes instead of single to work correctly? 
> You should always quote your globs in npm scripts. If you use quotes, the [node-glob](https://www.npmjs.com/package/glob) module will handle its expansion. For maximum compatibility, surround the entire expression with double quotes and refrain from $, ", ^, and \ within your expression.

sfc-gh-fpawlowski:
Wouldn't it make sense to use this HttpInterceptor class for headers customziation?

https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/test/integration/test_utils/httpInterceptorUtils.js

package.json
sfc-gh-ext-simba-jy: Yes, it worked well when I replaced the single quote with the double quote. Currently, I revert it back. If you want me to fix this in this PR, I will update it. Please let me know.

lib/constants/error_messages.js
sfc-gh-ext-simba-jy: When I replaced the double quote with the single quote, the linter said:

```
Error:   89:19   error  Strings must use singlequote      quotes
Error:   89:[11](https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/15858048698/job/44708146624?pr=1102#step:5:12)1  error  Unnecessary escape character: \'  no-useless-escape
Error:   89:[12](https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/15858048698/job/44708146624?pr=1102#step:5:13)1  error  Unnecessary escape character: \'  no-useless-escape
Error:   89:125  error  Unnecessary escape character: \'  no-useless-escape
Error:   89:[13](https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/15858048698/job/44708146624?pr=1102#step:5:14)7  error  Unnecessary escape character: \'  no-useless-escape
Error:   89:[14](https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/15858048698/job/44708146624?pr=1102#step:5:15)4  error  Unnecessary escape character: \'  no-useless-escape
Error:   89:[15](https://github.com/snowflakedb/snowflake-connector-nodejs/actions/runs/15858048698/job/44708146624?pr=1102#step:5:16)6  error  Unnecessary escape character: \'  no-useless-escape
```

Currently, the linter enforces the use of single quotes only.

sfc-gh-ext-simba-jy:
I left the comment in the ticket @sfc-gh-fpawlowski. Please have a look and let me know.

FYI: I changed the testing file name (ts to js) because of the testing failure. I will take a look into the issue and change it to ts before I merge it.

PR #1107, opened 2025-06-25
SNOW-1528817 Deprecate region parameter for node.js and PHP driver

sfc-gh-ext-simba-jy:
### Description

### Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,3 +28,5 @@ jobs:
         run: npm run prettier:check
       - name: Check formatting
         run: npm run lint:check
+      - name: Check if TypeScript still compiles
+        run: npm run check-ts
--- a/ci/typescript-test.ts
+++ b/ci/typescript-test.ts
@@ -1,0 +1,7 @@
+/**
+ * NOTE:
+ *
+ * Our index.d.ts is manually maintained.
+ * This import ensures that the types are compiling with no errors.
+ */
+import * as snowflake from '../';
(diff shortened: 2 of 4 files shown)


lib/connection/connection_config.js
```diff
@@ -87,6 +87,7 @@ function consolidateHostAndAccount(options) {
 
 
   if (Util.exists(options.region)) {
+    Logger.default().info('Deprecated usage: Please append a region or subdomain to the end of the account parameter, e.g., account=<account>.<region>.snowflakecomputing.com"');
```
sfc-gh-dszmolka: 1. i think for `account`, we anyways not using the full `host` type notation, only without `.snowflakecomputing.com`. So following the example, it would be `account=<account>.<region>`
2. for a while, the 'regionless' notation (myorg-myaccount) is preferred. 

Adding all up, can we consider something like this?
```
The 'region' configuration item is deprecated. Please refer to https://docs.snowflake.com/en/user-guide/admin-account-identifier and use either the regionless (myorgname-myaccountname) or the locator (myaccountlocator.myregion) in `account` configuration item instead. 
```

sfc-gh-ext-simba-jy: Hello @sfc-gh-dszmolka, thank you for your opinion. I will update the log message accordingly.

sfc-gh-dszmolka: thank you for the change! (i'm not code owner here so cannot officially 'review' this but nevertheless thank you for implementing it!)

sfc-gh-pbulawa (approved):

PR #1108, opened 2025-06-26
SNOW-727166: [snowflake-connector-nodejs][GH#61](feature parity) adding support for QUERY_TAG and other session parameters

sfc-gh-ext-simba-jy:
### Description

### Checklist


lib/connection/connection_config.js
```diff
@@ -711,6 +712,16 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
     passcode = options.passcode;
   }
 
+  let queryTag = null;
+  if (Util.exists(options.queryTag)) {
```
sfc-gh-rsavenok: I deprecated `Util.exists` as we can just use `if (options.queryTag)` in almost every block that uses `Util.exists` atm.

lib/connection/connection_config.js
```diff
@@ -1035,6 +1046,10 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
     return passcode;
   };
 
+  this.getQueryTag = function () {
```
sfc-gh-rsavenok: I added this `NOTE` block a few weeks ago - 
https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/connection/connection_config.js#L1110-L1115
```
  // NOTE:
  // Don't add new get* methods when implementing new options. Use this.optionName instead.
  // If you need advanced logic in getter, use JS object getters.
  //
  // Future plan is to reduce complexity of this file by merging 3 objects into one context:
  // defaults+options+custom getters = ConnectionConfig
```

Please use `this.queryTag` instead. Could be without `let queryTag` declaration on line 722:
```
this.queryTag = options.queryTag; 
```
Don't think we need default of `null` as well.

This will greatly simplify future work of typing ConnectionConfig correctly

lib/connection/statement.js
```diff
@@ -438,6 +438,11 @@ function createContextPreExec(statementOptions, services, connectionConfig) {
   Errors.assertInternal(Util.isObject(services));
   Errors.assertInternal(Util.isObject(connectionConfig));
 
+  if (Util.exists(connectionConfig.getQueryTag())) {
```
sfc-gh-rsavenok: Based on prev comments, this would be `if (connectionConfig.queryTag)`

sfc-gh-rsavenok (approved):
Approving, but added a minor change request

lib/connection/statement.js
sfc-gh-ext-simba-jy: Done.

lib/connection/connection_config.js
sfc-gh-ext-simba-jy: Done.

lib/connection/connection_config.js
sfc-gh-ext-simba-jy: Done.

PR #1109, opened 2025-06-27
testing

sfc-gh-ext-simba-jy:
### Description

### Checklist

### DIFF
--- a/ci/container/test_component.sh
+++ b/ci/container/test_component.sh
@@ -26,6 +26,22 @@ else
 fi
 npm install
 
+if [[ "$LOCAL_USER_NAME" == "jenkins" ]]; then
+    echo "Jenkins detected, download all modules"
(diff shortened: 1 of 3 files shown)


lib/file_transfer_agent/azure_util.js
```diff
@@ -1,12 +1,18 @@
-const AZURE = require('@azure/storage-blob');
+const Logger = require('../logger');
+let AZURE;
+try {
+  AZURE = require('@azure/storage-blob');
```
sfc-gh-rsavenok, line 4: If we agree to proceed with optional clouds, such code will not work with TypeScript because require blocks aren't typed, and import() would be a promise that can't be used here.

We need to look into other ways of importing optional cloud dependencies that works well with typescript. 

Also, this PR needs to handle dependencies that were added with with WIF auth - https://github.com/snowflakedb/snowflake-connector-nodejs/tree/master/lib/authentication/auth_workload_identity

sfc-gh-ext-simba-jy: Hello @sfc-gh-rsavenok, I changed the code. Please have a look. I will have a look at auth_workload_identity.

sfc-gh-pszpakowski:
Consulted with @sfc-gh-rsavenok that we will not be proceeding with this approach, closing this PR.

PR #1120, opened 2025-07-10
 SNOW-1447051: Unhandled resources after creating a Connection prevents process from terminating when using externalbrowser auth.

sfc-gh-ext-simba-jy:
### Description

- [x] Resolved the timeout issue in the External broswser.

### Checklist

### DIFF
--- a/lib/configuration/connection_configuration.js
+++ b/lib/configuration/connection_configuration.js
@@ -1,9 +1,9 @@
 const toml = require('toml');
 const os = require('os');
 const fs = require('fs');
-const { validateOnlyUserReadWritePermissionAndOwner, generateChecksum } = require('../file_util');
+const { validateNoExtraPermissionsForOthersSync, generateChecksum } = require('../file_util');
 const path = require('path');
-const Logger = require('../logger');
+const Logger = require('../logger').default;
 const AuthenticationTypes = require('../authentication/authentication_types');
 const Util = require('../util');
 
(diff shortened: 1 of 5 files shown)


sfc-gh-pbulawa (approved):
LGTM

lib/authentication/auth_web.js
```diff
@@ -137,13 +137,17 @@ function AuthWeb(connectionConfig, httpClient, webbrowser) {
...
+    let timeoutId;
+    const timeout = new Promise(
+      (resolve, reject) =>
+        (timeoutId = setTimeout(
+          () => reject(`Browser action timed out after ${browserActionTimeout} ms.`),
+          millis,
+        )),
     );
-    return Promise.race([promise, timeout]);
+    return Promise.race([promise, timeout]).finally(() => {
```
sfc-gh-rsavenok, line 148: Have you tried removing `exit: true` from `.mocharc.js` after this change?

I think we need to remove that option from mocha to catch such issues early

sfc-gh-ext-simba-jy: It seems like another hang issue exists in a different part of the code. We still need to investigate it under a separate ticket. https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1124. I checked that the hang issue in the external browser authentication flow has been resolved in my manual test.

sfc-gh-rsavenok (approved):

PR #1133, opened 2025-07-25
SNOW-2172800 [NodeJS] Wrong error when using GET command

sfc-gh-ext-simba-jy:
### Description

### Checklist


lib/errors.js
```diff
@@ -372,7 +372,7 @@ function createError(name, options) {
   error.code = code = options.code;
 
   // if no error message was specified in the options
-  let message = options.message;
+  let message = options.cause || options.message;
```
sfc-gh-rsavenok: Since cause is an Error object, I think this should be `let message = options.message ?? options.cause?.message`

Otherwise you'll get `Error: Error: cause error` instead of `Error: cause error`:
```
Error [NetworkError]: Error: ENOENT: no such file or directory, stat 'tmp2'
vs
Error [NetworkError]: ENOENT: no such file or directory, stat 'tmp2'
```

sfc-gh-rsavenok (approved):

lib/errors.js
sfc-gh-rsavenok: NOTE that `options.message` if first, otherwise we won't be able to create a network error with a message :P

sfc-gh-ext-simba-jy: I edited it. Thank you!

PR #1142, opened 2025-08-05
 SNOW-925245: [snowflake-connector-nodejs] please consider enhancing chunk download logging for supportability

sfc-gh-ext-simba-jy:
### Description

### Checklist


lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -292,11 +294,18 @@ function RemoteStorageUtil(connectionConfig) {
     let lastErr;
     const maxRetry = DEFAULT_MAX_RETRY;
 
+    Logger().info(
+      `Started downloading result. File: ${meta.dstFileName}, location: ${meta.stageInfo.location}`,
+    );
+    const timer = new ExecutionTimer().start();
     for (let retry = 0; retry < maxRetry; retry++) {
       // Download the file
       await utilClass.nativeDownloadFile(meta, fullDstPath, maxConcurrency);
```
PoornakumarRasiraju, line 303: To ensure precise performance benchmarking and deterministic behavior, consider explicitly calling `timer.stop()` immediately after the download completes. This small addition makes the timing more accurate.

lib/connection/result/result.js
```diff
@@ -443,6 +446,10 @@ Result.prototype.fetchRows = function (options) {
 
     // start processing rows
     processRows();
+    timer.stop();
+    Logger.debug(
+      `Finished processing Chunks: Processed ${rowsLength} rows in ${timer.getDuration()} ms`,
```
sfc-gh-dszmolka, line 451: nit / Q : is this gonna return the duration in 12.29382938382938 ms format ? (we already have some timers elsewhere in the code, and it looks weird) . 

If yes, can we consider leaving the fractions and round up/down to next whole number?

lib/file_transfer_agent/azure_util.js
```diff
@@ -246,6 +246,9 @@ function AzureUtil(connectionConfig) {
     try {
       const downloadBlockBlobResponse = await blockBlobClient.download(0);
       const readableStream = downloadBlockBlobResponse.readableStreamBody;
+      Logger().debug(
+        `Sent Get Request to ${blockBlobClient.url.split('?')[0]}, destination: ${fullDstPath}, http status: ${downloadBlockBlobResponse.originalResponse._response.status}`,
```
sfc-gh-dszmolka, line 250: what if there's timeout / no response / no status field in the response?

lib/file_transfer_agent/gcs_util.js
```diff
@@ -360,24 +361,27 @@ function GCSUtil(connectionConfig, httpClient) {
 
     try {
       if (this.shouldUseJsonApi(meta)) {
-        const gcsLocation = this.extractBucketNameAndPath(meta['stageInfo']['location']);
+        const { bucketName, path } = this.extractBucketNameAndPath(meta['stageInfo']['location']);
+        Logger().info(`Use GCP client to download the file.`);
```
sfc-gh-dszmolka: do we have the same in Azure and AWS clients too ?

lib/file_transfer_agent/gcs_util.js
```diff
@@ -360,24 +361,27 @@ function GCSUtil(connectionConfig, httpClient) {
...
+          .bucket(bucketName)
+          .file(path + meta['srcFileName'])
           .getMetadata();
 
         encryptionDataprop = metadata[0].metadata[ENCRYPTIONDATAPROP];
         matDescKey = metadata[0].metadata[MATDESC_KEY];
         sfcDigest = metadata[0].metadata[SFC_DIGEST];
         size = metadata[0].size;
       } else {
+        Logger().info(`Downloading file from GCS using Axios`);
```
sfc-gh-dszmolka: nit: can we consider setting this to debug level ?

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -292,11 +294,19 @@ function RemoteStorageUtil(connectionConfig) {
     let lastErr;
     const maxRetry = DEFAULT_MAX_RETRY;
 
+    Logger().info(
+      `Started downloading result. File: ${meta.dstFileName}, location: ${meta.stageInfo.location}`,
```
sfc-gh-dszmolka, line 298: i would again consider having this on debug level, sounds more like a troubleshooting data than something people wanna look at by default

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -292,11 +294,19 @@ function RemoteStorageUtil(connectionConfig) {
...
+    );
+    const timer = new ExecutionTimer().start();
     for (let retry = 0; retry < maxRetry; retry++) {
       // Download the file
       await utilClass.nativeDownloadFile(meta, fullDstPath, maxConcurrency);
 
       if (meta['resultStatus'] === resultStatus.DOWNLOADED) {
+        timer.stop();
+        Logger().info(
+          `Successfully downloaded result. File: ${meta.dstFileName}, download time: ${timer.getDuration()} ms, chunkSize: ${meta.srcFileSize} bytes.`,
```
sfc-gh-dszmolka, line 308: same question with getduration

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -312,13 +322,20 @@ function RemoteStorageUtil(connectionConfig) {
...
           const tmpDstName = await SnowflakeEncryptionUtil.decryptFile(
             fileHeader.encryptionMetadata,
             meta['encryptionMaterial'],
             fullDstPath,
             meta['tmpDir'],
           );
-
+          timer.stop();
+          Logger().info(
+            `Successfully decrypted the downloaded file for ${meta.dstFileName} in ${timer.getDuration()} ms.`,
```
sfc-gh-dszmolka, line 337: same question with getduration

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -352,16 +369,26 @@ function RemoteStorageUtil(connectionConfig) {
...
+            Logger().info(
+              `File size verification: Downloaded file at "${meta.fullDstPath}" is ${stat.size} bytes.`,
+            );
             resolve();
           });
         });
 
         return;
       } else if (meta['resultStatus'] === resultStatus.RENEW_TOKEN) {
+        Logger().warn(`Need to renew token for downloading file: ${meta.dstFileName}`);
```
sfc-gh-dszmolka: i think we should not log this at WARN level ; would cause unnecessary worry. Renewing token is part of the normal operation. Can we have this at debug?

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -352,16 +369,26 @@ function RemoteStorageUtil(connectionConfig) {
...
         return;
       } else if (meta['resultStatus'] === resultStatus.RENEW_PRESIGNED_URL) {
+        Logger().warn(
+          `Need to renew presigned URL for downloading. file: ${meta.dstFileName}, presignedUrl: ${meta.presignedUrl}`,
```
sfc-gh-dszmolka, line 385: similarly with the warn, is it necessary to bubble up , always, even if its successful?

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -352,16 +369,26 @@ function RemoteStorageUtil(connectionConfig) {
...
+        );
         return;
       } else if (meta['resultStatus'] === resultStatus.NEED_RETRY_WITH_LOWER_CONCURRENCY) {
+        Logger().warn(
+          `Need to retry with lower concurrency for downloading file: ${meta.dstFileName}`,
```
sfc-gh-dszmolka, line 390: please consider having this at debug level, it's troubleshooting data

sfc-gh-dszmolka (commented):
generally looks good

lib/file_transfer_agent/gcs_util.js
sfc-gh-ext-simba-jy: No, they don’t. GCP is a special case because it supports both the GCP client and Axios, allowing the user to choose between them. That’s why this kind of logging exists only in GCP.

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: Changed to the debug level

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: Changed to the debug level

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: Changed to the debug level

lib/file_transfer_agent/gcs_util.js
sfc-gh-ext-simba-jy: Changed to debug

lib/file_transfer_agent/azure_util.js
sfc-gh-ext-simba-jy: In that case, the error should occur in blockBlobClient.download(0). If it does, the catch block updates the result status, and remote_storage_util handles the error. I’ve already added the error message.

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: I changed it to the debug

lib/connection/result/result.js
sfc-gh-ext-simba-jy: I made the timer round the value up, leaving only two decimal places. https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1142/files#diff-4d881b637410df0c259f6928b749ad82694b51a22387e79a726120e461433827R39

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: Updated the execution_timer

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: Same as above

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -352,16 +369,26 @@ function RemoteStorageUtil(connectionConfig) {
         await new Promise((resolve) => {
           fs.stat(fullDstPath, (err, stat) => {
             meta['dstFileSize'] = stat.size;
+            Logger().info(
```
sfc-gh-rsavenok: I think this should be `debug`, we use `info` to notify developers about configuration issues that they can fix.

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -373,6 +400,7 @@ function RemoteStorageUtil(connectionConfig) {
           await new Promise((resolve) => setTimeout(resolve, sleepingTime));
         }
       } else if (meta['resultStatus'] === resultStatus.NEED_RETRY) {
+        Logger().warn(`Need to retry for downloading file: ${meta.dstFileName}`);
```
sfc-gh-rsavenok: I think this should be debug as every warn is displayed by default and customers might as developer can't do anything about them and might ask us to silence such messages

lib/file_transfer_agent/remote_storage_util.js
```diff
@@ -382,9 +410,10 @@ function RemoteStorageUtil(connectionConfig) {
       }
     }
     if (lastErr) {
+      Logger().error(`Failed to downloading file: ${meta.dstFileName} with error: ${lastErr}`);
```
sfc-gh-rsavenok, line 413: What's inside the `lastError` object? (Wondering if we really need an additional log message here)

lib/file_transfer_agent/s3_util.js
```diff
@@ -232,7 +233,12 @@ function S3Util(connectionConfig, s3, filestream) {
     try {
       await client
         .getObject(params)
-        .then((data) => data.Body.transformToByteArray())
+        .then((data) => {
+          Logger().debug(
```
sfc-gh-rsavenok, line 240: On error, we won't see this debug message. Is this the desired behavior?

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: The lastErr object stores the error value when either resultStatus.NEED_RETRY or resultStatus.NEED_RETRY_WITH_LOWER_CONCURRENCY occurs. If the retry attempts exceed the maximum retry count, the last error will be this stored error value. In my opinion, this message helps users understand why the execution failed. If you think this log is not necessary, please let me know.

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: Replaced it to debug

lib/file_transfer_agent/remote_storage_util.js
sfc-gh-ext-simba-jy: Replaced it to debug

lib/file_transfer_agent/s3_util.js
sfc-gh-ext-simba-jy: I added a new log above. If the operation succeeds, this log will display the HTTP status; if it fails, the error will be logged in the remote util.

sfc-gh-rsavenok (approved):

PR #1143, opened 2025-08-06
SNOW-2251295 Remove WIF auto-detect

sfc-gh-rsavenok:
### Description
- Remove WIF auto-detect feature
- Adjust azure timeout
- added a todo to simplify error types
- better error codes for WIF

### Checklist

- [X] Create tests which fail without the change (if possible)
- [X] Make all tests (unit and integration) pass (`npm run test:unit` and `npm run test:integration`)
- [X] Extend the types in index.d.ts file (if necessary)
- [X] Extend the README / documentation and ensure is properly displayed (if necessary)
- [X] Provide JIRA issue id (if possible) or GitHub issue id in commit message


lib/errors.js
```diff
@@ -59,6 +59,9 @@ exports.assertInternal = function (expression, message) {
   }
 };
 
+// TODO:
```
sfc-gh-pfus, line 62: Should we keep this todo?

sfc-gh-rsavenok: I prefer TODOs in code instead of jira tickets as there's high chance it will never be done :P 

But at least next person reading errors.js would consider using InvalidParameterError instead of MissingParameterError :)

lib/constants/error_messages.js
```diff
@@ -110,6 +110,7 @@ exports[405504] =
 exports[405505] = 'Configuration from client config file failed';
 exports[405506] = 'Wrong authorization type';
 exports[405507] = 'Authenticator not allowed';
+exports[405508] = 'Failed to fetch workload identity credentials from cloud provider';
```
sfc-gh-mhofman, line 113: if this makes sense - could we still provide which Provider was that? when customer has an invalid configuration we at least would know from this message?

sfc-gh-mhofman (approved):

lib/constants/error_messages.js
sfc-gh-rsavenok: When an invalid configuration is passed, `404062` will be returned. This error code is only for cases where config is valid, but driver can't fetch credentials on VM or serverless platform

PR #1144, opened 2025-08-06
Improve WIF attestation logs for AWS

sfc-gh-rsavenok:
### Description
Improve logging

### Checklist


lib/authentication/auth_workload_identity/attestation_aws.ts
```diff
@@ -10,7 +10,7 @@ export async function getAwsCredentials() {
     Logger().debug('Getting AWS credentials from default provider');
     return await defaultProvider()();
   } catch (error) {
-    Logger().debug('No AWS credentials were found.');
+    Logger().debug(`Failed to get AWS credentials. ${error}`);
```
PoornakumarRasiraju: If error is an object, using a template literal like `${error}` might result in [object Object] in the logs, which hides the actual error details. 
Could we consider updating it to:
`Logger().debug('Error getting Azure auth token:', error);`

or 

`Logger().debug('Error getting Azure auth token:', { error });`

This will ensure the full error and stack trace.

sfc-gh-rsavenok: Our logger doesn't work like JS `console.log`, it won't print objects (unless you specifically expect an object to be always there). N+1 arguments are for template strings. e.g. `.debug('a value is %s', value)`

Even though JS lets you throw anything, I don't expect cloud SDKs throwing objects instead of exceptions (they are convertable to strings)
```
try {
  throw { a: 2 }
} catch (e) {
    console.log(`---${e}`);
}
```

sfc-gh-rsavenok: @PoornakumarRasiraju thinking a bit more.... I'll remove generic `ERR_CONN_CONNECT_FAILED_TO_FETCH_WORKLOAD_IDENTITY_CREDENTIALS` error and will forward original cloud sdk error all the way up to .connect() result so the user could see the trace in their error callback.

Will ping you once PR is updated

sfc-gh-rsavenok: @PoornakumarRasiraju updated ;)

PoornakumarRasiraju: Looks good to me!

PR #1151, opened 2025-08-21
SNOW-2117146 Foundation for CRL Validation

sfc-gh-rsavenok:
### Description

Implements foundation for CRL validation:
* new connection config options
* connection config is drilled down to the http agent getter
* new error type `createCrlError` + retry adjustments
* Initialize `HttpsCrlAgent` if config options are present
* adjust `HttpsProxyAgent` to call crl validation if config options are present
* testing foundation
* initial CRL validation that checks for `crlDistributionPoints` and short-lived certificates

Additional:
Fixed `check-ts`, which was running typescript validation on the root folder without tsconfig instead of the dist folder that we pack as an npm dependency. This was causing unnecessary errors when the generator was used.

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -740,6 +740,7 @@ declare module 'snowflake-sdk' {
     start?: number;
     end?: number;
     fetchAsString?: DataType[];
+    each?: (row: any) => boolean | void;
   }
 
   export const ErrorCode: typeof ErrorCodeEnum;


lib/agent/crl_validator.ts
```diff
@@ -0,0 +1,65 @@
...
+    } catch (error: unknown) {
+      socket.destroy(error as Error);
+    }
+  });
+  socket.cork();
+}
+
+export function validateCrl(certChain: DetailedPeerCertificate, config: CRLValidatorConfig) {
+  for (const certificate of iterateCertChain(certChain)) {
+    Logger().debug('validateCrl: checking certificate %j', certificate.subject);
```
sfc-gh-rsavenok, line 38: Logs will be hard to understand if multiple connections are executed at the same time, change to tag each log message with certificate. 

Same is needed in crl_utils

lib/connection/connection_config.js
```diff
@@ -1139,6 +1144,13 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
   this.oauthEnableSingleUseRefreshTokens = options.oauthEnableSingleUseRefreshTokens;
   this.workloadIdentityProvider = options.workloadIdentityProvider;
   this.workloadIdentityAzureEntraIdResource = options.workloadIdentityAzureEntraIdResource;
+  this.crlValidatorConfig = {
+    checkMode: options.certRevocationCheckMode ?? 'DISABLED',
+    allowCertificatesWithoutCrlURL: options.crlAllowCertificatesWithoutCrlURL ?? true,
```
sfc-gh-rkowalski: Isn't it better to have this disabled by default?

sfc-gh-rkowalski (approved):

lib/agent/crl_validator.ts
sfc-gh-rsavenok: I'll adjust it in separate PR

PR #1157, opened 2025-08-29
SNOW-2236247 Add secret_scanning.yml

sfc-gh-fochnik:
Add secret_scanning.yml and CODEOWNERS entries for GitHub Advanced Security Secret Scanning configuration

### DIFF
--- a/.github/secret_scanning.yml
+++ b/.github/secret_scanning.yml
@@ -1,0 +1,2 @@
+paths-ignore:
+  - "**/test/data/**"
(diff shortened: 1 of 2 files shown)


.github/secret_scanning.yml
```diff
@@ -0,0 +1,2 @@
+paths-ignore:
+  - "**/test/data/**"
```
sfc-gh-rsavenok, line 2: @sfc-gh-fochnik can we remove this line? We don't have a `test/data` folder and I think any test-related file should be scanned for secrets because our tests connect to real accounts.

sfc-gh-fochnik: I guess historically there was a `test/data` folder.
If you don't want it excluded that's entirely up to you - we can close this PR.

PR #1162, opened 2025-09-04
SNOW-2117146 CRL Validation

sfc-gh-rsavenok:
### Description
- Implements full validation without caching
- Improved debug logs and errors

Integration tests are still disabled due to CI config issues, but locally they pass for preprod

### Checklist


test/unit/agent/test_utils.ts
```diff
@@ -1,122 +1,182 @@
...
+  if (!algorithm) {
+    throw new Error(`Unsupported algorithm OID: ${algorithmOid}`);
+  }
+
+  // RSA
+  if (algorithmOid.startsWith('1.2.840.113549.1.1.')) {
+    const pair = crypto.generateKeyPairSync('rsa', {
+      modulusLength: 512,
+      publicExponent: 0x10001,
+    });
```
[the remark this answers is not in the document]
sfc-gh-rsavenok: This is unneeded in tests, make each test run 200ms

lib/agent/crl_utils.ts
```diff
@@ -1,99 +1,133 @@
...
-    parameters?: any;
-  };
-  signature: Buffer;
+export const SUPPORTED_CRL_VERIFICATION_ALGORITHMS: Record<string, string> = {
+  '1.2.840.113549.1.1.11': 'sha256WithRSAEncryption',
+  '1.2.840.113549.1.1.12': 'sha384WithRSAEncryption',
+  '1.2.840.113549.1.1.13': 'sha512WithRSAEncryption',
+  '1.2.840.10045.4.3.2': 'SHA256',
+  '1.2.840.10045.4.3.3': 'SHA384',
+  '1.2.840.10045.4.3.4': 'SHA512',
```
sfc-gh-jkasten: These are not signature verification algorithms. They are just hash algorithms. I don't believe they belong here?

lib/agent/crl_utils.ts
```diff
@@ -1,99 +1,133 @@
...
-      crlEntryExtensions?: any;
-    }[];
-    crlExtensions?: any;
-  };
-  signatureAlgorithm: {
-    algorithm: string;
-    parameters?: any;
-  };
-  signature: Buffer;
+export const SUPPORTED_CRL_VERIFICATION_ALGORITHMS: Record<string, string> = {
```
sfc-gh-jkasten: You covered the applicable PKCS1 v1.5 algorithms, but we should support all of the available algorithms at
https://github.com/cabforum/servercert/blob/main/docs/BR.md#7132-signature-algorithmidentifier

(If you need to get the OIDs out of the encoded values listed, you can use this website - https://lapo.it/asn1js/#MA0GCSqGSIb3DQEBCwUA)

ECDSA signature algorithms are mandatory - we will run into those in the wild. 

It would be nice to have PSS as they are allowed and we don't control the CAs. If it is very problematic we might be able to find an alternative. (PSS has better properties than PKCS1 v1.5, but I honestly think we will move to PQ algorithms before PSS is used in the Web PKI)

No need to support the SHA1 algorithm listed. -- all of the conditions in the requirements won't apply to anything we see.

lib/agent/crl_utils.ts
```diff
@@ -1,99 +1,133 @@
...
+  '1.2.840.113549.1.1.11': 'sha256WithRSAEncryption',
+  '1.2.840.113549.1.1.12': 'sha384WithRSAEncryption',
+  '1.2.840.113549.1.1.13': 'sha512WithRSAEncryption',
+  '1.2.840.10045.4.3.2': 'SHA256',
+  '1.2.840.10045.4.3.3': 'SHA384',
+  '1.2.840.10045.4.3.4': 'SHA512',
+};
+
+export function getCertificateName(certificate: DetailedPeerCertificate) {
```
sfc-gh-jkasten: note: I think this works for the "debug log" use case, I thought I should mention that this isn't precise.

There may be some certificates with the same Organization (O) and commonName (CN), but a different serialNumber. Many certificates won't have an organization or a common name. It is possible that two certificates contain the same serial number (though very improbable given the 64-bit entropy requirements on the serial number)

optional: Maybe add a comment about its use case?

lib/agent/crl_utils.ts
```diff
@@ -1,99 +1,133 @@
...
-        }
-        if (fullNameEntry.value.startsWith('http')) {
-          // Even though the spec allows multiple http urls, we only pick first one and don't handle redundancy
-          result.push(fullNameEntry.value);
-          break;
-        } else {
-          Logger().debug('getCertificateCrlUrlss: skipping non-http value %j', fullNameEntry);
-        }
+      if (fullNameEntry.value.startsWith('http')) {
+        // Even though the spec allows multiple http urls, we only pick first one and don't handle redundancy
```
sfc-gh-jkasten, line 54: nit on the comment / FYI: It is slightly more complicated than not handling redundancy. We are making an assumption that the CA also isn't employing the `onlySomeReasons` field in the CRLs to segment them. I think this is OK for the CAs we are dealing with.

`onlySomeReasons` is [not recommended by the CABForum](https://github.com/cabforum/servercert/blob/main/docs/BR.md#:~:text=The%20onlySomeReasons%20field%20SHOULD%20NOT%20be%20included%3B%20if%20included%2C%20then%20the%20CA%20MUST%20provide%20another%20CRL%20whose%20scope%20encompasses%20all%20revocations%20regardless%20of%20reason%20code.), but is not restricted. There is an additional `reasons` field in the certificate's CRL distribution point which is explicitly disallowed, which does roughly the same thing, but that field is optional from an RFC standpoint. `onlySomeReasons` works without the CRLs `reasons` if I remember Section 6.3 of RFC 5280 correctly (CRL validation routine). I don't believe modern CAs are using it, but I haven't performed an analysis recently. (10 years ago it wasn't uncommon)

I think we decided to assume a full CRL at the first distribution point in other codebases, but I will check again.

lib/agent/crl_validator.ts
```diff
@@ -21,45 +30,81 @@ export function isCrlValidationEnabled(config: CRLValidatorConfig) {
...
+    yield current;
+    current = current.issuerCertificate;
+  }
+}
+
+// NOTE:
+// Sticking with asn1.js-rfc5280 + custom signature validation, because popular libraries have issues:
+// - jsrsasign: has outdated crypto library with CEV issues
+// - pkijs: takes 4 seconds to parse 9Mb CRL
+// - @peculiar/x509: takes 8 seconds to parse 9Mb CRL
```
sfc-gh-jkasten: This is extremely unfortunate. What type of times are you getting for this custom handler?

I know the CEO of Peculiar Ventures (peculiar/x509) and will surface the issue.

lib/agent/crl_utils.ts
```diff
@@ -1,99 +1,133 @@
...
+ */
+export function isShortLivedCertificate(decodedCertificate: ASN1.CertificateDecoded) {
+  const notBefore = new Date(decodedCertificate.tbsCertificate.validity.notBefore.value);
+  const notAfter = new Date(decodedCertificate.tbsCertificate.validity.notAfter.value);
 
   // Certificates issued before March 15, 2024 are not considered short-lived
-  if (validFrom < new Date('2024-03-15T00:00:00.000Z')) {
+  if (notBefore < new Date('2024-03-15T00:00:00.000Z')) {
     return false;
   }
```
sfc-gh-jkasten: Certificates have a max validity period of 398 days. You will never encounter a valid certificate that matches this definition. I think you can remove it.

https://github.com/cabforum/servercert/blob/main/docs/BR.md#632-certificate-operational-periods-and-key-pair-usage-periods

lib/agent/crl_validator.ts
```diff
@@ -21,45 +30,81 @@ export function isCrlValidationEnabled(config: CRLValidatorConfig) {
...
+      if (!isCrlSignatureValid(crl, crlIssuerPublicKey)) {
+        throw new Error(
+          `CRL ${crlUrl} signature is invalid. Expected signature by ${getCertificateName(certificate.issuerCertificate)}`,
+        );
+      }
+
+      logDebug(`checking if certificate is revoked in ${crlUrl}`);
+      if (isCertificateRevoked(decodedCertificate, crl)) {
+        throw new Error(`Certificate ${name} is revoked in ${crlUrl}`);
+      }
```
sfc-gh-jkasten: I think this code needs to prevent substitution attacks.

1. You need to make sure that CRL's subject matches the subject of the issuing CA.
2. You need to make sure that the IssuingDistributionPoint URL matches the URL you queried.
3. You need to make sure that the CRL is still valid. (before nextUpdate and ideally check that now > thisUpdate -- barring some time skew if you really want to be robust)

The full RFC algorithm is listed at https://datatracker.ietf.org/doc/html/rfc5280#section-6.3. We don't need to support things like deltaCRLs, but it should give some idea.

lib/agent/crl_validator.ts
sfc-gh-rsavenok: Custom handler is ~500ms for 9Mb CRL. pkijs is also from `@peculiar`, is's lower level api, but still solid to implement CRL validation.

Main problem is that this 4 or 8 seconds blocks entire IO of nodejs process so nothing else can be done during parsing

lib/agent/crl_validator.ts
sfc-gh-rsavenok: Few Questions:
- Isn't the signature inndicates that subject matches? 
- `IssuingDistributionPoint` isn't present in DecodedCertificateList that ASN1.js-rfc5080 provides, can we skip it?
- even if CRL is outdated, if a certificate is added to CRL is it there forever? So by checking nextUpdate we'll only skip unnecessary checks? (If the certificate has multiple distribution points, it will check all of them)

lib/agent/crl_utils.ts
sfc-gh-rsavenok: Aren't ECDSA algorithms oids 1.2.840.10045.4.3.*? 

Since we don't have open-source library that implements different kinds of algorythms, we agreed to implement same list of oids that were used for CRL and extend it if customers will complain. 

The integration test that talks to preprod in every cloud region will indicate that we cover the basics.

lib/agent/crl_utils.ts
sfc-gh-rsavenok: For OCSP verification, we had this map for 7 years with no complaints:
```
exports.sign = {
  '1.2.840.113549.1.1.5': 'sha1WithRSAEncryption',
  '1.2.840.113549.1.1.4': 'md5WithRSAEncryption',
  '1.2.840.113549.1.1.2': 'md2WithRSAEncryption', // SignatureALG HashMD2 PubKeyALG_RSA
  '1.2.840.113549.1.1.11': 'sha256WithRSAEncryption',
  '1.2.840.113549.1.1.12': 'sha384WithRSAEncryption',
  '1.2.840.113549.1.1.13': 'sha512WithRSAEncryption',
  '1.2.840.113549.1.1.14': 'sha224WithRSAEncryption',
  '1.2.840.10040.4.3': 'dsaWithSHA1', // 'SignatureALG HashSHA1 PubKeyALG_DSA'
  '1.2.840.10045.4.1': 'ecdsa-with-SHA1', // 'SignatureALG HashSHA1 PubKeyALG_EC'
  '1.2.840.10045.4.3.1': 'SHA224', // 'SignatureALG HashSHA224 PubKeyALG_EC'
  '1.2.840.10045.4.3.2': 'SHA256', // 'SignatureALG HashSHA256 PubKeyALG_EC'
  '1.2.840.10045.4.3.3': 'SHA384', // 'SignatureALG HashSHA384 PubKeyALG_EC'
  '1.2.840.10045.4.3.4': 'SHA512', // 'SignatureALG HashSHA512 PubKeyALG_EC'
  '2.16.840.1.101.3.4.2.1': 'SHA256', // 'SignatureALG HashSHA256 PubKeyALG_RSAPSS'
  '2.16.840.1.101.3.4.2.2': 'SHA384', // 'SignatureALG HashSHA384 PubKeyALG_RSAPSS'
  '2.16.840.1.101.3.4.2.3': 'SHA512', // 'SignatureALG HashSHA512 PubKeyALG_RSAPSS'
  '2.16.840.1.101.3.4.2.4': 'SHA224', // 'SignatureALG HashSHA224 PubKeyALG_RSAPSS'
  '2.16.840.1.101.3.4.3.1': 'SHA224', // 'SignatureALG HashSHA224 PubKeyALG_DSA'
  '2.16.840.1.101.3.4.3.2': 'SHA256' // 'SignatureALG HashSHA256 PubKeyALG_DSA'
}
```
`2.16.840.1.101.3.4.2.*` are hash oids, so skipped them. And skipped sha1 + md*

lib/agent/crl_utils.ts
sfc-gh-rsavenok: How would you write this comment in code?

lib/agent/crl_utils.ts
sfc-gh-rsavenok: There's a test that signs crl using each OID in the map using` createCertificateKeyPair` (test_utils and crl_utils_test). Could you please review if that generation is correct?

lib/agent/crl_utils.ts
sfc-gh-snoonan: The text is
> The onlySomeReasons field SHOULD NOT be included; if included, then the CA MUST provide another CRL whose scope encompasses all revocations regardless of reason code.

In code, I think you can do something like this to get all the CRL endpoints we care about?
```
function noReasonFlag(distributionPoint) {
  return distributionPoint.ReasonFlags & 0
  // you may need to check for all bits set as well because that is technically allowed by the wording...
}

function isValidEndpoint(fullNameEntry) {
 return fullNameEntry.type === 'uniformResourceIdentifier' && fullNameEntry.value.startsWith('http')
}

function extractEndpoints(distributionPoint) {
 return entry.distributionPoint.value.fullNameEntry.filter(isValidEndpoint)
}

distributionPoints.filter(noReasonFlag).map(extractEndpoints).flat()
```

If you like, just take the first one there I guess.

The reasonFlags are
```
ReasonFlags ::= BIT STRING {
     unused                  (0),
     keyCompromise           (1),
     cACompromise            (2),
     affiliationChanged      (3),
     superseded              (4),
     cessationOfOperation    (5),
     certificateHold         (6),
     privilegeWithdrawn      (7),
     aACompromise            (8) }
```

lib/agent/crl_utils.ts
sfc-gh-jkasten: OK, I see. The string identifier is just mislabled for the ECDSA algorithms.

e.g. 1.2.840.10045.4.3.2 - is actually "ecdsa-with-SHA256" (See https://oid-base.com/get/1.2.840.10045.4.3.2)

"SHA256" is a hash algorithm (as opposed to a signature algorithm). It has a different OID - https://oid-base.com/get/2.16.840.1.101.3.4.2.1

> Could you please review if that generation is correct?

Thanks! Yes, the generation looks correct for the algorithms that you support (minus the names. s/SHA256/ecdsa-with-SHA256/)

lib/agent/crl_utils.ts
sfc-gh-jkasten: Yes, the ECDSA algorithms are there, they are just mislabeled, Thanks! Responded more fully in the other [thread](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1162/files#r2324518144).

---

The problem is that building code for what CAs currently do is brittle [[writeup](https://docs.google.com/document/d/1KWDAF_QJnd3Yy1GEaVKUmJ_NoeyiyAALyBARZX3bU8s/edit?tab=t.0#heading=h.f2v9xe44m965)]. We need to implement to the specifications and underlying requirements. The CA is free to change the algorithm based upon the current requirements and once they do, the clients will break until they turn off revocation or update their driver after a patch. 

The PSS algorithms are covered in the OCSP code you [linked](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1162/files#r2324499377) in the other thread, but I can't find that code in this repo. I don't think it should be too hard to add support. I can also take a look after this PR. Can you mark it as a TODO?

lib/agent/crl_validator.ts
sfc-gh-jkasten: Thanks for the info! Peculiar Ventures mentioned they would spend a few days looking into it.

lib/agent/crl_utils.ts
sfc-gh-jkasten: Sorry for writing incoherently. I was trying to get a response out the door. I should have spent more time with both my writing and the slew of requirements.

tl;dr; - I think that we can simply note that "Since CRL Distribution Points reasons are not allowed by the Baseline Requirements, the associated CRLs MUST have all reason codes by RFC 5280." [[BR text](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712112-crl-distribution-points), [RFC 5280 text](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13:~:text=If%20the%20DistributionPoint%20omits%20the%20reasons%20field%2C%20the%20CRL%20MUST%0A%20%20%20include%20revocation%20information%20for%20all%20reasons.)] Every CRL distribution point found in a certificate will have all reasons.

For reference material only - there are actually two variables which dictate essentially the same information. One in the certificate and the other in the CRL.

1. `reasons` - Inside the **certificate's** CRL Distribution Point Extension. The Baseline Requirements have outlawed the inclusion of this field. See [Section 4.2.1.13 of RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13) and the BR text above.
2. `onlySomeReasons` - Inside the **CRL's** Issuing Distribution Point Extension. See [Section 5.2.5 of RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.5)

I believe the thought is that clients who download the certificate can choose which reason codes they cared about before they are fetched and CRLs that are cached can also be efficiently reasoned about.

The text that threw me off yesterday was that I had reread the CRL validation routine and it accounts for all scenarios where one of the two fields may be missing [Section 6.3.3.d in RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-6.3.3:~:text=Compute%20the%20interim_reasons_mask%20for%20this%20CRL%20as%20follows%3A). However, this should not be a real possibility. Section 4.2.1.13 states that if `reasons` is omitted, then corresponding CRL MUST contain all reason codes. 

The CRL validation algorithm accounts for violations of the RFC in an effort to be robust, but that isn't important for the modern Web PKI. The web PKI requires extremely strict enforcement to the RFCs, so it is fine to simply note it and move on.

The Baseline Requirements do not prohibit CAs issuing CRLs with `onlySomeReasons`, but it essentially can't be a CRL that is referenced inside a certificate due to the other requirements.

lib/agent/crl_validator.ts
sfc-gh-jkasten: > Isn't the signature inndicates that subject matches?

In almost all cases yes, but not necessarily. Certainly historically, different CA certificates (owned by the same company / org or sold) shared public keys. RFC 5280 certificate chaining relies upon the tuple of the Subject/Issuer + the subjectKeyID/authorityKeyID (SKID/AKID)

Now I think any modern CA worth their salt shouldn't be reusing key material, and there are actually new rules which require that new CAs use keys generated within X time, but I have not performed a recent audit. I know way back in the day it did occur though [EFF's SSL Observatory reported on it IIRC?]. I think it is easier to just follow RFC 5280 6.3 and implement the check. Certainly experts who inspect the code would notice that the code deviates from the algorithm in an unsafe manner. I think the check is cheap?

> IssuingDistributionPoint isn't present in DecodedCertificateList that ASN1.js-rfc5080 provides, can we skip it?

That is painful, but not overly surprising given the general poor state of revocation support.  :(

Consider...
The only reason to verify a certificate is to defend against MITM attackers.
If you are being attacked via a MITM, it would not be uncommon that the attacker can also MITM the connection for the CRL. (e.g. ARP spoofing attack)
CAs should be partitioning their CRLs to make sure they stay a reasonable size during mass revocation events. [[Microsoft - an example of what happens when you don't partition](https://bugzilla.mozilla.org/show_bug.cgi?id=1965612)]

As an attacker, I just need to respond with valid partition Y when the client attempts to fetch partition Z. It would look like the certificate is not revoked to the client. The Issuing Distribution Point extension is intended to stop that attack.

Section 6.3 of RFC 5280 covers this.

> even if CRL is outdated, if a certificate is added to CRL is it there forever? So by checking nextUpdate we'll only skip unnecessary checks? (If the certificate has multiple distribution points, it will check all of them)

No - same as above. As an attacker who is MITMing the connection, I will give the client an outdated CRL that was issued before the certificate was revoked. The client should reject outdated CRLs.

sfc-gh-jkasten (commented):
Thanks and sorry for so much text It is a messy subject.

lib/agent/crl_utils.ts
sfc-gh-rsavenok: This mapping is OID to nodejs crypto.verify algrotihm. I'll rename `SUPPORTED_CRL_VERIFICATION_ALGORITHMS` to make it more clear what values means :)

lib/agent/crl_utils.ts
sfc-gh-rsavenok: `RSASSA-PSS` is not covered in OCSP - https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/agent/cert_util.js#L197-L235 

`ocsp.utils.sign` is the map that I added to other comment. The code shows that the signature verification doesn't use [AlgorithmIdentifier.parametrs](https://github.com/indutny/asn1.js/blob/master/rfc/5280/index.js#L51-L59) when calling crypto.verify method

Currently implemented map in this PR:
```
export const SUPPORTED_CRL_VERIFICATION_ALGORITHMS: Record<string, string> = {
  '1.2.840.113549.1.1.11': 'sha256WithRSAEncryption',
  '1.2.840.113549.1.1.12': 'sha384WithRSAEncryption',
  '1.2.840.113549.1.1.13': 'sha512WithRSAEncryption',
  '1.2.840.10045.4.3.2': 'SHA256',
  '1.2.840.10045.4.3.3': 'SHA384',
  '1.2.840.10045.4.3.4': 'SHA512',
};
```
Which OIDs besides `1.2.840.113549.1.1.10` I should implement? 

I'd like to keep rfc5080 instead of x509 (handles verification), because it's much faster. Will make a jira ticket to implement `RSASSA-PSS` and link it as todo in code.

lib/agent/crl_utils.ts
sfc-gh-rsavenok: We don't check reason fields in any driver (either its not present in cert raw data or our parsers doesn't fetch it) and pick the first HTTP URL in each distribution point:
https://github.com/snowflakedb/snowflake-connector-nodejs/blob/rsavenok/crl-2/test/unit/agent/crl_utils_test.ts#L156-L197

sfc-gh-jkasten: Discussed this a bit more over in Slack. I got lost as to where we were in the stack. 

I originally made a comment on whether or not this was purely about "redundancy". In this case it is. This is looping through all GeneralNames in a singular DistributionPoint. All URLs within it MUST point to the same CRL by RFC 5280 [[text](https://datatracker.ietf.org/doc/html/rfc5280#:~:text=When%20the%20distributionPoint%20field%20is%20present%2C%20it%20contains%20either%20a%0A%20%20%20SEQUENCE%20of%20general%20names%20or%20a%20single%20value%2C%20nameRelativeToCRLIssuer.%0A%20%20%20If%20the%20DistributionPointName%20contains%20multiple%20values%2C%20each%20name%0A%20%20%20describes%20a%20different%20mechanism%20to%20obtain%20the%20same%20CRL.)].

We could optimize to handle reasons efficiently as I mentioned in the previous comment, but that is outside the scope of the original comment.

Sorry for the noise.

lib/agent/crl_validator.ts
sfc-gh-rsavenok: Actually `IssuingDistributionPoint` is present in the decoded crl, I'll be adding a commit implementing it

lib/agent/crl_utils.ts
sfc-gh-jkasten: Thanks for the clarification. I should have traced the code and saw how the user (crypto.verify) of the value was handling it. One problem with only performing a partial review of the code :/

It looks like they are actually looking for digest/hash algorithms as opposed to signature algorithms.
https://nodejs.org/api/crypto.html#cryptocreateverifyalgorithm-options

They state that inserting some signature algorithms into verify() works, but not all of them. I am guessing the RSA signature algorithms you list fit that bill. I also suppose that replacing `sha256WithRSAEncryption` with `SHA256` would actually be preferred based upon their documentation?

They confuse different classes of algorithms together. It looks like they are stuck with a bit of an API mess.

lib/types/asn1.js-rfc5280.d.ts
```diff
@@ -0,0 +1,105 @@
...
+    validity: Validity;
+    subject: NameRDNSequence;
+    subjectPublicKeyInfo: SubjectPublicKeyInfo;
+    extensions?: (CrlDistributionPointsExtension | UnknownExtension)[];
+  }
+
+  export interface TBSCertList {
+    version?: BN;
+    signature: AlgorithmIdentifier;
+    issuer: any;
```
sfc-gh-jkasten: nit/optional: This is defined as the same ASN.1 type as both the subject and issuer of the TBSCertificate.

[X509 TBSCertificate.Subject ](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2:~:text=The%20subject%20field%20is%20defined%20as%20the%20X.501%20type%20Name.) == X.501 Name == [TBSCertList.Issuer](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2:~:text=The%20issuer%20field%20is%20defined%20as%20the%20X.501%0A%20%20%20type%20Name%2C%20and%20MUST%20follow%20the%20encoding%20rules%20for%20the%20issuer%20name%0A%20%20%20field%20in%20the%20certificate)

Ramifications - I think this can be NameRDNSequence which has a concrete definition and you might be able to avoid "stringifying" before the comparisons between the fields up above. 

Technically, the RFC 5280 algorithm requires the exact byte-for-byte match rather than the string representation of the fields. If this isn't possible, I don't think it is a big deal given the constraints within the public web PKI. (Just looks like a minor bug to someone who knows the algorithms and is inspecting the code to gauge their confidence in it.)

lib/types/asn1.js-rfc5280.d.ts
```diff
@@ -0,0 +1,105 @@
...
+        type: 'fullName';
+        value: {
+          type: 'uniformResourceIdentifier' | 'dNSName';
+          value: string;
+        }[];
+      };
+    }[]
+  >;
+
+  type Time = { type: 'utcTime' | 'generalTime'; value: number };
```
sfc-gh-jkasten, line 52: I assume this `value` is milliseconds since epoch? If so, maybe we can clarify it with a comment.

Normally, UTCTime and GeneralizedTime only have seconds granularity, but the value is clearly being used as milliseconds.

lib/agent/crl_utils.ts
sfc-gh-jkasten: It is a more involved encoding that uses essentially a tuple of OIDs (parameters) instead of a single OID to represent a tuple. The top level is `1.2.840.113549.1.1.10` ([rsassa-pss](https://oid-base.com/get/1.2.840.113549.1.1.10)) in all :/

* RSASSA-PSS with SHA-256, MGF-1 with SHA-256, and a salt length of 32 bytes  [[full encoding](https://lapo.it/asn1js/#MEEGCSqGSIb3DQEBCjA0oA8wDQYJYIZIAWUDBAIBBQChHDAaBgkqhkiG9w0BAQgwDQYJYIZIAWUDBAIBBQCiAwIBIA)]
* RSASSA-PSS with SHA-384, MGF-1 with SHA-384, and a salt length of 48 bytes [[full encoding](https://lapo.it/asn1js/#MEEGCSqGSIb3DQEBCjA0oA8wDQYJYIZIAWUDBAICBQChHDAaBgkqhkiG9w0BAQgwDQYJYIZIAWUDBAICBQCiAwIBMA)]
* RSASSA-PSS with SHA-512, MGF-1 with SHA-512, and a salt length of 64 bytes [[full encoding](https://lapo.it/asn1js/#MEEGCSqGSIb3DQEBCjA0oA8wDQYJYIZIAWUDBAIDBQChHDAaBgkqhkiG9w0BAQgwDQYJYIZIAWUDBAIDBQCiAwIBQA)]

Looking more closely at the code, a TODO is absolutely fine. We should receive advanced warning before a CA switches and I imagine a lot of other things would break.

lib/agent/crl_validator.ts
```diff
@@ -16,50 +25,118 @@ export type CRLValidatorConfig = {
...
+}
+
 export function isCrlValidationEnabled(config: CRLValidatorConfig) {
   return config.checkMode !== 'DISABLED';
 }
 
 export function corkSocketAndValidateCrl(socket: TLSSocket, config: CRLValidatorConfig) {
-  socket.once('secureConnect', () => {
+  socket.once('secureConnect', async () => {
     const certChain = socket.getPeerCertificate(true);
```
sfc-gh-jkasten, line 42: Outside the scope of this PR: This function concerns me a bit. There are usually multiple chains associated with every set of certificates returned by the server. This seems to provide back a singular chain.

Here are some ["real-world" chains](https://docs.google.com/document/d/1cgkHlAqFBmF0qrnSgdwvGp1RhJdx0w2nayIl_iJIknE/edit?tab=t.l6soesqfmoiq#heading=h.bouoatrg7sjk) to demonstrate the complexity. We only need to find one path to a trusted source and we don't want to choke on a certificate that isn't necessary to the validation.

Do you happen to know if this dumps all certificates the server sent or just a validated chain? If you don't know, no worries, we can investigate offline.

lib/agent/crl_validator.ts
```diff
@@ -16,50 +25,118 @@ export type CRLValidatorConfig = {
...
+    const issuerSubject = JSON.stringify(decodedIssuerCertificate.tbsCertificate.subject);
+    const issuerPublicKey = crypto
+      .createPublicKey({
+        key: certificate.issuerCertificate.pubkey as Buffer,
+        format: 'der',
+        type: 'spki',
+      })
+      .export({ format: 'pem', type: 'spki' }) as string;
+
+    for (const crlUrl of crlUrls) {
```
sfc-gh-jkasten, line 112: Can we mark the issuing distribution point check as a TODO if you want to submit before that lands? (Just want to make sure the vuln is noted.)

lib/agent/crl_validator.ts
sfc-gh-rsavenok: Unlike Go/Java, node doesn't support alternative chains. It gives only 1 validated chain that was negotiated during the handshake.

The node flow:
* cork (queue) all data coming from the connection
* Set up a valid TLS connection
* validate crl on `secureConnect` event (valid connection established)
* if crl valid -> uncork the connection
* if not, destroy the connection

lib/agent/crl_validator.ts
sfc-gh-rsavenok: Implemented in same PR with https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1162/commits/0d7e21d2a59949715ad0e5f2aabceb9e0489f329

lib/agent/crl_utils.ts
sfc-gh-rsavenok: Renamed to `CRL_SIGNATURE_OID_TO_CRYPTO_DIGEST_ALGORITHM`

lib/agent/crl_utils.ts
sfc-gh-rsavenok: Updated code with todo link

lib/types/asn1.js-rfc5280.d.ts
sfc-gh-rsavenok: Both fields are actually `NameRDNSequence` (fixed type)

The format is:
```
issuerSubject {
  type: 'rdnSequence',
  value: [ [ [Object] ], [ [Object] ], [ [Object] ] ]
}
```
Where each Object is:
```
[ { type: [ 2, 5, 4, 6 ], value: <Buffer 13 02 55 53> } ]
```

While I can write an iterator to compare each item type and Buffer with Buffer, it's much easier just to run 
```
JSON.stringify(issuer.subject) === JSON.stringify(crl.issuer)
```

This would compare 2 strings like this:
```
{"type":"rdnSequence","value":[[{"type":[2,5,4,6],"value":{"type":"Buffer","data":[19,2,85,83]}}],[{"type":[2,5,4,10],"value":{"type":"Buffer","data":[19,12,68,105,103,105,67,101,114,116,32,73,110,99]}}],[{"type":[2,5,4,3],"value":{"type":"Buffer","data":[19,42,68,105,103,105,67,101,114,116,32,71,108,111,98,97,108,32,71,50,32,84,76,83,32,82,83,65,32,83,72,65,50,53,54,32,50,48,50,48,32,67,65,49]}}]]}
```

sfc-gh-jkasten: Thanks for the breakdown and examples! Sounds great! The string comparison is effectively doing the same thing.

lib/agent/crl_validator.ts
sfc-gh-jkasten: Thanks! It is probably just an upstream concern then. I do know that nodepq doesn't handle certificate branching correctly, so it wouldn't surprise me of any deficiencies.

I think Sean is working on some integration tests which should give us better ammo to work with.

lib/agent/crl_utils.ts
```diff
@@ -1,99 +1,157 @@
...
+) {
+  for (const revokedCert of crl.tbsCertList.revokedCertificates) {
+    if (revokedCert.userCertificate.eq(decodedCertificate.tbsCertificate.serialNumber)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+export function isCrlDistributionPointExtensionValid(
```
sfc-gh-jkasten:
```suggestion
export function isIssuingDistributionPointExtensionValid(
```
nit: Attempt to avoid name collisions between the two extensions.
(CRL Distribution Point is an extension defined for the certificate. This is testing the Issuing Distribution Point extension defined for CRLs)

lib/agent/crl_validator.ts
sfc-gh-jkasten: Thanks!

lib/agent/crl_validator.ts
```diff
@@ -16,50 +26,123 @@ export type CRLValidatorConfig = {
...
+      logDebug(`validating ${crlUrl} issuer`);
+      const crlIssuer = JSON.stringify(crl.tbsCertList.issuer);
+      if (issuerSubject !== crlIssuer) {
+        throw new Error(
+          `CRL ${crlUrl} issuer is invalid. Expected ${issuerSubject} but got ${crlIssuer}`,
+        );
+      }
+
+      logDebug(`validating ${crlUrl} nextUpdate`);
+      if (crl.tbsCertList.nextUpdate.value < Date.now()) {
```
sfc-gh-knozderko, line 137: will it work well if nextUpdate is in different time zone than the local time?

test/unit/agent/crl_utils_test.ts
```diff
@@ -1,224 +1,287 @@
...
+      name: 'returns false for 11-day certificate in 2024-2026 period',
+      notBefore: 'Mar 15 2024 00:00:00 GMT',
+      notAfter: 'Mar 26 2024 00:00:00 GMT',
+      expectedResult: false,
+    },
+    {
+      name: 'returns true for 10 days + 1 minute certificate in 2024-2026 period',
+      notBefore: 'Mar 15 2024 00:00:00 GMT',
+      notAfter: 'Mar 25 2024 00:00:59 GMT',
+      expectedResult: true,
```
sfc-gh-knozderko, line 36: actually it is 10 days + 59 seconds.
For 10 days + 1 minute it should be notAfter: 'Mar 25 2024 00:01:00 GMT' and the result should be false.

test/unit/agent/crl_utils_test.ts
```diff
@@ -1,224 +1,287 @@
...
+    {
+      name: 'returns false for 8-day certificate in 2026+ period',
+      notBefore: 'Mar 15 2026 00:00:00 GMT',
+      notAfter: 'Mar 23 2026 00:00:00 GMT',
+      expectedResult: false,
+    },
+    {
+      name: 'returns true for 7 days + 1 minute certificate in 2026+ period',
+      notBefore: 'Mar 15 2026 00:00:00 GMT',
+      notAfter: 'Mar 22 2026 00:00:59 GMT',
```
sfc-gh-knozderko, line 60: actually it is 7 days + 59 seconds.
For 7 days + 1 minute it should be notAfter: 'Mar 22 2026 00:01:00 GMT' and the result should be false.

test/unit/agent/crl_utils_test.ts
```diff
@@ -1,224 +1,287 @@
...
-              {
-                type: 'uniformResourceIdentifier',
-                value: 'http://crl2.example.com/cert.crl',
-              },
-            ],
-          },
-        },
-      ],
+      certificateParams: {
+        crlUrls: ['http://crl1.example.com/cert.crl', 'http://crl2.example.com/cert.crl'],
```
sfc-gh-knozderko: actually if there are multiple crl urls it was specified in the algorithm that for all the urls we should verify revocation status. Do we have this logic for all the urls?

sfc-gh-rsavenok: There are 2 different cases:
1) you have 1 distribution point with multiple crl urls. Only 1 http url must be picked. Usually there's one, or multiple entries with different protocols. But the spec allows any amount of fullName entries
2) there are multiple distribution points - this should fetch url from each and is covered in https://github.com/snowflakedb/snowflake-connector-nodejs/blob/rsavenok/crl-2/test/unit/agent/crl_utils_test.ts#L133-L141 this test covers

I'll adjust param names to make it more clear what crlUrls param for test certificate does

lib/agent/crl_validator.ts
sfc-gh-rsavenok: Yes, both are numbers in ms since epoch in utc

test/unit/agent/crl_utils_test.ts
sfc-gh-rsavenok: Fixed in https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1162/commits/2aa55c7aab1a9ecf80d2ad15e767f24069bb4fa4

test/unit/agent/crl_utils_test.ts
sfc-gh-rsavenok: Fixed in https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1162/commits/2aa55c7aab1a9ecf80d2ad15e767f24069bb4fa4

test/unit/agent/crl_utils_test.ts
sfc-gh-rsavenok: Fixed in https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1162/commits/2aa55c7aab1a9ecf80d2ad15e767f24069bb4fa4

test/unit/agent/crl_utils_test.ts
sfc-gh-jkasten: The `notAfter` field is inclusive. I believe the original was correct [[RFC 5280 Section 4.1.2.5](https://datatracker.ietf.org/doc/html/rfc5280#:~:text=The%20validity%20period%20for%20a%20certificate%20is%20the%20period%20of%20time%20from%0A%20%20%20notBefore%20through%20notAfter%2C%20inclusive.)]

The reason this minute padding exists though to begin with is that we don't want to have to worry about leap seconds or anything else affecting the calculation. 

I think the only goal is to make sure we don't enforce long-lived certificate requirements on "short-lived certificates" as defined by the Baseline Requirements.

lib/agent/crl_utils.ts
```diff
@@ -1,99 +1,157 @@
...
 
-  // Certificates issued before March 15, 2024 are not considered short-lived
-  if (validFrom < new Date('2024-03-15T00:00:00.000Z')) {
-    return false;
-  }
+/**
+ * See Short-lived Subscriber Certificate section\
+ * https://cabforum.org/working-groups/server/baseline-requirements/requirements/
+ */
+export function isShortLivedCertificate(decodedCertificate: ASN1.CertificateDecoded) {
```
sfc-gh-rkowalski, line 71: Why `validFrom < new Date('2024-03-15T00:00:00.000Z'` was removed?

test/unit/agent/crl_utils_test.ts
sfc-gh-jkasten: The notAfter field is inclusive. I believe the original was correct [[RFC 5280 Section 4.1.2.5](https://datatracker.ietf.org/doc/html/rfc5280#:~:text=The%20validity%20period%20for%20a%20certificate%20is%20the%20period%20of%20time%20from%0A%20%20%20notBefore%20through%20notAfter%2C%20inclusive.)]

The reason this minute padding exists though to begin with is that we don't want to have to worry about leap seconds or anything else affecting the calculation.

I think the only goal is to make sure we don't enforce long-lived certificate requirements on "short-lived certificates" as defined by the Baseline Requirements.

lib/agent/crl_utils.ts
sfc-gh-rsavenok: image

PR #1163, opened 2025-09-08
SNOW-2117146 CRL caching

sfc-gh-rsavenok:
### Description
- implement in-memory caching
- implement disk caching
- prevent multiple calls downloading the same CRL

### Checklist

### DIFF
--- a/lib/agent/crl_utils.ts
+++ b/lib/agent/crl_utils.ts
@@ -1,17 +1,7 @@
 import { DetailedPeerCertificate } from 'tls';
 import crypto from 'crypto';
 import ASN1 from 'asn1.js-rfc5280';
-import axios from 'axios';
 import Logger from '../logger';
-import GlobalConfigTyped from '../global_config_typed';
-import {
-  clearExpiredCrlFromDiskCache,
-  clearExpiredCrlFromMemoryCache,
-  getCrlFromDisk,
-  getCrlFromMemory,
-  setCrlInMemory,
-  writeCrlToDisk,
-} from './crl_cache';
 
 // TODO:
 // Implement RSASSA-PSS signature verification


lib/agent/crl_validator.ts
```diff
@@ -105,7 +105,11 @@ export async function validateCrl(certChain: DetailedPeerCertificate, config: CR
 
     for (const crlUrl of crlUrls) {
       logDebug(`fetching ${crlUrl}`);
-      const crl = await getCrl(crlUrl);
+      const crl = await getCrl(crlUrl, {
```
sfc-gh-rsavenok, line 113: These options are drilled down from the connection config. I'll submit a separate PR moving them to the global config

lib/agent/crl_cache.ts
```diff
@@ -0,0 +1,89 @@
...
+    crl,
+  });
+}
+
+export function getCrlCacheDir() {
+  return path.join(getDefaultCacheDir(), 'crls');
+}
+
+export function crlUrlToFileName(url: string) {
+  return encodeURIComponent(url).replace(/%/g, '_');
```
sfc-gh-pfus: Why do we need this replace?

lib/agent/crl_cache.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// NOTE:
+// Parsing CRLs is expensive, so we store expiration time in the file name
```
sfc-gh-pfus: What is the reasoning here? It's not aligning with the design.

lib/agent/crl_cache.ts
```diff
@@ -0,0 +1,89 @@
...
+export async function getCrlFromDisk(url: string) {
```
sfc-gh-pfus, line 68: This function may leverage some logs.

lib/agent/crl_cache.ts
sfc-gh-rsavenok: Because `\ / : * ? " < > |` are invalid characters in file names. But I guess I could just drop `http://` and have nicer file names :P

lib/agent/crl_cache.ts
sfc-gh-rsavenok: Node IO is being blocked while CRL is parsing. Current implementation is 500ms per CRL, but we might need to switch to slower library with 2-3sec per CRL. With such IO blocks, I don't like the idea of periodic cache clearing of entire folder where every file will be parsed

What's the downside of this? That someone could rename the file and trick the driver into thinking that is still valid? I could add an extra check in this block to ensure nextUpdate is ok:
```
} else if (crlFileName === crlUrlToFileName(url)) {
        const rawCrl = await fs.readFile(path.join(cacheDir, fileName));
        const decodedCrl = ASN1.CertificateList.decode(rawCrl, 'der');
        return decodedCrl;
      }
```

lib/agent/crl_cache.ts
sfc-gh-pfus: I meant the last replace, which replaces `%` char with underscore? The rest should be handled by URL encoding.

lib/agent/crl_cache.ts
sfc-gh-pfus: You can trick cache file name, but it should still be checked later - and if `nextUpdate` is outdated in the main algorithm, new CRL should be downloaded either way. Common file names were designed so various drivers on the same machine can reuse the same downloaded CRLs. I think that there are ways to satisfy both (slow JS and reusing CRLs among drivers) by introducing JS specific features like JSON file with expiry times of particular files, but I'm convinced. The choice is yours.

sfc-gh-pfus (approved):

lib/agent/crl_cache.ts
sfc-gh-rsavenok: Good feedback! I will add index file to keep same names as other drivers have

sfc-gh-rsavenok: Instead of index file, agreed to clear anything older than 30 days as it's much easier to code & test

sfc-gh-rkowalski (approved):
LGTM

PR #1171, opened 2025-09-16
SNOW-2117146 Move some CRL options to GlobalConfig

sfc-gh-rsavenok:
### Description
Related to CRL caching implemented in pending PR https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1163

* Moved some config options to the global config
* created new GlobalConfig with TypeScript 
* added a NOTE with migration plan

#### Notes on migration plan
Developers are reading our source code, and it's very difficult for them to understand what's going on in our config. There are 5000+ lines of config-related code with undocumented defaults, undocumented env variables, mistakes in value validators, and inconsistent error codes. All of this can be reduced to ~500 lines of code with Typescript + default values map.

Other popular SDKs don't do param validation either and rely on defaults, TS, and runtime errors. 

### Checklist

### DIFF
--- a/lib/global_config_typed.ts
+++ b/lib/global_config_typed.ts
@@ -1,0 +1,69 @@
+import Logger from './logger';
+
+/*
+ * NOTE:
+ * Work In Progress migration of global_config.js to TypeScript.
+ *
+ * Instead of gettter+setter for each config option, we should use a signle object
+ * Instead of manual input validation, we should rely on TypeScript's type system
+ * This will simplify erorr codes, typing, mocking and testing = less code to maintain
+ *
+ * TODO:
+ * When all global_config.js options are migrated:
+ * - rename to global_config.ts
+ * - removed "typed" from variable & function names
+ */
+export interface GlobalConfigOptionsTyped {
+  /**
+   * Enable CRL caching in memory. Cached entries are removed after crlCacheValidityTime.
+   *
+   * @default true
+   */
+  crlInMemoryCache: boolean;
+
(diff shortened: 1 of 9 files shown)


lib/global_config_typed.ts
```diff
@@ -0,0 +1,69 @@
...
+   * @default 86400000 (24 hours in ms)
+   */
+  crlCacheValidityTime: number;
+}
+
+export const GLOBAL_CONFIG_DEFAULTS: GlobalConfigOptionsTyped = {
+  crlInMemoryCache: true,
+  crlOnDiskCache: true,
+  crlDownloadTimeout: 10000,
+  crlCacheValidityTime: 86400000,
```
sfc-gh-dheyman, line 70: are these correct parameter names? I've seen some different naming in other drivers(e.g. JDBC)

sfc-gh-rsavenok: JDBC (CRL_CACHE_VALIDITY_TIME) and Go (SNOWFLAKE_CRL_CACHE_VALIDITY_TIME) are reading from env variables. Since in node we have `.configure()` for entire driver, I think it needs to be there. 

If it needs to be set from env variable, driver users would be able to do:
```
snowflake.configure({
  crlCacheValidityTime: proces.env.WHATEVER_ENV_THEY_WANT
});
```

Also, looking at other implementations of `crlInMemoryCache` and `crlOnDiskCache`, it must be configurable per connection. I'll change it before merging the PR.

Issue #1093, opened 2025-06-08
SNOW-2143625: Bind type is too narrow


mootari:
The TypeScript type `Bind` is currently defined as `string | number`([source](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/a9174fb7e61e8a896a6b066569fcbcd68b76054a/index.d.ts#L182C17-L182C21)).

However, the validating code actually only requires that the bound value can be stringified to JSON:
https://github.com/snowflakedb/snowflake-connector-nodejs/blob/a9174fb7e61e8a896a6b066569fcbcd68b76054a/lib/connection/statement.js#L329-L333

Even when just considering scalar values the type would therefore have to be:
```ts
export type Bind = string | number | boolean | null;
```

PR #1176, opened 2025-09-29
SNOW-2143625 add boolean and null to Bind typing

sfc-gh-dszmolka:
### Description

Addresses #1093 

Since we check if the bind variables' array is JSON.stringifiable ([ref](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/v2.2.0/lib/connection/statement.js#L363-L369)) and doesn't return `undefined`, in principle then the typing should also reflect it and list the objects which are supposed to be correctly stringifiable (have `toJSON()` method) 

We already had a complaint about this, see above Issue. So extending the Bind typing accordingly.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Improved debug logs when dowloading query result chunks (snowflakedb/snowflake-connector-nodejs#1142)
 - Fixed missing error handling in `getResultsFromQueryId()` (snowflakedb/snowflake-connector-nodejs#1173)
 - Fixed invalid transformation of `null` values to `""` when using stage binds (snowflakedb/snowflake-connector-nodejs#1166)
+- Extended typing of Bind (issue #1093)
 
 ## Prior Releases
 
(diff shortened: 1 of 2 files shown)


sfc-gh-rsavenok (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Improved debug logs when dowloading query result chunks (snowflakedb/snowflake-connector-nodejs#1142)
 - Fixed missing error handling in `getResultsFromQueryId()` (snowflakedb/snowflake-connector-nodejs#1173)
 - Fixed invalid transformation of `null` values to `""` when using stage binds (snowflakedb/snowflake-connector-nodejs#1166)
+- Extended typing of Bind (issue #1093)
```
sfc-gh-rsavenok: If you'll add `snowflakedb/snowflake-connector-nodejs#1090`, GH will automatically understand that it's a link to an issue and add tooltip to it

sfc-gh-dheyman (approved):

CHANGELOG.md
sfc-gh-dszmolka: initially used link to issue, but now added magic link to PR as suggested

PR #1184, opened 2025-10-09
SNOW-2398694 Fix encryption in PUT operations

sfc-gh-rsavenok:
#### Cause of the bug
* When the driver sends a PUT query to the back-end, it replies with metadata that the driver uses to encrypt the file. One of the metadata fields is `"smkId": 83807700463646` (number)
* When this number is higher than JavaScript `Number.MAX_SAFE_INTEGER` (9007199254740991), the driver encrypts the file with the wrong smkId
* There's a hack in NodeJS that converts the server response before passing it to the JSON parser: integer `"smkId": 83807700463646` converted to string `"smkId": "83807700463646"`
* Because of code regression caused by #1173, this hack wasn't executed

### Checklist

### DIFF
--- a/test/integration/testPutSmkIdStringConversion.ts
+++ b/test/integration/testPutSmkIdStringConversion.ts
@@ -2,6 +2,7 @@ import assert from 'assert';
 import sinon from 'sinon';
 import rewiremock from 'rewiremock/node';
 import os from 'os';
+import path from 'path';
 
 const OriginalFileTransferAgent = require('../../lib/file_transfer_agent/file_transfer_agent');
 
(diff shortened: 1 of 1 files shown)


lib/services/sf.js
```diff
@@ -645,28 +645,7 @@ function StateAbstract(options) {
             Logger.getInstance().debug('HTTP Error: %s', response.statusCode);
           }
         } else {
-          // if the response body is a non-empty string and the response is
```
sfc-gh-rkowalski, line 648: Are you double sure that removing this code will not cause some regression?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 2.4.0 (TBD)
 
+## 2.3.1
```
sfc-gh-dheyman, line 5: do we want to add info about unpublishing? like `2.3.0 (unpublished due to...)`

lib/services/sf.js
sfc-gh-rsavenok: Yes, this code is not executed anymore, because the response comes to this function already parsed. It's already parsed, because #1173 added `parseResponseData(response);`

test/integration/testPutSmkIdStringConversion.ts
```diff
@@ -0,0 +1,40 @@
+import assert from 'assert';
+import sinon from 'sinon';
+import rewiremock from 'rewiremock/node';
+import os from 'os';
+
+const OriginalFileTransferAgent = require('../../lib/file_transfer_agent/file_transfer_agent');
+
+describe('smkId patching in PUT statements', () => {
```
sfc-gh-dheyman, line 9: would it be beneficial to add wiremock test with mapping that includes smkId >  max int value in response?

CHANGELOG.md
sfc-gh-rsavenok: It won't be unpublished, we don't meet unpublishing criteria. We'll be marking it as deprecated on npm - https://docs.npmjs.com/policies/unpublish

test/integration/testPutSmkIdStringConversion.ts
sfc-gh-rsavenok: Good point, I'll rewrite this test to work with WireMock in a separate PR to not block the release

sfc-gh-rsavenok: Actually, it's not possible to create such a mock with our current stack:

`wiremock-rest-client` is parsing mapping files as JSON before registering them, so if we have a JSON mapping file with `251084776040741626364`, the parsing step will change it to `251084776040741630000` before registering the mock :/

PR #1192, opened 2025-10-30
SNOW-2476142 add new when creating snowflake https proxy agent

sfc-gh-dszmolka:
Description
Looks like v2.1.1 introduced a serious regression which ca be triggered when both
* OCSP checking (enabled by default)
* HTTP_PROXY env. variable

was enabled. Creating the agent fails with
```
{"level":"DEBUG","message":"[7:13:52.073 AM]: Encountered an error when sending the request. Details: {\"stack\":\"TypeError: Class constructor SnowflakeHttpsProxyAgent cannot be invoked without 'new'\\n    at exports.secureSocket (/test/node_modules/snowflake-sdk/dist/lib/agent/socket_util.js:56:21)\\n    at HttpsOcspAgent.agent.createConnection (/test/node_modules/snowflake-sdk/dist/lib/agent/https_ocsp_agent.js:34:27)\\n    at Agent.createSocket (node:_http_agent:324:26)\\n    at Agent.addRequest (node:_http_agent:276:10)\\n    at new ClientRequ"}
```

and thus driver cannot proceed.
The regression did not affect use-case where HTTPS_PROXY was used, or if OCSP was turned off.
Verified, also doesn't affect CRL = ENABLED (with OCSP turned off, of course)

Validated the issue and the fix both for HTTP_PROXY and HTTPS_PROXY use-cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "mocha": "^11.1.0",
     "mock-require": "^3.0.3",
     "nyc": "^15.1.0",
-    "prettier": "^3.5.3",
+    "prettier": "^3.7.1",
     "rewiremock": "^3.14.6",
     "sinon": "^20.0.0",
     "test-console": "^2.0.0",
--- a/test/unit/authentication/auth_workload_identity/attestation_aws_test.ts
+++ b/test/unit/authentication/auth_workload_identity/attestation_aws_test.ts
@@ -33,9 +33,8 @@ describe('Attestation AWS', () => {
       },
     });
     rewiremock.enable();
-    AttestationAws = await import(
-      '../../../../lib/authentication/auth_workload_identity/attestation_aws'
-    );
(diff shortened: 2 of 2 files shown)


system_test/testProxy.js
```diff
@@ -1,54 +0,0 @@
-const snowflake = require('../lib/snowflake');
-const async = require('async');
-const testUtil = require('../test/integration/testUtil');
-const connOptions = require('./connectionOptions');
-
-describe.skip('testProxy', function () {
```
sfc-gh-rsavenok, line 6: This file is deleted as basically same test is performed in testProxyExecute

sfc-gh-snoonan (approved):

sfc-gh-dszmolka:
thank you @sfc-gh-rsavenok for your guidance and significant contribution to this !

Issue #978, opened 2024-12-02
SNOW-1842116: Typescript: `getResultsFromQueryId` expects option of type `StatementOption`. `sqlText` should therefore be optional.


sjclemmy:
The typescript type for the `connect.getResultsFromQueryId` `options` argument is `StatementOption`, This is shared with `connection.execute` so I would expect the `sqlText` to be optional.

It would probably be a better idea to give `getResultsFromQueryId` its own argument type. 
This code yields a typescript error
```
    connection.getResultsFromQueryId({
      queryId: queryId,
      complete: function (err, stmt, rows) {
        if (err) {
          console.error(
            "Failed to execute statement due to the following error: " +
              err.message
          );
        } else {
          console.log(rows);
        }
      },
    });
```
```
Argument of type '{ queryId: string; complete: (err: SnowflakeError | undefined, stmt: RowStatement | FileAndStageBindStatement, rows: any[] | undefined) => void; }' is not assignable to parameter of type 'StatementOption'.
  Property 'sqlText' is missing in type '{ queryId: string; complete: (err: SnowflakeError | undefined, stmt: RowStatement | FileAndStageBindStatement, rows: any[] | undefined) => void; }' but required in type 'StatementOption'.ts(2345)
index.d.ts(604, 9): 'sqlText' is declared here.
```

This code passes the typescript validation, but I don't know if it actually works.
```
    connection.getResultsFromQueryId({
      // not optional (bug) so need to pass empty string
      sqlText: "",
      queryId: queryId,
      complete: function (err, stmt, rows) {
        if (err) {
          console.error(
            "Failed to execute statement due to the following error: " +
              err.message
          );
          reject(err);
        } else {
          resolve(rows);
        }
      },
    });
```

sfc-gh-dszmolka:
hi - thanks for raising this and providing the details too! we'll take a look

PR #1197, opened 2025-11-03
SNOW-1842116 Fix getResultsFromQueryId typing

sfc-gh-rsavenok:
### Description

Fixes https://github.com/snowflakedb/snowflake-connector-nodejs/issues/978

`getResultsFromQueryId` doesn't work with a full set of parameters used by statement builder. It only executes `createStatementPostExec`, thus different logic should be applied:
* `queryId` - must be required, as missing it will throw a runtime error anyway
* any other parameter of `StatementOption` should be optional (it has only 1 required: `sqlText`)

This method API is weird and requires more attention as `await` on this method doesn't wait for the result to be fetched and not every config option from `StatementOption` makes sense in this context. I'll be revisiting the api in a scope of async/await API redesign (SNOW-832857)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Upcoming Release
 
-- Fixed TypeScript definition for `getResultsFromQueryId`: `queryId` should be required and `sqlText` should be optional (snowflakedb/snowflake-connector-nodejs#1197)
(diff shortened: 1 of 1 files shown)


sfc-gh-fpawlowski (approved):
lgtm, but tests seem to be failing. Is there a known issue or is it caused by this PR?

PR #1200, opened 2025-11-05
NO-SNOW: Fixing CICD test flakiness

sfc-gh-pcyrek:
### Description

### Checklist

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -24,6 +24,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


test/integration/testLargeResultSet.js
```diff
@@ -25,15 +25,28 @@ describe('Large result Set Tests', function () {
         testUtil.checkError(err);
         const stream = stmt.streamRows();
         let rowCount = 0;
+        let doneCalled = false;
```
sfc-gh-rsavenok, line 28: I don't think this is the right approach. Node event emitter stream should produce only 1 end value: either error or end, not both. 

So seems like there's a bug in our RowStream implementation where under some cases it produces 2 events

PR #1207, opened 2025-11-21
NO-SNOW Fix slow & flaky tests

sfc-gh-rsavenok:
### Description

* `test/integration/ocsp_mock/testConnectionOcspMock.js` -> ridiculously slow with 2 tests taking 6 minutes to complete because of `/login-request` retries. Solved by reducing retry attempts to 2 and mocking the backoff strategy to 100ms repeats
* `test/integration/testHTAP.js` 
  * fixed error printing
  * actual error seems to indicate that 2 processes use the same dbname - fixed with uuids
  * `this.retries(3)` replaced with timeout as test is often on the edge of default execution time
 * `test/integration/testLargeResultSet.js` -> printing wrong error on stream error
 * `test/integration/testOcsp.js` -> wrong loop logic that ends the test while 1 query is still in progress

### Checklist

### DIFF
--- a/test/integration/testHTAP.js
+++ b/test/integration/testHTAP.js
@@ -1,14 +1,13 @@
 const assert = require('assert');
 const async = require('async');
+const uuid = require('uuid');
 const connOption = require('./connectionOptions').valid;
 const testUtil = require('./testUtil');
 
-function getRandomDBNames() {
-  const dbName = 'qcc_test_db';
+function getUniqueDBNames(amount = 3) {
   const arr = [];
-  const randomNumber = Math.floor(Math.random() * 10000);
-  for (let i = 0; i < 3; i++) {
-    arr.push(dbName + (randomNumber + i));
+  for (let i = 0; i < amount; i++) {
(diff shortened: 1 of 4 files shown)


test/integration/testHTAP.js
```diff
@@ -1,24 +1,25 @@
...
+    arr.push(`qcc_test_db_${Date.now()}_${uuid.v4().replaceAll('-', '_')}`);
   }
   return arr;
 }
 
 // Only the AWS servers support the hybrid table in the GitHub action.
 if (process.env.CLOUD_PROVIDER === 'AWS') {
   describe('Query Context Cache test', function () {
-    this.retries(3);
+    // Longer timeout as each create hybrid table takes ~45s to complete
```
sfc-gh-rkowalski, line 18: wow, why it is so slow? is this expected?

sfc-gh-rsavenok: ¯\_(ツ)_/¯, I executed this query in the worksheets and it's just how long the x-small warehouse needs to create hybrid table. So not a driver fault and I don't want to spend time investigating it ATM. 

Maybe in the future I will as it will make total test runs 7 minutes instead of 10

PR #1212, opened 2025-11-29
SNOW-2465271 Minicore telemetry

sfc-gh-rsavenok:
### Description

Implements minicore integration

RA result: SNOW-2465271_ra-1764775992420

### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -4,7 +4,7 @@ about: Something isn't working as expected? Here is the right place to report.
 labels: bug
 ---
 
-If you need **urgent assistance** then [file a case with Snowflake Support](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge).
+:exclamation: If you need **urgent assistance** then [file a case with Snowflake Support](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge).
 Otherwise continue here.
 
 Please answer these questions before submitting your issue.
--- a/.github/workflows/jira_issue.yml
+++ b/.github/workflows/jira_issue.yml
@@ -13,38 +13,92 @@ jobs:
       issues: write
     if: ((github.event_name == 'issue_comment' && github.event.comment.body == 'recreate jira' && github.event.comment.user.login == 'sfc-gh-mkeller') || (github.event_name == 'issues' && github.event.pull_request.user.login != 'whitesource-for-github-com[bot]'))
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          repository: snowflakedb/gh-actions
-          ref: jira_v1
-          token: ${{ secrets.SNOWFLAKE_GITHUB_TOKEN }} # stored in GitHub secrets
-          path: .
-
-      - name: Login
-        uses: atlassian/gajira-login@v2.0.0
+      - name: Create JIRA Ticket
+        id: create
         env:
           JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
           JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
           JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+          ISSUE_BODY: ${{ github.event.issue.body }}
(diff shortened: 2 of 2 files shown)


lib/minicore/is_musl.ts
```diff
@@ -0,0 +1,63 @@
...
+  return musl;
+}
+
+function isFileMusl(filePath: string): boolean {
+  return filePath.includes('libc.musl-') || filePath.includes('ld-musl-');
+}
+
+function isMuslFromFilesystem(): boolean | null {
+  try {
+    return readFileSync('/usr/bin/ldd', 'utf-8').includes('musl');
```
sfc-gh-pfus, line 28: I'm not sure we should hardcode `ldd` path. It may be in `/bin` or sth.

lib/minicore/minicore.ts
```diff
@@ -0,0 +1,60 @@
...
+let logDebugMinicoreError: unknown | null = null;
+
+if (process.env.SNOWFLAKE_DISABLE_MINICORE) {
+  minicoreStatus.error = 'Minicore is disabled with SNOWFLAKE_DISABLE_MINICORE env variable';
+} else {
+  try {
+    const binaryName = getBinaryName();
+    const minicoreModule = require(`./binaries/${binaryName}`) as typeof import('./binaries');
+    minicoreStatus.version = minicoreModule.sfCoreFullVersion();
+    minicoreStatus.binaryName = binaryName;
```
sfc-gh-pfus: Maybe we should do this earlier (after line 21)? In case of errors we lose it now.

lib/minicore/minicore.ts
```diff
@@ -0,0 +1,60 @@
...
+  binaryName: null,
+  error: null,
+};
+let logDebugMinicoreError: unknown | null = null;
+
+if (process.env.SNOWFLAKE_DISABLE_MINICORE) {
+  minicoreStatus.error = 'Minicore is disabled with SNOWFLAKE_DISABLE_MINICORE env variable';
+} else {
+  try {
+    const binaryName = getBinaryName();
```
sfc-gh-pfus: Is it going to be sync or async (the whole loading process)?

lib/minicore/is_musl.ts
sfc-gh-rsavenok: This entire file is copy-pate from napi-rs and I don't want to chenge it. There's a comment in minicore.ts explaining why napi-rs autogenerated loader isn't used for minicore

lib/minicore/minicore.ts
sfc-gh-rsavenok: It's sync. Don't see a reason for it to be async as the performance is +- the same as requiring any other javascript file.

lib/services/sf.js
```diff
@@ -1079,8 +1080,13 @@ StateConnecting.prototype.continue = function () {
   // if we have some information about the client environment, add it as well
   const clientEnvironment = this.connectionConfig.getClientEnvironment();
   if (Util.isObject(clientEnvironment)) {
+    const minicoreStatus = getMinicoreStatus();
     clientInfo.CLIENT_ENVIRONMENT = {
       CERT_REVOCATION_CHECK_MODE: this.connectionConfig.crlValidatorConfig.checkMode,
+      ISA: process.arch,
+      CORE_VERSION: minicoreStatus.version,
+      CORE_BINARY_NAME: minicoreStatus.binaryName,
```
sfc-gh-mgemra: I think it's called `CORE_FILE_NAME` in other drivers and in the design.

lib/minicore/minicore.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const minicoreStatus: MinicoreStatus = {
+  version: null,
+  binaryName: null,
+  error: null,
+};
+let logDebugMinicoreError: unknown | null = null;
+
+if (process.env.SNOWFLAKE_DISABLE_MINICORE) {
```
sfc-gh-mgemra, line 17: Can this be made async like in other drivers and in the design? Can't you use `Promise` + `async` and return `NOT_YET_LOADED` in case it takes too long?

lib/services/sf.js
sfc-gh-rsavenok: Good catch!

lib/minicore/minicore.ts
sfc-gh-rsavenok: I'd like it to be required top-level and sync as this is how it will work in universal driver: the application won't start until all imports are loaded.

Loading it in async way means that we might lose valuable data. The performance of `require('./minicore.node')` is 4x faster (40ms vs 120ms) than `require('@aws-sdk/client-s3')`, so I don't see any risk here.

PR #1213, opened 2025-12-03
Fix misleading connection success logs

vincentdchn:
### Description

The current logging in `connect()` and `connectAsync()` is misleading: it logs "connected successfully" immediately after initiating the request, not after receiving the response.

This caused confusion while debugging a connection issue with a wrong account name - the logs said "connected successfully after 3ms" but the connection was actually hanging until timeout.

**Changes:**

- Move success/failure logging into the actual callback where we know the real result
- Log "connection request initiated" at DEBUG level instead of false success at INFO level  
- Include the actual error message when connection fails

### Checklist

### DIFF
--- a/lib/connection/connection.js
+++ b/lib/connection/connection.js
@@ -241,7 +241,7 @@ function Connection(context) {
    */
   function connectCallback(self, callback, timer) {
     return function (err) {
-      const connectingDuration = timer ? timer.getDuration() : 'unknown';
+      const connectingDuration = timer.getDuration();
 
       if (err) {
         Logger.getInstance().error(


vincentdchn:

lib/connection/connection.js
```diff
@@ -235,12 +235,30 @@ function Connection(context) {
...
    * @param {Function} callback
+   * @param {ExecutionTimer} timer - timer started at the beginning of connect()
    *
    * @returns {function}
    */
-  function connectCallback(self, callback) {
+  function connectCallback(self, callback, timer) {
     return function (err) {
-      if (Parameters.getValue(Parameters.names.CLIENT_SESSION_KEEP_ALIVE)) {
+      const connectingDuration = timer ? timer.getDuration() : 'unknown';
```
sfc-gh-rsavenok: @vincentdchn thx for the PR! 

Question: It seems like the timer is always passed; can we remove this ternary? 

If there will be code regression, I'd rather have tests failing with "can not call getDuration of underfined" instead of the test passing with no indication that "unknown" is logged

vincentdchn: Good point, I've updated to use `timer.getDuration()` directly, thanks for the catch!

PR #1230, opened 2026-01-14
SNOW-267971 New request retry middleware

sfc-gh-rsavenok:
### Context: Fragmented Retry Logic

The codebase currently has multiple independent retry implementations scattered across different modules:

- **`lib/services/sf.js`** — Handles retries for session and authentication requests with its own retry loop
- **`lib/connection/statement.js`** — Implements retry logic for query execution with exponential backoff
- **`lib/services/large_result_set.js`** — Separate retry handling for result set fetching (includes HTTP 403 retry)

Each implementation has slightly different error handling, backoff strategies, and configuration options. More importantly, **no single implementation handles all the errors that should be retried** — some handle only HTTP errors, others handle network failures, and some miss specific status codes entirely.

### This PR: Unified Retry Middleware Foundation

This PR introduces a centralized retry middleware using Axios interceptors (`lib/http/axios_instance.ts`). The middleware:

- Uses exponential backoff with decorrelated jitter (reusing existing `Util.nextSleepTime`)
- Retries on transient HTTP errors (408, 429, 5xx) and network failures
- Respects client-initiated cancellations (no retry on `ERR_CANCELED`)
- Is opt-in via the `useSnowflakeRetryMiddleware` request option

**Starting point: Cancel Query requests** — This is the first endpoint to use the new middleware because it currently has **no retry logic at all**. This makes it a safe, isolated place to validate the new approach without affecting existing retry behavior elsewhere.

### What's Next

Subsequent PRs will migrate the existing retry implementations to use this middleware, consolidating retry logic into a single, well-tested location.

## Changes

- **`lib/http/axios_instance.ts`** — New Axios instance with retry interceptor
- **`lib/http/base.js`** — Switch to custom Axios instance, pass through `useSnowflakeRetryMiddleware` option
- **`lib/services/sf.js`** — Propagate `useSnowflakeRetryMiddleware` to HTTP layer
- **`lib/connection/statement.js`** — Enable retry middleware for cancel query requests
- **`lib/util.ts`** — Add `sleep()` utility function
- **`test/wiremockRunner.js`** — Fix variable substitution for non-string values

## Testing

- Unit tests for the retry interceptor covering success, retryable/non-retryable errors, max retries, and cancellation
- Integration tests with WireMock simulating HTTP errors and network faults for cancel query scenarios

### Checklist


wiremock/mappings/request_retries/cancel_query_http_fail.json
```diff
@@ -0,0 +1,95 @@
...
+      "request": {
+        "urlPathPattern": "/queries/v1/query-request*",
+        "method": "POST"
+      },
+      "response": {
+        "fixedDelayMilliseconds": 30000
+      }
+    },
+    {
+      "scenarioName": "Cancel query fails 3 times due to http error before success",
```
sfc-gh-dheyman, line 13: there's no success in the end, right? Scenario can be renamed

sfc-gh-dheyman (approved):

wiremock/mappings/request_retries/cancel_query_http_fail.json
sfc-gh-rsavenok: There is, last request is a success. So in total there will be 3 fails and 1 success. This ensures that actual backoff strategy is also working without throwing errors.  
  
For example, there's a bug in query execution (visible in 3d PR of this stack), that always retry any failure without any retry-related code. So doing 3 times ensures that we actually do proper retry+backoff

PR #1231, opened 2026-01-14
Fix OAuth authorization code scope parameter handling

sfc-gh-mborins:
Prevent scope parameter from being included in authorization URL when undefined. Previously, scope was always set in URL search params even when undefined, resulting in malformed OAuth URLs with 'scope=undefined'. Now only includes scope when explicitly provided, ensuring cleaner OAuth flows.

_**NOTE**_: this is applied only to the `oauth_authorization_code` code path sniffing for a specific string pattern rather than properly passing around "undefined"... this is because I'm unsure what other APIs are sharing the createScope function and didn't want to break anything.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### Description

### Checklist

- [ ] ~Extend the types in index.d.ts file (if necessary)~
- [ ] ~Extend the README / documentation and ensure is properly displayed (if necessary)~
- [ ] ~Provide JIRA issue id (if possible) or GitHub issue id in commit message~

### DIFF
--- a/lib/authentication/auth_oauth_authorization_code.js
+++ b/lib/authentication/auth_oauth_authorization_code.js
@@ -222,7 +222,7 @@ function AuthOauthAuthorizationCode(connectionConfig, httpClient) {
     authorizationUrl.searchParams.set('client_id', client.client_id);
     authorizationUrl.searchParams.set('redirect_uri', redirectUri);
     authorizationUrl.searchParams.set('response_type', 'code');
-    if (scope !== undefined) {
+    if (scope !== 'session:role:undefined') {
       authorizationUrl.searchParams.set('scope', scope);
     }
     authorizationUrl.searchParams.set('code_challenge', codeChallenge);


lib/authentication/auth_oauth_authorization_code.js
```diff
@@ -222,7 +222,9 @@ function AuthOauthAuthorizationCode(connectionConfig, httpClient) {
     authorizationUrl.searchParams.set('client_id', client.client_id);
     authorizationUrl.searchParams.set('redirect_uri', redirectUri);
     authorizationUrl.searchParams.set('response_type', 'code');
-    authorizationUrl.searchParams.set('scope', scope);
+    if (scope !== undefined) {
```
sfc-gh-huguo: maybe also skip empty string scope?

lib/authentication/auth_oauth_authorization_code.js
```diff
@@ -222,7 +222,9 @@ function AuthOauthAuthorizationCode(connectionConfig, httpClient) {
     authorizationUrl.searchParams.set('client_id', client.client_id);
     authorizationUrl.searchParams.set('redirect_uri', redirectUri);
     authorizationUrl.searchParams.set('response_type', 'code');
-    authorizationUrl.searchParams.set('scope', scope);
+    if (scope !== 'session:role:undefined') {
+      authorizationUrl.searchParams.set('scope', scope);
+    }
```
sfc-gh-mborins, line 227: This is kinda gross. I'm checking for the pattern where the role is undefined and explicitly ignoring when the string matches. I did this to avoid modifying the [`prepareScope` function ](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/2838e32010bb6eb71adb9ed7a21f8da7b2512383/lib/authentication/authentication_util.js#L88) as I was unsure if that would have a wider potential impact.

test/unit/authentication/authentication_test.js
```diff
@@ -449,6 +449,80 @@ describe('oauth authentication', function () {
...
+      authorizationUrl.searchParams.set('redirect_uri', redirectUri);
+      authorizationUrl.searchParams.set('response_type', 'code');
+      if (scope !== undefined) {
+        authorizationUrl.searchParams.set('scope', scope);
+      }
+      authorizationUrl.searchParams.set('code_challenge', codeChallenge);
+      authorizationUrl.searchParams.set('code_challenge_method', codeChallengeMethod);
+
+      assert.strictEqual(authorizationUrl.searchParams.has('scope'), true);
+      assert.strictEqual(authorizationUrl.searchParams.get('scope'), scope);
```
sfc-gh-rsavenok, line 474: I think this test is unnecessary. It doesn't test anything driver-specific; it just verifies that the URL API works as expected which we can trust without any test

test/unit/authentication/authentication_test.js
```diff
@@ -449,6 +449,80 @@ describe('oauth authentication', function () {
...
+      authorizationUrl.searchParams.set('redirect_uri', redirectUri);
+      authorizationUrl.searchParams.set('response_type', 'code');
+      if (scope !== undefined) {
+        authorizationUrl.searchParams.set('scope', scope);
+      }
+      authorizationUrl.searchParams.set('code_challenge', codeChallenge);
+      authorizationUrl.searchParams.set('code_challenge_method', codeChallengeMethod);
+
+      assert.strictEqual(authorizationUrl.searchParams.has('scope'), false);
+      assert.strictEqual(authorizationUrl.searchParams.get('scope'), null);
```
sfc-gh-rsavenok, line 497: Same here, if you don't call `searchParams.set`, you don't get the value :P

lib/authentication/auth_oauth_authorization_code.js
```diff
@@ -222,7 +222,9 @@ function AuthOauthAuthorizationCode(connectionConfig, httpClient) {
     authorizationUrl.searchParams.set('client_id', client.client_id);
     authorizationUrl.searchParams.set('redirect_uri', redirectUri);
     authorizationUrl.searchParams.set('response_type', 'code');
-    authorizationUrl.searchParams.set('scope', scope);
+    if (scope !== 'session:role:undefined') {
```
sfc-gh-rsavenok, line 225: I think the proper way to fix this is to adjust `prepareScope` in `authentication_utils.js`, but I'm not sure if our back-end actually supports an empty scope.

In the Go driver, for example, when the role is not set, it uses `session:role:`. Will sync with the team to better understand this edge case.

sfc-gh-rsavenok (commented):
Once the proper fix is made, let's have an integration test using wiremock that starts a connection without a role and checks that underlying request has the correct scope.

lib/authentication/auth_oauth_authorization_code.js
sfc-gh-rsavenok: @sfc-gh-mborins, what exactly are we trying to fix? Is there an edge where back-end rejects undefined, or do we just want to have a nicer URL? 

In JDBC driver it's `session:role:null`, in .NET it will throw an error, in Python it will do `scope=` (empty string)

sfc-gh-mborins: Python will not send scope in the URL is it doesn't exist. The backend will fail if `session:role:undefined`

PR #1262, opened 2026-02-06
SNOW-3031461 Do not send oauth scope when both role and oauthScope is empty 

sfc-gh-rsavenok:
### Description

Currently, we send `session:role:undefined`. This results in Snowflake OAuth to throw an error:
image

The same skip logic is implemented in the Python driver. Even though Jenkins is red, its auth tests are passing.

### Checklist

### DIFF
--- a/lib/authentication/authentication_util.js
+++ b/lib/authentication/authentication_util.js
@@ -85,7 +85,7 @@ function getTokenUrl(options) {
   return new URL(tokenUrl);
 }
 
-async function prepareScope(options) {
+function prepareScope(options) {
   const oauthScope = options.getOauthScope();
   const role = options.getRole();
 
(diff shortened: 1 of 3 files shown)


wiremock/mappings/oauth/client_credentials/successful_flow_without_scope.json
```diff
@@ -0,0 +1,776 @@
...
+            "masterValidityInSeconds": 14400,
+            "displayUserName": "OAUTH_TEST_AUTH_CODE",
+            "serverVersion": "8.48.0 b2024121104444034239f05",
+            "firstLogin": false,
+            "remMeToken": null,
+            "remMeValidityInSeconds": 0,
+            "healthCheckInterval": 45,
+            "newClientForUpgrade": "3.12.3",
+            "sessionId": 1172562260498,
+            "parameters": [
```
sfc-gh-rkowalski, line 84: do we actually need parameters? It bloats the mapping file

sfc-gh-rkowalski (approved):
Looks good

wiremock/mappings/oauth/client_credentials/successful_flow_without_scope.json
sfc-gh-rsavenok: UP to dissolution during UD migration. I personally prefer clean Wiremocks without unneeded stuff in them, but I know that some people just prefer recording entire backend communication instead

I've added a todo in the code to clean up this during UD migration.

PR #1268, opened 2026-02-10
Update minimum node version

sfc-gh-rsavenok:
### Description

There was dead code that throws an error when the node was lower than 6.x. Our code won't run on anything lower than 12, and we officially support Node 18+ since the 2.0.0 release.

So removed the in-code error and moved this responsibility to the `package.json` -> `engines` field. This throws a warning during `npm install`.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Fixed OAuth crashing when using bundlers (snowflakedb/snowflake-connector-nodejs#1266)
 - Added Node 18+ to `engines`, which is our minimum officially supported version since the 2.x release (snowflakedb/snowflake-connector-nodejs#1268)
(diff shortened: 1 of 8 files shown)


lib/util.ts
```diff
@@ -178,56 +178,6 @@ export const string = {
...
   },
-
-  /**
-   * Compares two version numbers of the form 'a.b.c' where a, b and c are
-   * numbers (e.g. '1.0.12'). If one or both inputs are invalid versions, the
-   * function will return NaN, otherwise, it will return -1 if the first
-   * version is smaller, 1 if the first version is bigger, and 0 if the two
-   * versions are equal.
-   */
-  compareVersions: function (version1: string, version2: string) {
```
sfc-gh-dheyman, line 189: after update, we won't be needing this anymore?

sfc-gh-rsavenok: Yes

sfc-gh-pbulawa (approved):

PR #1287, opened 2026-02-21
fix: add bn.js override to deduplicate v4/v5 versions

cjnoname:
## Problem

`asn1.js@5.4.1` declares `bn.js@^4.0.0` while the project directly depends on `bn.js@^5.2.1`, causing two versions (4.12.3 and 5.2.3) to be installed simultaneously. This leads to `instanceof` mismatches when `bn.js` v5 instances created in `cert_util.js` are passed into `asn1.js` internals that use `bn.js` v4.

## Solution

Add `"bn.js": "^5.2.1"` to the npm `overrides` field to force all transitive dependencies to use a single version.

## Safety verification

- Compared all prototype methods between bn.js v4.12.3 and v5.2.3 — **fully backward compatible**
- All APIs used by `asn1.js` (`toArray`, `copy`, `negative`, `length`, `toString`) behave identically in both versions
- No public API was removed in v5; only internal methods were renamed (`strip` → `_strip`) and new methods added

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,8 @@
     "asn1.js": "^5.4.1"
   },
   "overrides": {
-    "semver": "^7.5.2"
+    "semver": "^7.5.2",
+    "bn.js": "^5.2.1"
   },
   "repository": {
     "type": "git",


sfc-gh-rsavenok:
Hi @cjnoname!

According to npm docs, this won't work for published packages:
> Overrides are only considered in the root package.json file for a project. Overrides in installed dependencies (including [workspaces](https://docs.npmjs.com/cli/v11/using-npm/workspaces)) are not considered in dependency tree resolution. Published packages may dictate their resolutions by pinning dependencies or using an [npm-shrinkwrap.json](https://docs.npmjs.com/cli/v11/configuring-npm/npm-shrinkwrap-json) file.

A project that installed `snowflake-sdk` would still resolve 2 instances of bn.js.

Looking at the source code of `asn1.js`, it doesn't use `instance of BN`; it always calls `bn.toArray()` and creates a buffer from that array. So the version mismatch shouldn't cause any bugs.

Seems like asn1.js decodes buffers, so there might be a way to completely drop bn.js dependency.

Issue #1286, opened 2026-02-21
SNOW-3132803: Security: Transitive dependency bn.js affected by infinite loop vulnerability (GHSA-378v-28hj-76wf)


cjnoname:
## Vulnerability Details

The `snowflake-sdk` package has a transitive dependency on a vulnerable version of `bn.js` through the following chain:

```
snowflake-sdk > asn1.js > bn.js (4.12.3)
```

### Advisory

| Field | Value |
|-------|-------|
| **Package** | bn.js |
| **Vulnerability** | Infinite loop |
| **Severity** | Moderate |
| **Affected Versions** | < 5.2.3 |
| **Patched Versions** | >= 5.2.3 |
| **Advisory** | [GHSA-378v-28hj-76wf](https://github.com/advisories/GHSA-378v-28hj-76wf) |

### Impact

All downstream consumers of `snowflake-sdk` inherit this vulnerability. The infinite loop in `bn.js` can potentially be exploited to cause denial-of-service conditions.

### Steps to Reproduce

1. Install `snowflake-sdk@2.3.4`
2. Run `npm audit` (or `pnpm audit` / `yarn audit`)
3. The `bn.js` infinite loop vulnerability is reported through the `asn1.js` dependency

spdaley:
Per https://github.com/indutny/bn.js/issues/316, 4.12.3 should have the fix as well. Looks like the advisories have not been updated with that information yet though.

sfc-gh-dszmolka:
hi @cjnoname , thanks for reporting this. looks like a patched version is already pulled in by the dependencies:
```
# npm list bn.js
snowflake-sdk@2.3.4 /snowflake-connector-nodejs
+-- asn1.js@5.4.1
| `-- bn.js@4.12.3 <<<
`-- bn.js@5.2.3 <<<
```

the direct dep allows for the fixed version to be installed (`"bn.js": "^5.2.1",`) but we can bump of course.

sfc-gh-dszmolka:
PR is now merged and is part of the next release. Will keep this thread posted about that.

PR #1294, opened 2026-02-23
SNOW-3132803: retire bn.js

sfc-gh-rsavenok:
### Description

We don't really need the `bn.js` dependency. `asn1.js` exports `asn1.bigint`, which is a reference to the underlying `bn.js` dependency.

This fixes concerns in PR https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1287 and vulenerability in https://github.com/snowflakedb/snowflake-connector-nodejs/issues/1286

### Checklist

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -11,7 +11,7 @@
     "strict": true,
     "skipLibCheck": true,
     "paths": {
-      "asn1.js-rfc5280": ["./lib/types/asn1.js-rfc5280.d.ts"]
+      "asn1.js": ["./lib/types/asn1.js.d.ts"]
     }
   }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -18,12 +18,12 @@
     "@smithy/protocol-http": "^5.3.8",
     "@smithy/signature-v4": "^5.3.8",
     "@techteamer/ocsp": "1.0.1",
+    "asn1.js": "^5.0.0",
```
sfc-gh-rsavenok, line 21: This is not a new dependency, asn1.js-rfc* already requires it. It's just to be explicit that our code directly calls it now.

PR #1299, opened 2026-02-25
NO-SNOW Remove samples folder

sfc-gh-rsavenok:
### Description

There's no reason to keep outdated performance tests that are executed locally in the repo. If we ever need similar tests in the future, llms can generate them locally in a minute.

Also, the json parser test is wrong - function parsing shows better numbers, because each iteration uses the same JSON, resulting in an evaluated result cache. In real life, the JSONs will differ, and the performance will be 10x slower

### Checklist


sfc-gh-rsavenok:
> Shouldn't we wait with samples removal until we have equivalent setup in docs.snowflake.com?

The existing samples don't provide new usage examples that aren't already in the docs. It's a folder used internally to do performance tests on json parsers and query execution

sfc-gh-dheyman (approved):

PR #1300, opened 2026-02-25
SNOW-3161103 JSON.parse as default variant column parser

sfc-gh-rsavenok:
### Description

The test that was performed in https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/samples/jsonParserComparison.js 8 years ago is incorrect. `new Function()` is faster only when you construct the same JSON string, in a real-world scenario where each row contains different JSON, it's 10 times slower.

Test on 30k medium-sized JSONs:
* `JSON.parse` - 600ms
* `new Function` - 6s

### Checklist

### DIFF
--- a/test/integration/testDataType.js
+++ b/test/integration/testDataType.js
@@ -35,8 +35,6 @@ describe('Test DataType', function () {
     'insert into testNumber values (12345678901234567890123456789012345678)';
   const insertRegularSizedNumber = 'insert into testNumber values (100000001)';
   const insertVariantJSON =
-    "insert into testVariant select parse_json('{a : 1 , b :[1 , 2 , 3, -Infinity, undefined], c : {a : 1}}')";
-  const insertVariantJSONForCustomParser =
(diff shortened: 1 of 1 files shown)


test/integration/testDataType.js
```diff
@@ -35,8 +35,6 @@ describe('Test DataType', function () {
     'insert into testNumber values (12345678901234567890123456789012345678)';
   const insertRegularSizedNumber = 'insert into testNumber values (100000001)';
   const insertVariantJSON =
-    "insert into testVariant select parse_json('{a : 1 , b :[1 , 2 , 3, -Infinity, undefined], c : {a : 1}}')";
```
sfc-gh-rsavenok, line 38: According to RFC, `Infinity` and `undefined` are not allowed in JSONs, so we shouldn't allow them in our JSON parsing either. Any custom JSON parser or any other language driver will fail to parse such JSONs

PR #1304, opened 2026-02-25
feat: export normalizeConnectionOptions for snake_case TOML keys

sfc-gh-mborins:
## Summary

- Export `normalizeConnectionOptions()` as a public utility for converting snake_case connection keys (from TOML config files) to the camelCase equivalents expected by `ConnectionConfig`
- Uses a generic `snakeToCamel` regex with 4 explicit overrides for acronym keys (MFA, GCP, URL) and 2 key aliases (`user`->`username`, `private_key_file`->`privateKeyPath`)
- Strips `undefined` values defensively; camelCase keys take precedence when both forms are provided
- Called internally by `loadConnectionConfiguration()` so `createConnection(null)` works automatically
- `ConnectionConfig` itself stays strictly camelCase -- normalization is the caller's responsibility

## Motivation

TOML config files and the Python connector use snake_case keys, but the JS SDK only accepts camelCase. This forces every consumer that parses TOML to maintain manual key mapping layers. This change lets consumers call `snowflakeSdk.normalizeConnectionOptions(rawTomlConfig)` and pass the result directly to `createConnection()`.

## Test plan

- 12 unit tests covering: basic conversion, key aliases, OAuth keys, mixed casing, camelCase precedence, undefined stripping, acronym edge cases, workload identity keys, single-word passthrough, validateDefaultParameters compatibility, and null/undefined input guard
- All 148 existing + new tests passing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,7 +37,7 @@ timestamps {
       'Test': {
         stage('Test') {
           def params = [
-            string(name: 'svn_revision', value: 'bptp-stable'),
+            string(name: 'svn_revision', value: 'temptest-deployed'),
             string(name: 'branch', value: 'main'),
             string(name: 'client_git_commit', value: scmInfo.GIT_COMMIT),
             string(name: 'client_git_branch', value: scmInfo.GIT_BRANCH),
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -59,13 +59,9 @@ for name in "${!TARGET_TEST_IMAGES[@]}"; do
     echo "[INFO] Testing $DRIVER_NAME on $name"
     docker pull  "${TARGET_TEST_IMAGES[$name]}"
     docker run \
-        --net $NETWORK_NAME \
+        --network=host \
         -v $(cd $THIS_DIR/.. && pwd):/mnt/host \
         -v $WORKSPACE:/mnt/workspace \
-        --add-host snowflake.reg.local:$GATEWAY_HOST \
-        --add-host testaccount.reg.snowflakecomputing.com:$GATEWAY_HOST \
-        --add-host snowflake.reg.snowflakecomputing.com:$GATEWAY_HOST \
-        --add-host externalaccount.reg.local.snowflakecomputing.com:$GATEWAY_HOST \
         -e LOCAL_USER_ID=$(id -u $USER) \
         -e LOCAL_USER_NAME=$USER \
         -e USERID \


lib/connection/connection_config.js
```diff
@@ -196,6 +250,8 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
   // check for invalid options
   Errors.checkArgumentValid(Util.isObject(options), ErrorCodes.ERR_CONN_CREATE_INVALID_OPTIONS);
 
+  options = normalizeConnectionOptions(options);
```
sfc-gh-rsavenok: Are you ok with moving this normalization from connection_config to toml parsing logic? The reason behind this is that ConnectionConfig is being typed (see lib/connection/types.ts). So at some point, it won't accept snake_case or aliases.

index.d.ts
```diff
@@ -751,6 +751,8 @@ declare module 'snowflake-sdk' {
    */
   export function createConnection(options: ConnectionOptions): Connection;
 
+  export function normalizeConnectionOptions(options: Record<string, unknown>): ConnectionOptions;
```
sfc-gh-rsavenok, line 681: Please add JSDoc above this method explaining what it does and maybe a usecase where you want to reuse toml from other drivers or parse toml manually

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@
 - Removed `bn.js` dependency (snowflakedb/snowflake-connector-nodejs#1294)
 - Added a warning when converting query result to JavaScript numbers with precision loss (snowflakedb/snowflake-connector-nodejs#1295, snowflakedb/snowflake-connector-nodejs#1296)
 - Changed default `jsonColumnVariantParser` to `JSON.parse` (snowflakedb/snowflake-connector-nodejs#1300)
-- Fixed a bug where parsing the JSON media type failed when it included an optional parameter from Microsoft Identity Platform v2.0 tokens, failing OAuth Client Credentials flow (snowflakedb/snowflake-connector-nodejs#1301)
+- Exported `normalizeConnectionOptions()` to convert snake_case TOML keys to camelCase for `ConnectionConfig`, with key aliases and acronym overrides (snowflakedb/snowflake-connector-nodejs#1304)
```
sfc-gh-rsavenok: This PR also added snake_case support to the existing toml reading.

Maybe add 2 separate entries?
- Added snake_case and support for connections.toml
- Added new public method `normalizeConnectionOptions()` allowing to .. description of normalization logic .. to be used in `createConnection()`

lib/connection/connection_config.js
```diff
@@ -85,6 +85,60 @@ const DEFAULT_PARAMS = [
   'rowStreamHighWaterMark',
 ];
 
+const KEY_ALIASES = {
```
sfc-gh-rsavenok: Please move this logic to a new .ts file in the connection folder called `normalize_connection_options.ts`

lib/connection/connection_config.js
```diff
@@ -85,6 +85,60 @@ const DEFAULT_PARAMS = [
...
+  disable_saml_url_check: 'disableSamlURLCheck',
+  crl_allow_certificates_without_crl_url: 'crlAllowCertificatesWithoutCrlURL',
+};
+
+function snakeToCamel(key) {
+  return key.replace(/_([a-z])/g, (_, char) => char.toUpperCase());
+}
+
+function normalizeConnectionOptions(options) {
+  if (!options || typeof options !== 'object') {
```
sfc-gh-rsavenok: Once the file becomes .ts, this won't be needed as you'll have strict arg check

lib/connection/connection_config.js
```diff
@@ -85,6 +85,60 @@ const DEFAULT_PARAMS = [
...
+    return options;
+  }
+
+  const normalized = {};
+  for (const [key, value] of Object.entries(options)) {
+    if (value === undefined) {
+      continue;
+    }
+
+    if (KEY_ALIASES[key] !== undefined) {
```
sfc-gh-rsavenok: `if (KEY_ALIASES[key]) {}`

lib/connection/connection_config.js
```diff
@@ -85,6 +85,60 @@ const DEFAULT_PARAMS = [
...
+      if (normalized[KEY_ALIASES[key]] === undefined) {
+        normalized[KEY_ALIASES[key]] = value;
+      }
+      continue;
+    }
+
+    if (SNAKE_TO_CAMEL_OVERRIDES[key] !== undefined) {
```
sfc-gh-rsavenok: `if (SNAKE_TO_CAMEL_OVERRIDES[key]) {}`

lib/connection/connection_config.js
```diff
@@ -85,6 +85,60 @@ const DEFAULT_PARAMS = [
...
+    if (key === key.toLowerCase() && key.includes('_')) {
+      const camelKey = snakeToCamel(key);
+      if (normalized[camelKey] === undefined) {
+        normalized[camelKey] = value;
+      }
+    } else {
+      normalized[key] = normalized[key] !== undefined ? normalized[key] : value;
+    }
+  }
+  return normalized;
```
sfc-gh-rsavenok: We don't have a good way to filter out keys that aren't valid connection options atm, so for TypeScript, just do `as Partial<WIP_ConnectionOptions>` for now.

test/unit/connection/connection_config_test.js
```diff
@@ -1720,3 +1721,148 @@ describe('ConnectionConfig: basic', function () {
     });
   });
 });
+
+describe('normalizeConnectionOptions', function () {
```
sfc-gh-rsavenok: Please move this to a separate test file `test/unit/cinnection/normalize_connection_options_test.ts`

test/unit/connection/connection_config_test.js
```diff
@@ -1720,3 +1721,148 @@ describe('ConnectionConfig: basic', function () {
...
+  it('should convert snake_case keys to camelCase', function () {
+    const normalized = normalizeConnectionOptions({
+      account: 'testaccount',
+      username: 'testuser',
+      password: 'testpass',
+      private_key_path: '/path/to/key.p8',
+    });
+    const config = new ConnectionConfig(normalized);
```
sfc-gh-rsavenok: With TypeScript, the normalization would return `Partial<WIP_ConnectionOptions>`, so the unit tests don't need to create connection config:
```
assert.strictEqual(normalized.privateKeyPath, ...)
```

test/unit/connection/connection_config_test.js
```diff
@@ -1720,3 +1721,148 @@ describe('ConnectionConfig: basic', function () {
...
+    assert.strictEqual(config.getPrivateKeyPath(), '/path/to/key.p8');
+  });
+
+  it('should map user to username', function () {
```
sfc-gh-rsavenok: Check `.cursor/rules/testing.mdx`, we have a rule to name tests with "should"

lib/configuration/connection_configuration.js
sfc-gh-rsavenok: Since this is now done in normalization, fixUserKey can be removed as dead code

test/unit/connection/connection_config_test.js
sfc-gh-rsavenok: Additionally, if we won't create connection object and trust TypeScript, each test case won't need to add account,username,password just to pass

sfc-gh-rsavenok: Once we simplified the tests, all of them could be written as an array of name, input, and output. 

See example in test/unit/util_test.js describe('isPrivateKey'

lib/connection/connection_config.js
sfc-gh-mborins: Done — removed the null/typeof guard. TS strict args handle this now.

lib/connection/connection_config.js
sfc-gh-mborins: Done.

lib/connection/connection_config.js
sfc-gh-mborins: Done — moved to `lib/connection/normalize_connection_options.ts`. ConnectionConfig stays strictly camelCase.

lib/connection/connection_config.js
sfc-gh-mborins: Done — return type is `Partial<WIP_ConnectionOptions>` with an `as` cast.

test/unit/connection/connection_config_test.js
sfc-gh-mborins: Done — created `test/unit/connection/normalize_connection_options_test.ts` and removed the old tests from `connection_config_test.js`.

CHANGELOG.md
sfc-gh-mborins: Done — split into 2 entries: one for snake_case support in `connections.toml`, one for the exported `normalizeConnectionOptions()` utility.

lib/connection/connection_config.js
sfc-gh-mborins: Done — created `lib/connection/normalize_connection_options.ts` and removed the code from `connection_config.js`.

lib/configuration/connection_configuration.js
sfc-gh-mborins: Done — removed `fixUserKey()`. The `user→username` alias is handled by `normalizeConnectionOptions` via `KEY_ALIASES`.

index.d.ts
sfc-gh-mborins: Done — added JSDoc explaining snake_case→camelCase conversion, key aliases, and the manual TOML parsing use case.

lib/connection/connection_config.js
sfc-gh-mborins: Done.

test/unit/connection/connection_config_test.js
sfc-gh-mborins: Done — all tests use the table-driven `[{name, input, expected}].forEach()` pattern.

test/unit/connection/connection_config_test.js
sfc-gh-mborins: Done — removed "should" prefix from all test names.

test/unit/connection/connection_config_test.js
sfc-gh-mborins: Done — tests only pass the keys relevant to each case, no boilerplate account/username/password.

sfc-gh-mborins: Done — tests now assert directly on the normalized output with `deepStrictEqual`, no ConnectionConfig construction.

index.d.ts
```diff
@@ -108,21 +108,6 @@ declare module 'snowflake-sdk' {
      */
     region?: string;
 
-    /**
```
sfc-gh-rsavenok, line 111: I will make another PR to move the rest of the connection options to the `types.ts`

sfc-gh-rkowalski (approved):

Issue #1308, opened 2026-02-26
SNOW-3170447: isByPassProxy does not handle dot-prefixed NO_PROXY entries (.domain.com)


jaysobel:
## Bug Report

The `isByPassProxy` function in `lib/proxy_util.js` does not correctly handle the standard `NO_PROXY` convention where a leading dot (`.domain.com`) means "match any subdomain of `domain.com`."

## Affected Code

```javascript
// lib/proxy_util.js
exports.isByPassProxy = function (proxy, destination) {
  if (proxy && proxy.noProxy) {
    const bypassList = proxy.noProxy.split('|');
    return bypassList.find((noProxy) => {
      let host = noProxy.trim();
      host = host.replace('*', '.*?');
      const noProxyRegex = new RegExp(`^${host}`);
      return noProxyRegex.test(destination);
    });
  }
};
```

## Problem

The regex is `^`-anchored. When `noProxy` contains `.snowflakecomputing.com`, the constructed regex is `^\.snowflakecomputing\.com`. This never matches a destination like `myaccount.us-east-1.snowflakecomputing.com` because the destination string does not start with a literal dot.

The leading-dot convention (`.domain.com` = "any subdomain of domain.com") is standard across HTTP client libraries:

- **curl**: [docs](https://curl.se/libcurl/c/CURLOPT_NOPROXY.html) — "A host name that matches without a leading dot will also match against a host name with a leading dot prepended"
- **Python `requests`** / `urllib3`: leading-dot entries match any subdomain
- **Go `net/http`**: same behavior
- **Node.js `global-agent`** and **`proxy-agent`**: same behavior

## Reproduction

```javascript
const { isByPassProxy } = require('./lib/proxy_util');

const proxy = { noProxy: '.snowflakecomputing.com' };
const destination = 'myaccount.us-east-1.snowflakecomputing.com';

console.log(isByPassProxy(proxy, destination));
// => undefined (no match — should match)
```

## Workaround

Converting `.domain.com` to `*.domain.com` before passing to the connector works, because `*` is replaced with `.*?`, producing regex `^.*?\.snowflakecomputing\.com` which matches correctly.

```javascript
// Workaround: rewrite dot-prefix to wildcard-prefix
noProxy = noProxy.replace(/\.snowflakecomputing\.com/, '*.snowflakecomputing.com');
```

This should not be necessary — the SDK should handle it.

## Suggested Fix

When a `noProxy` entry starts with `.`, either:

[...]

PR #1309, opened 2026-02-27
SNOW-3170447 isByPassProxy fix  and tests

sfc-gh-dszmolka:
### Description

This is for https://github.com/snowflakedb/snowflake-connector-nodejs/issues/1308. `isByPassProxy` did not match `.domain.com` for `NO_PROXY` entries despite it's entirely valid wildcard reference next to the `*.domain.com` one.

Addressing with a simpler fix, which:
* applies a `*` for a host entry which starts with a `.` , so the subsequent replace could produce a working regex

Furthermore, `isByPassProxy` did not seem to have any tests; now added some:
* matching `*.domain.com` wildcard patterns
* match exact host entries
* matching `.domain.com` wildcard pattern
* and, when destination is a RegExp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fixed `disableSamlUrlCheck` typing to use correct casing: `disableSamlURLCheck` (snowflakedb/snowflake-connector-nodejs#1304)
 - Added snake_case key support when loading `connections.toml` via `createConnection()` with no arguments (snowflakedb/snowflake-connector-nodejs#1304)
 - Exported `normalizeConnectionOptions()` utility to convert snake_case connection keys to camelCase, with key aliases and acronym overrides (snowflakedb/snowflake-connector-nodejs#1304)
+- Fixed `getDefaultCacheDir()` crashing in environments where no user home directory is configured by falling back to `os.tmpdir()` (snowflakedb/snowflake-connector-nodejs#1312)
 - Fixed a bug where host specified in `NO_PROXY` in the `.domain.com` wildcard format were not correctly matching the destination host (snowflakedb/snowflake-connector-nodejs#1309)
 
 ## 2.3.4
--- a/lib/disk_cache.ts
+++ b/lib/disk_cache.ts
@@ -9,13 +9,20 @@ import os from 'os';
 //
 // We should refactor the code so every place is using utils from this file
 export function getDefaultCacheDir() {
+  let rootDir: string;
+  try {
+    rootDir = os.homedir();
+  } catch {
+    rootDir = os.tmpdir(); // fallback to TMP if user home doesn't exist
+  }
+
   switch (process.platform) {
     case 'win32':
-      return path.join(os.homedir(), 'AppData', 'Local', 'Snowflake', 'Caches');
+      return path.join(rootDir, 'AppData', 'Local', 'Snowflake', 'Caches');
     case 'linux':
-      return path.join(os.homedir(), '.cache', 'snowflake');
+      return path.join(rootDir, '.cache', 'snowflake');
     case 'darwin':
-      return path.join(os.homedir(), 'Library');
+      return path.join(rootDir, 'Library');
     default:
       throw new Error(`Unsupported platform: ${process.platform}`);
   }
--- a/lib/minicore/minicore.ts
+++ b/lib/minicore/minicore.ts
@@ -30,7 +30,10 @@ if (process.env.SNOWFLAKE_DISABLE_MINICORE) {
 
     minicoreStatus.version = minicoreModule.sfCoreFullVersion();
   } catch (error: unknown) {
-    minicoreStatus.errorType = 'Failed to load binary';
+    minicoreStatus.errorType = isBinaryIgnoredByBundlers(error)
+      ? 'Binary is missing from the bundle'
+      : 'Failed to load binary';
+
     if (error instanceof Error) {
       minicoreStatus.errorDetails = error;
     }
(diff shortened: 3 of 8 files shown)


test/unit/proxy_util_test.js
```diff
@@ -448,3 +448,161 @@ describe('Proxy Util for Azure', function () {
...
+
+describe('isByPassProxy', function () {
+  function makeProxy(noProxy) {
+    return { host: 'my.pro.xy', port: 8080, noProxy };
+  }
+
+  describe('wildcard patterns', function () {
+    [
+      {
+        name: 'wildcard matches single subdomain',
```
sfc-gh-rsavenok: Since the describe has "wildcard", I think we can remove wildcard word from the case names

test/unit/proxy_util_test.js
```diff
@@ -448,3 +448,161 @@ describe('Proxy Util for Azure', function () {
...
+        shouldMatch: true,
+      },
+      {
+        name: 'wildcard matches multi-level subdomain',
+        noProxy: '*.snowflakecomputing.com',
+        destination: 'myaccount.us-east-1.snowflakecomputing.com',
+        shouldMatch: true,
+      },
+      {
+        name: 'wildcard matches amazonaws.com',
```
sfc-gh-rsavenok: Why do we need a test for Amazon? Looking at the code, there is no special logic for AWS. So this is the same test as multi-level domain

test/unit/proxy_util_test.js
```diff
@@ -448,3 +448,161 @@ describe('Proxy Util for Azure', function () {
...
+        name: 'wildcard in multi-entry list matches correct entry',
+        noProxy: '*.amazonaws.com|*.snowflakecomputing.com|localhost',
+        destination: 'myaccount.snowflakecomputing.com',
+        shouldMatch: true,
+      },
+    ].forEach(({ name, noProxy, destination, shouldMatch }) => {
+      it(name, function () {
+        const result = ProxyUtil.isByPassProxy(makeProxy(noProxy), destination);
+        if (shouldMatch) {
+          assert.ok(result, `Expected "${destination}" to match noProxy "${noProxy}"`);
```
sfc-gh-rsavenok: This is a very good `assertion.ok` block; you clearly see both the noProxy and destination values.

Would suggest creating a single array instead of 4 describe blocks and keep this message

test/unit/proxy_util_test.js
sfc-gh-dszmolka: true, removing

lib/proxy_util.js
sfc-gh-boler, line 358: We can't just turn something that has a meaning of glob pattern into a regexp. The string must first be escaped, to avoid unexpected matches.

test/unit/proxy_util_test.js
```diff
@@ -448,3 +448,111 @@ describe('Proxy Util for Azure', function () {
...
+    },
+    {
+      name: 'wildcard pattern - does not match unrelated domain',
+      noProxy: '*.snowflakecomputing.com',
+      destination: 'example.com',
+      shouldMatch: false,
+    },
+    {
+      name: 'wildcard pattern - in multi-entry list matches correct entry',
+      noProxy: '*.amazonaws.com|*.snowflakecomputing.com|localhost',
```
sfc-gh-boler, line 478: I was looking at the NodeJS spec, and the syntax is to separate patterns by commas not pipes: https://nodejs.org/api/http.html#no_proxy-format

sfc-gh-dszmolka: we already using pipes in tests which is not part of the pr, see https://github.com/snowflakedb/snowflake-connector-nodejs/blob/v2.3.4/test/unit/proxy_util_test.js#L254

but changed to commas

lib/proxy_util.js
sfc-gh-dszmolka: since `host = host.replace(/\*/g, '.*?');` is an optional step and not related to fixing the original issue, removed

test/unit/proxy_util_test.js
sfc-gh-rsavenok: We can not change to commas; it would be a breaking change.

The `|` is already in our documentation:
```
For multiple hosts, separate the hostnames with a pipe symbol (|).
```

I agree that we need to support comas, so it has to be both separators and the docs need to be updated

lib/proxy_util.js
sfc-gh-rsavenok: This is a good catch!

The following code would pass because `.` is a special regex character
```
new RegExp('google.pl').test('googlempl') 
```

We already have util for this in file_util.js called `globToRegex`. Let's move it to utils.ts and use it in 2 places. I think it's ok that it also supports `?`, in future releases we will just document that the no proxy support full glob range

sfc-gh-rsavenok: UPD: we might create breaking changes because of globToRegex as no proxy doesn't matches end of string.

So lets do it safe and move `char.replace(/[.+^${}()|[\]\\]/g, '\\$&')` as escapeRegex to utils.ts and use it here + globToRegex function

test/unit/proxy_util_test.js
sfc-gh-dszmolka: okay, for now can we use the documented method which is also used in other tests, which is pipes?

lib/proxy_util.js
sfc-gh-dszmolka: can it be done in a separate PR, since it's unrelated to fixing the issue which this one targets to fix?

sfc-gh-boler (approved):
lgtm

PR #1310, opened 2026-02-27
SNOW-3029730 Collect libc details in telemetry

sfc-gh-rsavenok:
### Description

* Collect `LIBC_FAMILY` + `LIBC_VERSION` in the logic request telemetry. 
* Replaced `isMusl()` in the minicore with`getLibcDetails()` 

### Checklist

### DIFF
--- a/lib/minicore/minicore.ts
+++ b/lib/minicore/minicore.ts
@@ -30,7 +30,10 @@ if (process.env.SNOWFLAKE_DISABLE_MINICORE) {
 
     minicoreStatus.version = minicoreModule.sfCoreFullVersion();
   } catch (error: unknown) {
-    minicoreStatus.errorType = 'Failed to load binary';
+    minicoreStatus.errorType = isBinaryIgnoredByBundlers(error)
+      ? 'Binary is missing from the bundle'
+      : 'Failed to load binary';
+
     if (error instanceof Error) {
       minicoreStatus.errorDetails = error;
     }
@@ -68,4 +71,22 @@ export function getMinicoreStatus() {
     isMinicoreStatusLogged = true;
   }
   return minicoreStatus;
+}
+
+function isBinaryIgnoredByBundlers(error: unknown) {
+  if (error instanceof Error) {
+    const message = error.message;
+    return (
+      // eval('require') was ignored by bundlers
+      message.startsWith('Cannot find module') ||
+      // bundled code running in ESM environment
+      message === 'require is not defined' ||
+      // Windows: DLL mapped to a network path that the bundled app can't reach
+      message === 'LoadLibrary failed: The network name cannot be found.' ||
+      // Windows: DLL not found
+      message === 'LoadLibrary failed: The specified module could not be found.'
+    );
+  } else {
+    return false;
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Upcoming Release
 
+- Added ability to skip token file permission checks using `SF_SKIP_TOKEN_FILE_PERMISSIONS_VERIFICATION` env variable (snowflakedb/snowflake-connector-nodejs#1314)
 - Fixed OAuth crashing when using bundlers (snowflakedb/snowflake-connector-nodejs#1266)
 - Added Node 18+ to `engines`, which is our minimum officially supported version since the 2.x release (snowflakedb/snowflake-connector-nodejs#1268)
 - Added `PLATFORM` field to login-request telemetry (snowflakedb/snowflake-connector-nodejs#1269)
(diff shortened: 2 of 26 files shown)


lib/minicore/minicore.ts
```diff
@@ -46,7 +46,7 @@ export function getBinaryName() {
   const suffix: string[] = [platform, arch];
 
   if (platform === 'linux') {
-    suffix.push(isMusl() ? 'musl' : 'gnu');
+    suffix.push(getLibcDetails().family === 'musl' ? 'musl' : 'gnu');
```
sfc-gh-boler, line 52: Overall I think it's fine. But I'd like to have 3 values: musl, gnu (assuming that's glibc), and "other". There are a few other libc implementations out there so we shouldn't mask them with "gnu".

sfc-gh-rsavenok: This is a minicore binary loader. It will try to load GNU `sf_mini_core_0.0.1.linux-x64-gnu.node` and fail with compatibility error.

The actual telemetry is in `sf.js`, it its not musl/glibc - the value would be null  there

sfc-gh-boler: Ah, of course!

lib/telemetry/libc_details.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  cachedResult = result;
+  return cachedResult;
+}
+
+function detectFromFilesystem(): LibcDetails | null {
+  try {
+    const content = readFileSync(LDD_PATH, 'utf-8');
```
sfc-gh-pfus, line 53: I think we can have both glibc and musl in LDD_PATH and it shows nothing - just that such a library is available in the ldd path. I don't think we should put such a detection function as a first one.

lib/telemetry/libc_details.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (isMusl) {
+      return { family: MUSL, version: null };
+    }
+  }
+  return null;
+}
+
+function detectFromCommand(): LibcDetails | null {
+  try {
+    const out = execSync('getconf GNU_LIBC_VERSION 2>&1 || true; ldd --version 2>&1 || true', {
```
sfc-gh-pfus, line 104: Same.

lib/telemetry/libc_details.ts
```diff
@@ -0,0 +1,120 @@
...
+    return cachedResult;
+  }
+
+  if (process.platform !== 'linux') {
+    cachedResult = { family: null, version: null };
+    return cachedResult;
+  }
+
+  const result: LibcDetails = { family: null, version: null };
+  for (const detectFn of [detectFromFilesystem, detectFromReport, detectFromCommand]) {
```
sfc-gh-pfus, line 34: Take a look at go implementation - it may be worth checking what libc version this particular process is using.

lib/telemetry/libc_details.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (result.family && result.version) {
```
sfc-gh-pfus, line 35: Don't you prefer to put such a condition at the end of the loop?

lib/telemetry/libc_details.ts
sfc-gh-rsavenok: There's no attached version to the node process. `detect-libc`, which my implementation is based on, reads musl/glibc from the process, but that source doesn't include the version :/ 

So, unless I will see in the telemetry that we're using the wrong libc, I don't want to include it

lib/telemetry/libc_details.ts
sfc-gh-rsavenok: No :P

lib/telemetry/libc_details.ts
sfc-gh-rsavenok: Both napi.rs and detect-libc use the same detection order:
1) (detect-libc only) musl/libc from mounted libc reference to the process (searches string `/ld-musl-` or `/ld-linux-`). This doesn't have a version, so I skipped it
2) ldd reads from the filesystem value. It's a shell script that often contains a static string for glibc or musl. (validated this manually)
3) Node.js process details (won't contain musl version)
4) Executes the child process with the ldd version

As the code comment says, they are sorted from fastest to slowest:
```
/*
 * Detection uses three strategies in order of increasing cost:
 * 1. Filesystem: read /usr/bin/ldd content — a simple file read, cheapest option
 * 2. Process report: process.report.getReport() — no filesystem access but heavier than a file read
 * 3. Child process: exec `getconf` and `ldd --version` — last resort, spawns a shell process
 *
 * Inspired by detect-libc (https://github.com/lovell/detect-libc)
 */
```

lib/telemetry/libc_details.ts
sfc-gh-rsavenok: Don't get this "same" comment, it's a last detector in the list

PR #1312, opened 2026-03-02
SNOW-2198224 Fall back to os.tmpdir() in getDefaultCacheDir() when home directory is unavailable

sfc-gh-rsavenok:
### Description

`getDefaultCacheDir()` in `lib/disk_cache.ts` calls `os.homedir()` which throws `ERR_SYSTEM_ERROR` in environments where no user home directory is configured. This wraps the call in a try/catch that falls back to `os.tmpdir()`, matching the existing pattern in `global_config.js`.

Adds unit tests for `getDefaultCacheDir()` covering all three platform paths, the unsupported platform error, and the tmpdir fallback.

### Checklist

### DIFF
--- a/lib/minicore/minicore.ts
+++ b/lib/minicore/minicore.ts
@@ -30,7 +30,10 @@ if (process.env.SNOWFLAKE_DISABLE_MINICORE) {
 
     minicoreStatus.version = minicoreModule.sfCoreFullVersion();
   } catch (error: unknown) {
-    minicoreStatus.errorType = 'Failed to load binary';
+    minicoreStatus.errorType = isBinaryIgnoredByBundlers(error)
+      ? 'Binary is missing from the bundle'
+      : 'Failed to load binary';
+
     if (error instanceof Error) {
       minicoreStatus.errorDetails = error;
     }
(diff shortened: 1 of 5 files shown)


lib/disk_cache.ts
```diff
@@ -9,13 +9,20 @@ import os from 'os';
...
 // We should refactor the code so every place is using utils from this file
 export function getDefaultCacheDir() {
+  let rootDir: string;
+  try {
+    rootDir = os.homedir();
+  } catch {
+    rootDir = os.tmpdir(); // fallback to TMP if user home doesn't exist
+  }
+
   switch (process.platform) {
```
sfc-gh-boler, line 19: shouldn't those paths be used only if we have homedir, and if we get to use tmp, then choose something simpler?

/tmp/.cache/snowflake or /tmp/Library are going to look odd

sfc-gh-rsavenok: Maybe in UD, but that's how it was initially implemented for ocsp cache and I don't want to touch it further :P

sfc-gh-boler: :(

sfc-gh-boler (approved):

PR #1315, opened 2026-03-03
SNOW-3170447 bugfixes for noProxy

sfc-gh-rsavenok:
### Description

Follow up after https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1309, as additional bugs were found

- Fix 2 bugs in `noProxy` / `NO_PROXY` pattern matching in `isByPassProxy`:
  - `.` in host patterns (e.g. `snowflakecomputing.com`) was treated as a regex wildcard instead of a literal dot, causing `snowflakecomputingccom` to incorrectly match
  - Partial `noProxy` entries (e.g. `snowflake`) incorrectly matched longer destinations (e.g. `snowflakecomputing.com`) because the regex was not anchored at the end
  - Added unit tests covering 2 bugs
- Extract `escapeRegex` helper into `lib/util.ts`, used by both `proxy_util.js` and `file_util.js`
- Improve `noProxy` JSDoc in `index.d.ts` with structured description and example

### Checklist

### DIFF
--- a/lib/proxy_util.js
+++ b/lib/proxy_util.js
@@ -347,14 +347,20 @@ exports.isByPassProxy = function (proxy, destination) {
   if (proxy && proxy.noProxy) {
     const bypassList = proxy.noProxy.split('|');
     return bypassList.find((noProxy) => {
-      let host = noProxy.trim();
+      const host = noProxy.trim();
       if (destination instanceof RegExp) {
         return destination.test(host);
       }
-      if (host.startsWith('.')) {
-        host = '*' + host;
+      let regexPattern = '';
+      for (let i = 0; i < host.length; i++) {
+        const char = host[i];
+        if (char === '*' || (char === '.' && i === 0)) {
+          regexPattern += '.*';
+        } else {
+          regexPattern += Util.escapeRegex(char);
+        }
       }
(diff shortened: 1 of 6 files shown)


lib/proxy_util.js
```diff
@@ -347,15 +347,20 @@ exports.isByPassProxy = function (proxy, destination) {
...
-      if (host.startsWith('.')) {
-        host = '*' + host;
-      }
-      host = host.replace('*', '.*?');
+      const host = noProxy.trim();
       if (destination instanceof RegExp) {
         return destination.test(host);
       }
-      const noProxyRegex = new RegExp(`^${host}`);
+      let regexPattern = '';
```
sfc-gh-dheyman, line 354: sidenote: do we have a test for a potential REDOS attack? There was similar thing in JDBC recently: https://github.com/snowflakedb/snowflake-jdbc/commit/5fb0a8a318a2ed87f4022a1f56e742424ba94052

sfc-gh-rsavenok: "does not treat special regex characters as regex" covers this a bit (checks that the regex is actually escaped). I think this test is good enough until the UD replaces the codebase

PR #1320, opened 2026-03-05
SNOW-3184292 Restore CRL tests

sfc-gh-rsavenok:
### Description

Restore CRL integration tests that were previously skipped because of failures

- Replace the platform-specific proxy test with a custom CONNECT tunnel proxy that works on all platforms. WireMock can't be used here because its `--enable-browser-proxying` always performs TLS interception (MITM), replacing the server certificate with a generated one that lacks extensions like CRL Distribution Points
- Fix the axios spy target to use `axiosInstance` instead of `axios`
- Move proxy connection options from `index.d.ts` to `WIP_ConnectionOptions`

### Checklist


index.d.ts
```diff
@@ -118,43 +118,6 @@ declare module 'snowflake-sdk' {
      */
     openExternalBrowserCallback?: (url: string) => void;
 
-    /**
```
sfc-gh-boler, line 121: It's just my weak typescript skill, but doesn't this change the public interface? I know it moves to types.ts, but that's a different type?

lib/agent/crl_validator/crl_fetcher.ts
```diff
@@ -12,8 +12,17 @@ import {
 } from './crl_cache';
 
 export const PENDING_FETCH_REQUESTS = new Map<string, Promise<ASN1.CertificateListDecoded>>();
+
+let memoryCacheCleanerInterval: NodeJS.Timeout | undefined;
+let diskCacheCleanerInterval: NodeJS.Timeout | undefined;
 let crlCacheCleanerCreated = false;
 
+export function resetCrlCacheCleaner() {
+  clearInterval(memoryCacheCleanerInterval);
```
sfc-gh-boler, line 21: reset these to their default values to avoid surprises

index.d.ts
sfc-gh-rsavenok: At the very beginning of this code block, there's `export type ConnectionOptions = WIP_ConnectionOptions & {`. So it merges options from 2 different files.

I didn't migrate all of them at once as I noticed poor descriptions for some of them and processing them in batches when I touch related code

lib/agent/crl_validator/crl_fetcher.ts
sfc-gh-rsavenok: There won't be any surprises. Once the clearInterval is called, you can't do anything with the `NodeJS.Timeout` that was assigned to it.

Since these variables are top-level, GC will never clean them, but memory-wise, it doesn't really matter whether we have an undefined or a destroyed timeout object there; the difference is insignificant for this use case.

PR #1321, opened 2026-03-05
SNOW-2395146 Improve CRL validation security

sfc-gh-rsavenok:
### Description

Harden CRL validation with two security improvements:

1. **Download size limit**: Added `crlDownloadMaxSize` global config option (default 20 MB) that enforces a `maxContentLength` on CRL download HTTP responses via axios. This prevents an attacker or misconfigured CA from serving arbitrarily large responses and consuming excessive memory.

2. **ADVISORY mode logging level**: Changed the CRL ADVISORY mode failure log from `debug` to `warn` level, so that allowed-but-failed CRL checks are visible without enabling debug logging.

Additionally, moved `PENDING_FETCH_REQUESTS.delete` to finally block to handle crl download errors properly

### Checklist


lib/global_config_typed.ts
```diff
@@ -45,6 +52,7 @@ export interface GlobalConfigOptionsTyped {
 
 export const GLOBAL_CONFIG_DEFAULTS: GlobalConfigOptionsTyped = {
   crlDownloadTimeout: 10000,
+  crlDownloadMaxSize: 200,
```
sfc-gh-boler: Not sure what's the convention in the node ecosystem, but could we put the unit in the name?

sfc-gh-boler: 200MB is really large for a CRL. Most systems will not handle 50MB. For example, our infamous .net issue was caused by just 16MB large CRL list. OpenSSL needed a lot of memory to parse it. Recently Microsoft had an incident which would have produced a CRL approx. 90MB large; they were managing the incident response very carefully to do that.

So, for this limit to have a practical effect, we should aim for something lower. 10MB is pretty unusual most of the time.

lib/global_config_typed.ts
```diff
@@ -30,6 +30,13 @@ export interface GlobalConfigOptionsTyped {
    */
   crlCacheValidityTime: number;
 
+  /**
+   * Maximum allowed size for a CRL download response, in megabytes.
```
sfc-gh-boler: Let's describe what happens if the declared response is larger than this.

lib/agent/crl_fetcher.ts
```diff
@@ -63,27 +63,31 @@ export async function getCrl(
...
-    const { data } = await axios.get(url, {
-      timeout: GlobalConfigTyped.getValue('crlDownloadTimeout'),
-      responseType: 'arraybuffer',
-    });
+    try {
+      logDebug(`Downloading CRL`);
+      const { data } = await axios.get(url, {
+        timeout: GlobalConfigTyped.getValue('crlDownloadTimeout'),
+        responseType: 'arraybuffer',
+        maxContentLength: GlobalConfigTyped.getValue('crlDownloadMaxSize') * 1024 * 1024,
```
sfc-gh-boler: How does this behave when response does not provide `content-length` header?

sfc-gh-rsavenok: It doesn't use the header, it calculates the actual response size that was returned

lib/global_config_typed.ts
sfc-gh-rsavenok: There's no node-specific convention like Ms/MB suffix. Since all other similar properties in the driver are named without a suffix, I don't want to be inconsistent. 

As a developer, when you hover on a property name or use autocomplete, you'll get a full description on which unit to use from the attached JSDoc

lib/global_config_typed.ts
sfc-gh-rsavenok: Isn't it obvious that if "maximum allowed" is used, it means that the code won't let a higher response size?

sfc-gh-boler: It is far from obvious what "won't let higher response size" means in practical terms. Does it download only that much data and then does validation? Does it straight out allows or denies connection? And how does it interact with the advisory mode (which I'm not sure if we have in this drive, but may add it in the future).

lib/global_config_typed.ts
sfc-gh-boler: Looks like our Go driver already calls it `CrlDownloadMaxSize`. Let's keep it that way for consistency.

sfc-gh-jkasten: Microsoft has still not resolved that incident.  They are still getting hammered for it 9 months later, FWIW.
https://bugzilla.mozilla.org/show_bug.cgi?id=1965612

Although Microsoft Root Program guidelines state that CAs should keep CRLs under 10MB... the guidelines aren't followed and last I checked, they could not be followed. (Considering the certificate population underneath a particular intermediate and divided by the number of CRL shards the intermediate is currently supporting. CAs are supposed to revoke all affected certificates during an incident. The math doesn't work out. All of this is exacerbated by CCADB restricting their CRL field size.).

For instance, snowflake.com's DigiCert certificate's CRL currently sits at 8.54MB and that is when DigiCert hasn't even been forced to revoke their certificates (Heartbleed, public incident, etc). 
http://crl3.digicert.com/DigiCertGlobalG2TLSRSASHA2562020CA1-1.crl
This is a steady state for them. 

I would personally bump it up a bit to avoid unnecessary availability outages for those in enforcement mode.

sfc-gh-boler: Didn't realize that Microsoft was still on it. Looking at the revocation plan published here https://bugzilla.mozilla.org/attachment.cgi?id=9494215, it assumed revocation of 800k certs per week. That's cumulative over the lifetime of those certs. WIth ~30 bytes per CRL entry, we are at ~24MB weekly growth of the list. 

If we were to take this kind of events into account when setting the default limit, it wouldn't be effective and applications would just OOM in low memory environments. It's better to fail the connection to Snowflake than crashing the whole application.

With guidelines of some root programs recommending keeping the lists under 10MB, setting our default limit to 20MB seems reasonable and leaves some room for unexpected events like the Digicert mass-revocation from 2024.

What do you think, James?

sfc-gh-jkasten: I think 20MB is probably fine. We are essentially trading an availability risk being OOM which would affect ADVISORY mode, with an unavailability risk of those with ENFORCEMENT mode. Those in ENFORCEMENT mode already run an availability risk and they can crank up their limit if the want to.

Here is the rest of my reasoning so that it can be critqued.

**Pros**
* Certificate lifetimes are being forced down which means the steady state of CRL sizes should shrink naturally. Everyone will still be victim to large mass revocation, but the CRL will shrink more quickly afterwards. DigiCert won't start with an 8.54MB CRL
* Microsoft's root program requirements give us some cover when this blows up. Here is the [text](https://learn.microsoft.com/en-us/security/trusted-root/program-requirements#:~:text=All%20issuing%20CA,should%20be%20%3C10MB.) in case you are interested.

**Cons**
* Let's Encrypt and GTS take sharding seriously -- though they are overprovisioned (steady state is small, but mass revocation would turn ugly.
  * Last time I checked, LE was running at like an 80MB CRL if they had a problem with the "http-01" challenge.
* Amazon Trust Services appears to be unprepared. They don't do CRL sharding (only intermediate sharding) and they currently have 40 million active certs issued under one of their intermediates. [[Censys Query](https://platform.censys.io/search?q=%28cert.parsed.issuer_dn%3A+%22Amazon+RSA+2048+M04%22%29+and+cert.labels+%3D+%22unexpired%22)]. 
* Pressure is on CAs to expand their CRL sizes and revoke for every minor infraction.
  * Evidence: Entrust's distrust -- one of many; recent mass revocation testing / planning requirements from root programs; see how angry the community is still at Microsoft for their failure -- they are certainly on thin ice at this point; I have plenty of stories here.

sfc-gh-boler: I agree with your risk analysis, specifically users who enable ENFORCEMENT hopefully are informed enough to adjust the max allowed CRL size to balance resource requirements and the reliability risk.

sfc-gh-snoonan: Can we be consistent in the units across drivers? Go is using bytes; you're using m(ega?|ibi?)bytes

sfc-gh-rsavenok: Changed to bytes!

lib/global_config_typed.ts
sfc-gh-rsavenok: Updated the description

PR #1325, opened 2026-03-06
SNOW-2333028 Add RSASSA-PSS signature verification for CRL validation

sfc-gh-rsavenok:
### Description

Implement RSASSA-PSS signature verification for CRL validation (SNOW-2333028).

Some Certificate Authorities sign CRLs using the RSASSA-PSS algorithm (OID `1.2.840.113549.1.1.10`) instead of the traditional PKCS#1 v1.5 algorithms. Previously, encountering such a CRL would throw an "Unsupported signature algorithm" error. This change adds full support by:

- **New `rsassa_pss_parser.ts` module** — defines an ASN.1 schema for the `RSASSA-PSS-params` structure (RFC 4055) and parses the hash algorithm and salt length from the signature algorithm parameters. Supports SHA-256, SHA-384, and SHA-512 (SHA-1 is explicitly rejected). When salt length is omitted, it defaults to the hash digest length per RFC convention.
- **`crl_signature_verifier.ts`** — adds a `pssVerifier` function that uses `crypto.verify()` with `RSA_PKCS1_PSS_PADDING` and the parsed PSS parameters, and registers it in `CRL_SIGNATURE_VERIFIERS` for the RSASSA-PSS OID.
- **`oids.ts`** — adds hash algorithm OIDs (SHA-1, SHA-256, SHA-384, SHA-512) and the RSASSA-PSS OID.
- **`asn1.js.d.ts`** — extends TypeScript declarations for the `asn1.js` library with `define()`, `ASN1Builder`, and `ASN1Entity` types to support the custom ASN.1 schema definition.
- **Test utilities (`test_utils.ts`)** — adds `createPSSAlgorithmIdentifier()` helper and updates `createTestCRL`/`createTestCertificate` to propagate `signatureAlgorithmOid` properly and handle PSS signing.
- **New unit tests for `parseRSASSAPSSParams`** — covers SHA-256/384/512 parsing, salt length defaults/overrides, rejection of SHA-1 and missing parameters, and malformed input.
- **New integration-style test with real OpenSSL-generated fixtures** — validates the full pipeline against a real RSASSA-PSS SHA-256 CRL + CA pair generated by the included `generate.sh` script, and verifies rejection with an unrelated key.

### Checklist

### DIFF
--- a/ci/test_revocation.sh
+++ b/ci/test_revocation.sh
@@ -33,9 +33,9 @@ REVOCATION_BRANCH="${REVOCATION_BRANCH:-main}"
 
 rm -rf "$REVOCATION_DIR"
 if [ -n "$GITHUB_USER" ] && [ -n "$GITHUB_TOKEN" ]; then
-    git clone --depth 1 --branch "$REVOCATION_BRANCH" "https://${GITHUB_USER}:${GITHUB_TOKEN}@github.com/snowflakedb/revocation-validation.git" "$REVOCATION_DIR"
+    git clone --depth 1 --branch "$REVOCATION_BRANCH" "https://${GITHUB_USER}:${GITHUB_TOKEN}@github.com/snowflake-eng/revocation-validation.git" "$REVOCATION_DIR"
 else
-    git clone --depth 1 --branch "$REVOCATION_BRANCH" "https://github.com/snowflakedb/revocation-validation.git" "$REVOCATION_DIR"
+    git clone --depth 1 --branch "$REVOCATION_BRANCH" "https://github.com/snowflake-eng/revocation-validation.git" "$REVOCATION_DIR"
 fi
 
 cd "$REVOCATION_DIR"
--- a/index.d.ts
+++ b/index.d.ts
@@ -119,43 +119,6 @@ declare module 'snowflake-sdk' {
     openExternalBrowserCallback?: (url: string) => void;
 
     /**
-     * Specifies a list of hosts that the driver should connect to directly, bypassing the proxy server.
-     *
-     * - Use a pipe symbol (`|`) to separate multiple hosts.
-     * - Use `*` as a wildcard (e.g., `*sub.amazonaws.com`).
-     * - A leading dot (e.g., `.amazonaws.com`) matches any subdomain.
-     *
-     * @example
-     * noProxy: ".amazonaws.com|*sub.my_company.com"
-     */
-    noProxy?: string;
-
-    /**
-     * Specifies the hostname of an authenticated proxy server.
-     */
-    proxyHost?: string;
-
-    /**
-     * Specifies the username used to connect to an authenticated proxy server.
-     */
-    proxyUser?: string;
-
-    /**
-     * Specifies the password for the user specified by proxyUser.
-     */
-    proxyPassword?: string;
-
-    /**
-     * Specifies the port of an authenticated proxy server.
-     */
-    proxyPort?: number;
-
-    /**
-     * Specifies the protocol used to connect to the authenticated proxy server. Use this property to specify the HTTP protocol: http or https.
-     */
-    proxyProtocol?: string;
-
-    /**
      * Specifies whether the OCSP request is also sent to the proxy specified.
      */
     useConnectionConfigProxyForOCSP?: boolean;
--- a/lib/agent/crl_validator/crl_fetcher.ts
+++ b/lib/agent/crl_validator/crl_fetcher.ts
@@ -12,7 +12,16 @@ import {
 } from './crl_cache';
 
 export const PENDING_FETCH_REQUESTS = new Map<string, Promise<rfc5280.CertificateListDecoded>>();
+
+let memoryCacheCleanerInterval: NodeJS.Timeout | undefined;
+let diskCacheCleanerInterval: NodeJS.Timeout | undefined;
 let crlCacheCleanerCreated = false;
+
+export function resetCrlCacheCleaner() {
+  clearInterval(memoryCacheCleanerInterval);
+  clearInterval(diskCacheCleanerInterval);
+  crlCacheCleanerCreated = false;
+}
 
 export async function getCrl(
   url: string,
@@ -28,11 +37,11 @@ export async function getCrl(
     const oneHour = 1000 * 60 * 60;
 
     logDebug('Starting periodic memory cache cleaner');
-    setInterval(clearExpiredCrlFromMemoryCache, oneHour).unref();
+    memoryCacheCleanerInterval = setInterval(clearExpiredCrlFromMemoryCache, oneHour).unref();
 
     logDebug('Starting periodic disk cache cleaner');
     clearExpiredCrlFromDiskCache();
-    setInterval(clearExpiredCrlFromDiskCache, oneHour).unref();
+    diskCacheCleanerInterval = setInterval(clearExpiredCrlFromDiskCache, oneHour).unref();
   }
 
   const pendingFetchRequest = PENDING_FETCH_REQUESTS.get(url);
--- a/lib/connection/types.ts
+++ b/lib/connection/types.ts
@@ -264,6 +264,43 @@ export interface WIP_ConnectionOptions {
    * @default false
    */
   validateDefaultParameters?: boolean;
+
+  /**
+   * Specifies a list of hosts that the driver should connect to directly, bypassing the proxy server.
+   *
+   * - Use a pipe symbol (`|`) to separate multiple hosts.
+   * - Use `*` as a wildcard (e.g., `*sub.amazonaws.com`).
+   * - A leading dot (e.g., `.amazonaws.com`) matches any subdomain.
+   *
+   * @example
+   * noProxy: ".amazonaws.com|*sub.my_company.com"
+   */
+  noProxy?: string;
+
+  /**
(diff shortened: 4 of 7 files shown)


lib/agent/crl_validator/crl_signature_verifier.ts
```diff
@@ -1,28 +1,40 @@
...
-// Implement RSASSA-PSS signature verification
-// https://snowflakecomputing.atlassian.net/browse/SNOW-2333028
+function pssVerifier(crl: rfc5280.CertificateListDecoded, issuerPublicKey: string) {
+  const pssParams = parseRSASSAPSSParams(crl.signatureAlgorithm.parameters);
+  const tbsEncoded = rfc5280.TBSCertList.encode(crl.tbsCertList, 'der');
+  return crypto.verify(
+    pssParams.hashAlgorithm,
+    tbsEncoded,
+    {
+      key: issuerPublicKey,
```
sfc-gh-boler, line 22: We using the default mgf1 hash, but echnically the signature may use a different one than used for the message hash (AI says it's unlikley, but no point leaving that gap open). We should get mgf1 from the signature.

lib/agent/crl_validator/crl_signature_verifier.ts
```diff
@@ -1,28 +1,40 @@
...
 import rfc5280 from 'asn1.js-rfc5280';
 import { ALGORITHM_OID } from './oids';
+import { parseRSASSAPSSParams } from './rsassa_pss_parser';
 
 type SignatureVerifier = (crl: rfc5280.CertificateListDecoded, issuerPublicKey: string) => boolean;
 
 function digestVerifier(digestAlg: string): SignatureVerifier {
   return (crl, issuerPublicKey) => {
     const tbsEncoded = rfc5280.TBSCertList.encode(crl.tbsCertList, 'der');
-    const verify = crypto.createVerify(digestAlg);
```
sfc-gh-boler, line 10: why the change from createVerify() to verify()?

test/unit/agent/crl_validator/test_utils.ts
```diff
@@ -110,13 +137,16 @@ export function createTestCertificate(
...
+  const signatureAlgorithmOid = options.signatureAlgorithmOid ?? ALGORITHM_OID.ECDSA_WITH_SHA256;
   const keyPair = options.keyPair ?? createCertificateKeyPair(signatureAlgorithmOid);
 
-  const signatureAlgorithm = {
-    algorithm: signatureAlgorithmOid.split('.').map(Number),
-    parameters: Buffer.from([0x05, 0x00]),
-  };
+  const signatureAlgorithm =
+    signatureAlgorithmOid === ALGORITHM_OID.RSASSA_PSS
+      ? createPSSAlgorithmIdentifier({ hashOid: ALGORITHM_OID.SHA256 })
```
sfc-gh-boler: It would be good to run tests against different algos, since they will have different digest sizes.

lib/agent/crl_validator/rsassa_pss_parser.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (!hashAlgorithm) {
+    throw new Error(
+      `RSASSA-PSS unsupported hashAlgorithm OID: ${hashOid} (only SHA-256, SHA-384, and SHA-512 are supported)`,
+    );
+  }
+
+  return {
+    hashAlgorithm,
+    saltLength:
+      params.saltLength != null ? params.saltLength.toNumber() : HASH_DIGEST_LENGTH[hashAlgorithm],
```
sfc-gh-boler, line 55: let's ensure we don't get a negative number here

test/unit/agent/crl_validator/test_utils.ts
```diff
@@ -110,13 +137,16 @@ export function createTestCertificate(
   const notBefore = options.notBefore ?? '2026-01-01T00:00:00Z';
   const notAfter = options.notAfter ?? '2026-12-31T00:00:00Z';
   const subject = options.subject ?? createCertificateNameField();
-  const signatureAlgorithmOid = options.signatureAlgorithmOid ?? ALGORITHM_OID.SHA256_WITH_RSA;
+  const signatureAlgorithmOid = options.signatureAlgorithmOid ?? ALGORITHM_OID.ECDSA_WITH_SHA256;
```
sfc-gh-boler, line 141: I don't follow why this default is changing. Can you add a bit of explanation here?

lib/agent/crl_validator/crl_signature_verifier.ts
sfc-gh-rsavenok: I parsed this this initially in `parseRSASSAPSSParams` implementation, but Node `crypto` module won't be able to verify it.

> Node.js automatically uses the same hash algorithm for both the signature and MGF1—there's no way to specify a different MGF1 hash. So if the RSASSA-PSS parameters specify a different hash for MGF1 than for the signature, you cannot verify it correctly with Node's built-in crypto.

lib/agent/crl_validator/crl_signature_verifier.ts
sfc-gh-rsavenok: It's the same behavior, just 1 line of code instead of 3. When I coded first crl implementation, I didn't know that you could do 1-liner :P

test/unit/agent/crl_validator/test_utils.ts
sfc-gh-rsavenok: Don't you think that's overkill? 

We have unit tests that verify that the salt length is either the default (for 3 algos) or a custom value (a higher number or 0). And since every verification goes through the same `crypto.verify` with the same parsed params flow, we expect the same behavior

Don't have a strong opinion on this, if you insist, I'll add more tests

lib/agent/crl_validator/rsassa_pss_parser.ts
sfc-gh-rsavenok: Cursor even wrote a test for this, but it won't pass, the asn1.js won't let you have a negative number. The parsed value is always a positive number, so the verification will fail because we got the wrong number

test/unit/agent/crl_validator/test_utils.ts
sfc-gh-rsavenok: There was this line in the tests: `crypto.getFips() ? 2048 : 512, // faster test runs` it was actually wrong, and PSS would have made this `if` even worse, so I switched the default algorithm to ECDSA as the node generates and signs content in tests faster with them. 

By "faster", I mean the number of unit tests that take 150-200ms to complete because of slow signing

lib/agent/crl_validator/rsassa_pss_parser.ts
sfc-gh-rsavenok: Additional context:
```
asn1.js does NOT handle negative integers correctly for ASN.1 INTEGER decoding.

In ASN.1 DER encoding, integers use two's complement representation. A negative integer has its most significant bit set to 1. For example:

-1 is encoded as 0xFF
-128 is encoded as 0x80
-20 is encoded as 0xEC
However, _decodeInt simply does new bignum(raw) where raw is the Buffer of bytes. The bn.js constructor (_initArray) treats the byte array as an unsigned big-endian number — it never inspects the sign bit. It just reads the raw bytes as a positive magnitude.

So for a negative ASN.1 integer:

0xFF (which should be -1) would be decoded as 255
0x80 (which should be -128) would be decoded as 128
0xFEC0 (which should be some negative number) would be decoded as a large positive number
```

sfc-gh-rsavenok: Also, to properly unit test that the validation fail, I would have to generate another fixture using openssl script as unit tests generate them with asn1 which will fail encoding negative number as well

lib/agent/crl_validator/crl_signature_verifier.ts
sfc-gh-rsavenok: I guess it makes sense to add a comment in `RSASSAPSSParamsEntity` why `maskGenAlgorithm` field isn't parsed

sfc-gh-boler: Yes, a comment would be handy.

lib/agent/crl_validator/rsassa_pss_parser.ts
sfc-gh-boler: Graphite concurs! But still, this is a bug in asn1.js. If they ever fix it, we won't notice and will be exposed. Just like the JDBC AWS dependency on default params to a crypto module. Let's apply defense in depth and crash or error out if we ever get a negative value.

We can skip a tests, since we can't easily construct an invalid value.

test/unit/agent/crl_validator/test_utils.ts
sfc-gh-boler: It seems straightforward to add the extra coverage, so let's do it.

lib/agent/crl_validator/rsassa_pss_parser.ts
sfc-gh-rsavenok: Worst-case scenario:
* asn1 fixes the issue
* somebody builds a CRl that is not valid according to RFC
* crypto.verify is called with a negative saltLength and returns false (that validation has failed)

PR #1326, opened 2026-03-08
Fix TypeError when query response omits rowset in Result.createChunks

sfc-gh-rgaunker:
### Description

  Fixes a crash in result processing when query responses omit data.rowset.

  Result.createChunks() accessed rowset.length without guarding for undefined, which can throw:
  TypeError: undefined is not an object (evaluating 'rowset.length').

### Root Cause

  In lib/connection/result/result.js, createChunks() normalized chunkCfgs but not rowset, then dereferenced rowset.length.

### Changes

  - Added defensive normalization in createChunks():
      - rowset = rowset || [];
  - Added regression test in test/unit/connection/result/result_test.js:
      - does not throw when rowset is missing
      - removes response.data.rowset, sets chunks = [], and verifies new Result(...) does not throw.

### Impact

  - Prevents runtime crashes when rowset is absent in response payloads.
  - Backward-compatible for existing result parsing paths.

### Validation

  - npm run test:single -- test/unit/connection/result/result_test.js
      - 20 passing

### Checklist

- [ ] Create tests which fail without the change (if possible)
- [ ] Make all tests (unit and integration) pass (`npm run test:unit` and `npm run test:integration`)
- [ ] Extend the types in index.d.ts file (if necessary)
- [ ] Extend the README / documentation and ensure is properly displayed (if necessary)
- [ ] Provide JIRA issue id (if possible) or GitHub issue id in commit message

PR #1329, opened 2026-03-12
NO-SNOW Fix unhandled promise rejection

sfc-gh-rsavenok:
### Description

Related to https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1326

When statement.js fails to process a query response, the failure isn't piped to the top level as an err value in a callback. There are many cases of such issues in statement.js, but focusing only on one where we don't have control - malformed server response.

Added:
- try-catch around `onStatementRequestSucc` to prevent unhandled promise rejections when the server returns malformed responses (e.g., `rowset: null`)
- `unhandledRejection` listener in Mocha config to convert leaked rejections into test failures instead of process crashes. This resulted in the discovery of 2 places where we have unhandled rejections:
  - `file_util_test.js` - missing `await` on `assert.rejects` / `assert.doesNotReject`, and updated to use `validateNoExtraPermissionsForOthers` (the prev function doesnt exist)
  - `auth_oauth_authorization_code.js` - race condition where `receiveData` promise could reject while `codeProvider` was still being awaited, leaving the rejection unhandled. Fixed by not awaiting `codeProvider` (fire-and-forget pattern), since the server response determines success/failure.
- WireMock-based integration test that verifies the driver properly handles malformed query responses

### Checklist


test/integration/wiremock/testStatementReceivingMalformedResponse.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  after(async () => {
+    await wiremock.global.shutdown();
+  });
+
+  it('returns runtime error', async function () {
+    const connection = testUtil.createConnection(connectionConfig);
+    await testUtil.connectAsync(connection);
+    try {
+      await testUtil.executeCmdAsync(connection, 'SELECT 1');
```
sfc-gh-pbulawa, line 47:
```suggestion
      await testUtil.executeCmdAsync(connection, 'SELECT 1');
      assert.fail('No expected error thrown');
```

lib/connection/statement.js
```diff
@@ -643,7 +643,12 @@ function BaseStatement(statementOptions, context, services, connectionConfig) {
 
     // if there was no error, call the success function
     if (!err) {
-      await context.onStatementRequestSucc(body);
+      try {
+        await context.onStatementRequestSucc(body);
+      } catch (unexpectedError) {
+        Logger.getInstance().error('Unexpected error in onStatementRequestSucc:', unexpectedError);
```
sfc-gh-pbulawa, line 649: Shouldn't we update the `err = unexpectedError` here? In line 668, we emit the event, but if the operation fails, the error is silently omitted.

sfc-gh-rsavenok: That error is used for some other context that I don't fully understand. 

If there's an error, the stream always ends with `close(context.resultError)`. This is confirmed by the test written.

test/integration/wiremock/testStatementReceivingMalformedResponse.ts
sfc-gh-rsavenok: Here, I want to confirm that the error matches my expectations.

It could be done with `assert.rejects` as well, but rejects returns a very poor error: "expected rejection, but got none".

My implementation gives a nice error "expected error ... but got ....".

When we'll migrate to UD, this won't matter as I plan to use Vitest instead of mocha which has much better assertions :)

PR #1332, opened 2026-03-13
SNOW-3184242 Build minicore for glibc 2.18

sfc-gh-rsavenok:
### Description

Build Linux GNU minicore binaries targeting **glibc 2.18** to support older Linux distributions (e.g., RHEL 7, Amazon Linux 2).

The napi-rs CLI doesn't support glibc version suffixes in target triples (e.g., `aarch64-unknown-linux-gnu.2.18`), because its artifact copy step expects the output directory to match the full target string. To work around this, GNU targets now build directly via `cargo zigbuild` with the `.2.18` suffix, and the resulting `.so` is manually copied/renamed to the expected `.node` path.

Non-GNU targets (macOS, Windows, musl) continue to use the existing `napi build` path.

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -109,16 +109,6 @@ declare module 'snowflake-sdk' {
     region?: string;
 
     /**
-     * Specifies the timeout, in milliseconds, for browser activities related to SSO authentication. The default value is 120000 (milliseconds).
-     */
-    browserActionTimeout?: number;
-
-    /**
-     * Customize implementation for opening the browser window used for SSO authentication. By default, the npm `open` package is used.
-     */
-    openExternalBrowserCallback?: (url: string) => void;
-
-    /**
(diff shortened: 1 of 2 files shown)


ci/build_minicore.js
```diff
@@ -39,27 +39,68 @@ const BUILD_TARGETS = [
...
+     * copy/rename the .so to .node.
+     */
+    if (target.includes('gnu')) {
+      execSync(
+        [
+          'cargo',
+          'zigbuild',
+          '--release',
+          '--manifest-path',
+          'lib/minicore/rust_minicore/Cargo.toml',
```
sfc-gh-rsavenok, line 61: NOTE: I don't really care that we repeat values like `lib/minicore/rust_minicore/Cargo.toml` or `sf_mini_core_0.0.1` in multiple places as in UD we probably won't use cross-platform builds

PR #1335, opened 2026-03-16
SNOW-3018695 SPCS_TOKEN support

sfc-gh-rsavenok:
### Description

Snowpark Container Services (SPCS) is introducing a new `SPCS_TOKEN` field in the login-request to help identify service requests. This is an opaque service identifier token (distinct from the SPCS OAuth token) that cannot be used independently to log in to Snowflake — it only serves to identify the originating service.

The connector reads this token from `/snowflake/session/spcs_token` (overridable via `SF_SPCS_TOKEN_PATH` env var) and attaches it as `SPCS_TOKEN` in the login-request payload. When the token file is absent (i.e. not running in an SPCS container), the field is omitted.

**Changes:**
- **`lib/authentication/spcs_token.ts`** — new module with `getSpcsToken()` that reads the token file, trims whitespace, and returns `null` on failure.
- **`lib/services/sf.js`** — calls `getSpcsToken()` during login and attaches the token to the request body when present.
- **`test/unit/authentication/spcs_token_test.ts`** — unit tests for `getSpcsToken()` covering missing file, default path, custom path via env var, and whitespace trimming.
- **`test/integration/testLoginRequestBody.ts`** — replaces the old `testClientEnvironment.ts` with a broader test suite for the login-request body, adding an `SPCS_TOKEN` section alongside the existing `CLIENT_ENVIRONMENT` tests.

### Checklist

### DIFF
--- a/test/integration/testLoginRequestBody.ts
+++ b/test/integration/testLoginRequestBody.ts
@@ -152,6 +152,7 @@ describe('/login-request body', () => {
     it('includes SPCS_TOKEN in SPCS environment', async () => {
       sinon
         .stub(fs, 'readFileSync')
+        .callThrough()
         .withArgs('/snowflake/session/spcs_token', 'utf-8')
         .returns('test-spcs-token');
       await initConnection();


lib/authentication/spcs_token.ts
```diff
@@ -0,0 +1,12 @@
+import fs from 'fs';
+import Logger from '../logger';
+
+export function getSpcsToken(): string | null {
+  try {
+    const tokenPath = process.env.SF_SPCS_TOKEN_PATH || '/snowflake/session/spcs_token';
+    return fs.readFileSync(tokenPath, 'utf-8').trim();
```
sfc-gh-rsavenok, line 7: We had some concerns regarding additional permissions checks before reading the file on Slack. This will be polished in another PR if needed

sfc-gh-rsavenok: Same with the read cache, discussion in progress

PR #1381, opened 2026-04-13
SNOW-3362718: mimic 'glob' dot:false behaviour which was the default pre v2.3.3

sfc-gh-dszmolka:
### Description

Prior to v2.3.3 the `glob` dependency was used to match file names for `PUT` etc., and we used it with the [default](https://github.com/isaacs/node-glob/blob/v10.5.0/README.md?plain=1#L411-L412) `dot: false` behaviour:
> All options are optional, boolean, and false by default, unless
> otherwise noted.

We ditched `glob` due to CVE but now we seem to have a subtle issue where previously (see above) the leading dot in filenames (e.g. `.streamlit`) was not matched, thus the `PUT` did not attempt to upload the whole directory. Now it does. 

This change aims to 
* check for first char of the filename to get back to the previous behaviour
* add some tests to prevent it from breaking silently in the future

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 Changes:
 
 - Replaced deprecated Node.js `url.parse()` with the WHATWG `URL` constructor (snowflakedb/snowflake-connector-nodejs#1380)
-- Fixed file name pattern wildcard matching behaviour, should be now similar to the dropped 'glob' package's default 'dot:false' behaviour (snowflakedb/snowflake-connector-nodejs#1381)
+- Fixed file name pattern matching to not match dot-prefixed files/directories by default, aligning with standard glob behavior and default of `dot: false`. Was lingering around since v2.3.3. (snowflakedb/snowflake-connector-nodejs#1381)
 
 ## 2.4.0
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 Changes:
 
 - Replaced deprecated Node.js `url.parse()` with the WHATWG `URL` constructor (snowflakedb/snowflake-connector-nodejs#1380)
+- Fixed file name pattern wildcard matching behaviour, should be now similar to the dropped 'glob' package's default 'dot:false' behaviour (snowflakedb/snowflake-connector-nodejs#1381)
```
sfc-gh-rsavenok: Please mention which version broke this and explicitly describe what exactly we fixed e.g. "Fixed file name pattern matching to not match dot-prefixed files/directories by default, aligning with standard glob behavior"

sfc-gh-rsavenok (approved):

CHANGELOG.md
sfc-gh-dszmolka: sure can do; how about this:
> - Fixed file name pattern matching to not match dot-prefixed files/directories by default, aligning with standard glob behavior and default of `dot: false`. Was lingering around since v2.3.3. (snowflakedb/snowflake-connector-nodejs#1381)


(commit pushed but if you prefer still something else let me know please @sfc-gh-rsavenok

Issue #790, opened 2024-03-06
Out of Memory Issue while streaming data


kartikgupta2607:
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1. What version of NodeJS driver are you using? ->snowflake-sdk -> Tried with 1.6.20, 1.9.3
   
2. What operating system and processor architecture are you using? -> macOS 14.3.1, arm64
   
3. What version of NodeJS are you using? -> v18.12.1

4. What are the component versions in the environment (`npm list`)? -> NA

5. Server version: -> 8.9.1

6. What did you do?
Tried running this ([snowflake_OOM.txt](https://github.com/snowflakedb/snowflake-connector-nodejs/files/14513890/snowflake_OOM.txt)) script to export records from `SNOWFLAKE_SAMPLE_DATA.TPCH_SF1000.CUSTOMER` table, after limiting max-old-space-size=150 while running the script. A simple `SELECT * FROM CUSTOMER;` was used, but the node process exited after exporting some rows (200000) with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` error but on modifying the same query with a LIMIT clause (with limit of 1,000,000), it was able to export 1M records with the same old-space limit. I tried workarounds mentioned in #43 and the linked issues but none worked. Tried using the `streamResult` option in the connection config and while executing the query, tried degrading to 1.6.20.
Following is the metadata of the source table
<img width="662" alt="image" src="https://github.com/snowflakedb/snowflake-connector-nodejs/assets/55034563/add54290-9840-4dde-8041-3021ec1c97fd">
As it doesn't have any VARIANT type column, so #536 and the related issues because of it shouldn't matter in our case
  
7. What did you expect to see?

   It should be able to export the same number of records before the OOM issue occurs, also it seems a large buffer is fetched if the LIMIT isn't present in the query as it seems the GC triggers a bit late for query without LIMIT

8. Can you set logging to DEBUG and collect the logs? -> Can't upload logs due to company security policies.

```
var snowflake = require('snowflake-sdk');
snowflake.configure(
{
  logLevel: 'trace'
});
```
   
9. What is your Snowflake account identifier, if any? (Optional)

sfc-gh-dszmolka:
hi - thanks for raising this issue with us, i'll take a look and see how we can proceed

shelbylsmith:
[...]

Issue #1256, opened 2026-02-04
SNOW-3051230: feedback about `minicore`


mtoy-googly-moogly:
## What is the current behavior?

Starting in version 2.3.2, the `snowflake-sdk` package includes a native NAPI addon (`sf_mini_core`) for telemetry collection. This native binary is bundled directly as a dependency, meaning it's always included regardless of the consumer's use case.

This causes significant problems when bundling the SDK for VS Code extensions (and likely Electron apps), where native binaries require special handling across multiple platforms (macOS, Windows, Linux). Standard bundlers like esbuild and webpack cannot properly handle `.node` files, and VS Code extensions have additional constraints around native module loading.

As a workaround, we are pinned to version 2.3.1 to avoid the native binary.

## What is the desired behavior?

We'd like to see the native telemetry addon an **optional dependency** using npm's `optionalDependencies` with platform-specific packages, similar to how DuckDB handles this.

DuckDB's approach:
- The main package (`@duckdb/node-api`) contains no native code
- Native bindings are in separate platform-specific packages (`@duckdb/node-bindings-darwin-arm64`, etc.)
- These are listed as `optionalDependencies`, so npm only installs the one matching your platform
- Consumers who don't need the native bindings (or handle them differently) can exclude them

For snowflake-sdk, this could look like:
- `snowflake-sdk` - main package, pure JavaScript
- `@snowflake/telemetry-darwin-arm64`, `@snowflake/telemetry-darwin-x64`, `@snowflake/telemetry-linux-x64`, `@snowflake/telemetry-win32-x64` - optional native packages

Alternatively, provide a configuration option to disable the telemetry native addon entirely.

## How would this improve `snowflake-connector-nodejs`?

1. **VS Code extension compatibility** - Extensions can bundle the SDK without native binary issues
2. **Electron app compatibility** - Similar bundling challenges exist for Electron
3. **Smaller bundle sizes** - Users only get binaries for their platform, not all platforms
4. **Serverless compatibility** - AWS Lambda and similar environments often have issues with native modules
5. **Maintains existing functionality** - Users who want telemetry still get it automatically via optionalDependencies

## References, Other Background

- DuckDB's approach: https://duckdb.org/2024/12/18/duckdb-node-neo-client
- Related issue #449 (optional peer dependencies for cloud SDKs) addresses a similar concern
[...]

PR #1382, opened 2026-04-15
Reduce peak memory when streaming large result sets

kartikgupta2607:
### Description

When streaming large result sets via `streamRows()`, the previous chunk's `_rows` array is freed only after `getRows()` on the next chunk - meaning the old rows coexist with the new chunk's `JSON.parse` allocation and any prefetched `_rowsetAsString` buffers, raising the transient memory peak unnecessarily.

**Changes:**

1. Move `previousChunk.clearRows()` before `chunk.getRows()` in `onResultStreamData` - frees the previous chunk's rows before the next chunk's parse allocates.
2. Null consumed `rowBuffer` entries in the `processRowBuffer` while loop - allows GC to reclaim row objects incrementally rather than retaining the entire buffer.

These are minor SDK-side optimizations that reduce transient peak memory during chunk transitions. For environments streaming very large exports, consumers can additionally tune:

- `resultPrefetch: 1` - reduces prefetched chunks from 2 to 1, lowering baseline memory from buffered `_rowsetAsString` payloads
- `rowStreamHighWaterMark` (v2.3.5) - controls row-level Readable buffering
- `CLIENT_RESULT_CHUNK_SIZE` session parameter - controls server-side chunk sizing (currently requires server-side enablement, not yet GA)
- Streaming JSON parsing and Arrow format support would yield the largest gains but require significant SDK changes

Related: #790

### Checklist

- [ ] ~~Extend the types in index.d.ts file (if necessary)~~ - N/A, no public API change
- [ ] ~~Extend the README / documentation~~ - N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changes:
 
 - Replaced deprecated Node.js `url.parse()` with the WHATWG `URL` constructor (snowflakedb/snowflake-connector-nodejs#1380)
 - Fixed file name pattern matching to not match dot-prefixed files/directories by default, aligning with standard glob behavior and default of `dot: false`. Was lingering around since v2.3.3. (snowflakedb/snowflake-connector-nodejs#1381)
+- Reduced peak memory usage when streaming large result sets by reordering chunk lifecycle to free the previous chunk before parsing the next one (snowflakedb/snowflake-connector-nodejs#790)
 
 ## 2.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,15 @@
 # Changelog
 
-##
+## Upcoming Release
 
 Changes:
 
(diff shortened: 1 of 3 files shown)


sfc-gh-rsavenok:
Hi! Thx for the improvement! 

The change looks good, starting the full test build

sfc-gh-rsavenok:
Also, a comment on 
> Streaming JSON parsing and Arrow format support would yield the largest gains but require significant SDK changes

3.x driver is going to be written in Rust and compiled as a .node addon. By default, it will use the arrow format ([context](https://github.com/snowflakedb/snowflake-connector-nodejs/issues/1256#issuecomment-3852678552))

If you curios how Universal Driver would look like, here's the repo - https://github.com/snowflakedb/universal-driver There's also a vibe coded draft PR with a sample node driver that could connect and select 1

sfc-gh-rsavenok (approved):

PR #1407, opened 2026-05-14
NO-SNOW Fix OAuth invalid-token test against new server error format

sfc-gh-rkowalski:
## Summary
- The OAuth invalid-token authentication test (`testOauth.js > Oauth authentication > Invalid token`) was failing because the server now appends a request ID in brackets to the error message (e.g. `Invalid OAuth access token. [83807702746770]`), breaking the exact `assert.strictEqual` assertion in `AuthTest.verifyErrorWasThrown`.
- Extend `verifyErrorWasThrown` to accept either a string (preserving the existing exact-match semantics for all 25 other call sites) or a `RegExp` (used via `assert.match`).
- Update only the single failing test (`testOauth.js:38`) to pass a regex that tolerates the optional bracketed request ID: `/^Invalid OAuth access token\. (\[\d+\] ?)?$/`.

## Test plan
- [ ] `Invalid token` test in `testOauth.js` passes against the current server (with bracketed request ID).
- [ ] All other auth tests using `verifyErrorWasThrown` continue to assert exact-equals — no behavioral change for them (`testPAT`, `testOauthOktaClientCredentials`, `testOauthSnowflakeWildcardsAuthorizationCode`, `testOauthOktaAuthorizationCode`, `testOauthSnowflakeAuthorizationCode`, `testOkta`, `testExternalBrowser`, plus the wiremock OAuth integration tests).

### DIFF
--- a/test/authentication/testOauth.js
+++ b/test/authentication/testOauth.js
@@ -35,7 +35,7 @@ describe('Oauth authentication', function () {
     const connectionOption = { ...connParameters.oauth, token: 'invalidToken' };
     authTest.createConnection(connectionOption);
     await authTest.connectAsync();
-    authTest.verifyErrorWasThrown('Invalid OAuth access token. ');
+    authTest.verifyErrorWasThrown(/Invalid OAuth access token/);
     await authTest.verifyConnectionIsNotUp(
       'Unable to perform operation using terminated connection.',
     );
--- a/test/authentication/authTestsBaseClass.js
+++ b/test/authentication/authTestsBaseClass.js
@@ -78,8 +78,12 @@ class AuthTest {
     assert.equal(this.error, null);
   }
 
-  verifyErrorWasThrown(message) {
-    assert.strictEqual(this.error?.message, message);
+  verifyErrorWasThrown(messageOrRegex) {
+    if (messageOrRegex instanceof RegExp) {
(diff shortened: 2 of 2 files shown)


test/authentication/testOauth.js
```diff
@@ -35,7 +35,7 @@ describe('Oauth authentication', function () {
     const connectionOption = { ...connParameters.oauth, token: 'invalidToken' };
     authTest.createConnection(connectionOption);
     await authTest.connectAsync();
-    authTest.verifyErrorWasThrown('Invalid OAuth access token. ');
+    authTest.verifyErrorWasThrown(/^Invalid OAuth access token\. (\[\d+\] ?)?$/);
```
sfc-gh-rsavenok: I'd just match `/Invalid OAuth access token/` like we do in other tests. We don't really care about the number at the end or if it will be prefixed

sfc-gh-rsavenok (approved):

test/authentication/testOauth.js
sfc-gh-rkowalski: Gotcha, will update the assertion

sfc-gh-rkowalski: Updated to use `/Invalid OAuth access token/` — thanks!

PR #1408, opened 2026-05-15
Add native session_token authentication support

sfc-gh-mborins:
## Summary
- Adds native `SESSION_TOKEN` authenticator so the SDK can connect using pre-existing session/master tokens from `connections.toml` (the format Cortex Code Desktop writes)
- Auto-detects `SESSION_TOKEN` when `sessionToken` is present without an explicit `authenticator` field, and skips credential validation (username/password) for this path
- Fixes the `Cannot read properties of undefined (reading 'toLowerCase')` crash that occurred when a TOML entry had `session_token`/`master_token` but no `authenticator` field
- Makes connection pools compatible with session token auth by short-circuiting the login flow when tokens are already available
- Warns at construction time when `masterToken` is missing, since token renewal will fail when the session expires

## Test plan
- [x] New unit tests (12 cases) covering: AuthSessionToken behavior, auto-detection of SESSION_TOKEN authenticator, ConnectionConfig with/without session tokens, credential validation bypass, masterToken warning, connection.connect() succeeding without login, pool creation
- [ ] Existing unit test suite passes
- [ ] Manual verification with a real `connections.toml` containing `session_token`/`master_token` fields
- [ ] Verify pool.use() works end-to-end with session token connections

.... Generated with [Cortex Code](https://docs.snowflake.com/en/user-guide/cortex-code/cortex-code)


sfc-gh-rsavenok:
Closing as we're going to have a new API to create child session and session+master token sharing won't be available in the future

PR #1411, opened 2026-05-19
SNOW-3351450: emit client_connection_identifier_shape in-band telemetry

sfc-gh-boler:
## Summary

Node.js port of the cross-driver `client_connection_identifier_shape` in-band telemetry feature. Emits one telemetry record per successful login describing which connection-identifier fields the user supplied — `account_provided`, `account_with_region`, `account_org_provided`, `region_provided`, `host_provided` — without including any hostname or account value.

This mirrors the parallel ports already prepared for the Go driver (snowflakedb/gosnowflake#1797) and the Python connector (snowflakedb/snowflake-connector-python#2877). The wire-format payload (the five stringified-boolean keys, plus the `client_connection_identifier_shape` event-type string) is byte-identical across all four drivers.

## Node.js-specific decisions

- **Single capture site.** Node.js has no DSN parser — `snowflake.createConnection(options)` is the only public surface, and both the programmatic-options path and the `connections.toml` path funnel through `consolidateHostAndAccount(options)` in `lib/connection/connection_config.js`. Capturing the shape as the very first statement in that function (before the dotted-account split, the `-org` strip, and the `Util.constructHostname` synthesis) covers both paths from one place.
- **`accessUrl` collapses into `host_provided`.** Node.js exposes a fully-qualified `accessUrl` (scheme + host + port) as a fourth identifier dimension that Go/Python/JDBC do not have. For shape purposes both `options.host` and `options.accessUrl` mean "the user supplied an explicit endpoint", so they collapse into the single `host_provided` flag and the wire schema stays the five-keys flat shared across drivers.
- **Emission is unconditional.** Unlike Go/Python/JDBC — which gate on the server-side `CLIENT_TELEMETRY_ENABLED` parameter — the Node.js emit follows the existing `minicore_error` precedent and just sits next to it as a sibling in-band-telemetry record. The local kill switch `SF_TELEMETRY_DISABLE_CONNECTION_SHAPE=true` (case-insensitive `true` only, no truthy aliases) still suppresses emission.
- **Wire envelope nesting.** This driver's `buildInbandTelemetryRequest` builder nests the payload under `message.value.*` rather than flattening it onto `message.*` like Go/Python do. This is a pre-existing server-side concern, not something to fix from the client side.
- **Public API surface preserved.** The captured shape lives on an internal-only `_inputShape` property on the `ConnectionConfig` instance, accessed only by the emit site via the private `_getInputShape()` getter. Nothing was added to `WIP_ConnectionOptions`, `index.d.ts`, or any other public typing.

## Changes

New files:

- `lib/connection/connection_identifier_shape.ts` — pure-helper module with the `ConnectionIdentifierShape` interface, the `recordInputShape(options)` capture function, and the `shapeToTelemetryPayload` serializer.
- `test/unit/connection/connection_identifier_shape_test.ts` — table-driven tests for the capture helper covering the truth table from `consolidateHostAndAccount`'s existing test cases (bare account, dotted account, org-prefixed account, both, region-tail dashes don't count, host only, accessUrl only, host+accessUrl together, defensive null/undefined/empty/non-string cases).
- `test/unit/telemetry/client_connection_identifier_shape_test.ts` — unit tests for the env-switch logic and the wire-format envelope built via `buildInbandTelemetryRequest`.

Modified files:

- `lib/connection/connection_config.js` — capture the shape inside `consolidateHostAndAccount(options)` before any normalization runs; lift it onto `this._inputShape` in the constructor; expose `_getInputShape()` for the emit site.
- `lib/telemetry/inband_telemetry.ts` — export `CLIENT_CONNECTION_IDENTIFIER_SHAPE`, `DISABLE_CONNECTION_SHAPE_ENV`, and `isConnectionShapeTelemetryDisabledByEnv()` so the event-type string and env-var name live next to the envelope builder, matching the colocation pattern Go and Python use.
- `lib/services/sf.js` — fire the telemetry record as a sibling block right after the existing `minicore_error` emit in `StateConnecting.requestCallback`, fire-and-forget with a `.catch` debug log.
- `test/integration/testInbandTelemetry.ts` — extended with three new wiremock-backed cases: shape emitted with the right payload for an `accessUrl`-only connection; the org-prefixed dotted account flips both `account_with_region` and `account_org_provided`; the env-switch suppresses emission entirely.
- `CHANGELOG.md` — `## Upcoming Release` entry mirroring the Go/Python wording, called out as time-boxed.

## Removal tracking

Every new field, function, and comment that exists only because of this telemetry carries a `TODO(SNOW-3548350)` line so a future grep finds them. Removal is tracked in **SNOW-3548350** (target: 2026-11-30).

## Test plan

- [x] `npx mocha test/unit/connection/connection_identifier_shape_test.ts test/unit/telemetry/client_connection_identifier_shape_test.ts` — 44 new tests pass.
- [x] `npx mocha test/unit/connection/connection_config_test.js` — 135 existing connection-config tests still pass (no regression on the truth table that our shape-capture cases borrow from).
- [x] `npx mocha 'test/unit/connection/**/*.{js,ts}' 'test/unit/telemetry/**/*.{js,ts}' 'test/unit/services/**/*.{js,ts}'` — 334 tests pass.
- [x] `npx oxlint --max-warnings=0 .` clean.
- [x] `npx prettier --check .` clean.
- [ ] Wiremock-backed integration tests in `test/integration/testInbandTelemetry.ts` to be exercised by CI (require local wiremock + Java, skipped here).


sfc-gh-boler:
Pushed a CI fix for the integration test failures introduced by this PR.

**Root cause.** The three new shape-telemetry integration tests in `test/integration/testInbandTelemetry.ts` originally used the local `initConnection` helper, which calls `testUtil.createConnection` → `coreInstance.createConnection({...connOptions.valid, ...override})`. `connOptions.valid` folds in `account`, `host`, and (transitively, on some runners) `region` from `SNOWFLAKE_TEST_*` env vars — which are empty locally but populated in CI. As a result the shape assertions were environment-dependent: they passed locally and failed on Ubuntu/AWS, Mac/AZURE, etc. with diffs like `+ account_provided: 'true'` / `- account_provided: 'false'`.

**Fix.** Added a `initShapeConnection` helper that calls `snowflake.createConnection(...)` directly with an explicit minimal options object (`username`, `password`, plus whatever the test wants to assert against). Bypasses `connOptions.valid` entirely so the shape payload is deterministic regardless of `SNOWFLAKE_TEST_*` env state. Updated the three shape tests to use it; the existing `minicore error` test still goes through `initConnection` as before. Also adjusted test 1's expected payload to `{account_provided: 'true', host_provided: 'true', ...}` since the new explicit options include a bare-locator `account: 'shape-test-account'` (the old "host-only" framing relied on the leak-induced absence of `account` and can't be reproduced without a `*.snowflakecomputing.com`-shaped wiremock URL anyway).

**Other failing checks (pre-existing, not introduced here):**
- `Audit / Check all dependencies` — the same `npm audit` advisories on `diff`, `mocha`, `sinon`, `elliptic`, `serialize-javascript` have been failing on `master` for every run going back to May 13 ([master history](https://github.com/snowflakedb/snowflake-connector-nodejs/actions/workflows/audit.yml?query=branch%3Amaster)). This PR does not touch `package.json` / `package-lock.json`.
- `connection with CRL validation` (testCrl.ts) and `SF service tests > GUID called for all` (testRequestParams.js) — also flaky/pre-existing; this PR doesn't touch those code paths.

Local checks: `prettier`, `oxlint --max-warnings=0`, `tsc --noEmit`, and the 44 unit tests under `test/unit/connection/connection_identifier_shape_test.ts` + `test/unit/telemetry/client_connection_identifier_shape_test.ts` all pass.

sfc-gh-boler:
Follow-up CI fix: my first round used `'shape-test-account'` as the test account name, but the dashes in that string correctly tripped `account_org_provided` (the implementation looks at the account portion before the first `.` for dashes, which means `shape-test-account` is treated as an org-prefixed form). Switched to `'shapetestacct'` (no dashes, no dots) so the bare-locator test case actually exercises the bare-locator path. Asserted-against payload is now byte-identical to what the implementation emits.

sfc-gh-boler:
Pushed review-round-1 fixes addressing reviewer findings on cross-driver consistency with gosnowflake (the spec) and the `_inputShape` courier cleanup.

## Changes

- **`account_org_provided` logic aligned to Go** — `recordInputShape` now gates the dot-split on `dotIndex > 0` (matching gosnowflake's `recordAccountShape` in `internal/config/dsn.go`). Pathological leading-dot inputs like `.us-east-1` now correctly report `account_org_provided=true` / `account_with_region=false` rather than the previous (Node.js-only) `false` / `true` interpretation. The dash search runs over the full string when there is no real account/region split, mirroring the Go reference byte-for-byte. Sibling Python PR updated in the same direction.
- **`consolidateHostAndAccount` returns the shape directly** — dropped the `options._inputShape` courier and the `delete options._inputShape` cleanup. `consolidateHostAndAccount(options)` now returns a `ConnectionIdentifierShape` that the `ConnectionConfig` constructor assigns to `this._inputShape` in one expression. The public `options` object stops carrying a temporary internal key, removing the info-leakage and cleanup-gap risk the reviewer flagged.
- **`_getInputShape` JSDoc tightened** — return type is now `import('./connection_identifier_shape').ConnectionIdentifierShape | undefined` instead of the bare `Object | undefined`.
- **Unit test name + expectation updated** — `'leading-dot account does not credit account_with_region for an empty portion'` renamed to `'leading-dot account does not split — dash search runs over the full raw value'`, with expectations flipped to `accountWithRegion: false, accountOrgProvided: true` to match the new Go-aligned semantics.

70 telemetry- and shape-related unit tests pass locally. `oxlint`, `prettier --check`, and `tsc --noEmit dist/index.d.ts` are all clean. The unrelated pre-existing CWD-dependent failures in `test/unit/configuration/configuration_finding_test.js` and the network-dependent `LargeResultSetService` tests are not touched by these changes.

sfc-gh-boler:
Pushed CI fix for the three pre-existing assertions that the new `client_connection_identifier_shape` telemetry emission knocks off-by-one.

## Root cause

Adding the post-login `POST /telemetry/send` call (the in-band envelope `buildInbandTelemetryRequest` already exposed for `minicore_error`, now fired on every successful login) shifts two distinct observable counts:

1. **`testCrl.ts`** — CRL validation runs on every HTTPS request made by the connection. The extra telemetry POST goes through the same `Agent` and therefore bumps `validateCrlSpy.callCount` from 1 to 2 in two cases (plain TLS and proxy).
2. **`testRequestParams.js::'GUID called for all'`** — the test enforces an invariant that *every request bearing a request-GUID header must be on a known path*. `/telemetry/send` is now legitimately on that list (it has always been GUID-tagged, just never previously fired during a successful login in CI), so `totalCallsWithGUIDCount === expectedCallsWithGUIDCount` fails with "Total calls with guid: 3. Expected calls with guid: 2".

## Fixes

- **`testCrl.ts`** (two sites) — assert that CRL ran *at least once* rather than exactly once. The original intent was "CRL validation was applied to this connection," which is what `>= 1` actually captures; the strict `=== 1` was an artifact of pre-PR call counts. Added a comment so future telemetry additions don't re-trigger the same flake.
- **`testRequestParams.js`** — added `'telemetry/send'` to `pathsExpectedToIncludeGuid` alongside the existing `session/v1/login-request`, `queries/v1/query-request`, and `session?delete=true&requestId` entries. The endpoint already carried a GUID (it shares `HttpClient` with the auth-side requests); the test list just hadn't enumerated it.

No production-code changes. Pushed as `cc4e743a`.

Local `npx oxlint` and `npx prettier --check` on both test files come back clean. I'd retry the matrix once this lands; based on the previous run the same 53 jobs were failing on the same 3 tests, so this should turn them green.

(`Check all dependencies` is still red, but that's pre-existing on `master` per my earlier comment — no `package.json` / `package-lock.json` deltas in this PR.)

sfc-gh-rsavenok (approved):

PR #1412, opened 2026-05-21
NO-SNOW  Fixed platform-detection probe not aborting within 200ms on Bun

sfc-gh-rsavenok:
### Description

Fixes a bug where the platform-detection telemetry probe could block far beyond its intended 200ms timeout when the connector runs on Bun.

**Root cause:** `httpRequest()` in `lib/telemetry/platform_detection.ts` used `node:http` with an `AbortSignal`. On Bun, `node:http` does not honor `AbortSignal` during the connect phase (see [oven-sh/bun#31167](https://github.com/oven-sh/bun/issues/31167)), so probes to unreachable cloud-metadata endpoints (e.g. `169.254.169.254`) would hang well past the 200ms budget — delaying the first query on connect.

We can't simply switch everything to `fetch` either: on Node.js, aborting an in-flight `fetch` can leave undici's socket alive and delay process exit (tested on Node 22 & 24), which is why `node:http` was chosen originally.

**Fix:** Dispatch the HTTP helper by runtime.

- On **Node.js**, keep using `node:http` (avoids the undici socket-leak / delayed-exit issue).
- On **Bun**, use `fetch` (correctly aborts on `AbortSignal`).

Runtime is detected via `!!process.versions?.bun`. Both code paths return the same `HttpRequestResult` shape, so callers are unchanged.

No public API or behavior change on Node.js; Bun users now see platform detection respect the 200ms timeout.

### Checklist

- [x] Extend the types in index.d.ts file (if necessary) — N/A, internal change
- [x] Extend the README / documentation and ensure is properly displayed (if necessary) — N/A
- [x] Provide JIRA issue id (if possible) or GitHub issue id in commit message — `NO-SNOW` used per existing commit

### DIFF
--- a/lib/telemetry/platform_detection.ts
+++ b/lib/telemetry/platform_detection.ts
@@ -159,14 +159,14 @@ type HttpRequestResult = {
   body: string;
 };
 
-const IS_BUN = typeof process !== 'undefined' && !!process.versions?.bun;
 function httpRequest(
   url: string,
   signal: AbortSignal,
   headers?: Record<string, string>,
   method: string = 'GET',
 ): Promise<HttpRequestResult> {
-  return IS_BUN
+  const isBun = typeof process !== 'undefined' && !!process.versions?.bun;
+  return isBun
     ? httpRequestFetch(url, signal, headers, method)
     : httpRequestNode(url, signal, headers, method);
 }
--- a/lib/telemetry/platform_detection.ts
+++ b/lib/telemetry/platform_detection.ts
@@ -146,19 +146,52 @@ function envPresent(...envVars: string[]): boolean {
 }
 
 /*
- * We intentionally avoid using fetch() here, since aborting fetch requests
- * can sometimes leave sockets open and prevent the Node.js process from exiting (tested on 22).
+ * Runtime-dispatched HTTP helper.
+ *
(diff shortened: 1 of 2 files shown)


lib/telemetry/platform_detection.ts
```diff
@@ -146,19 +146,52 @@ function envPresent(...envVars: string[]): boolean {
...
   signal: AbortSignal,
   headers?: Record<string, string>,
   method: string = 'GET',
-): Promise<{
-  ok: boolean;
-  headers: http.IncomingHttpHeaders;
-  body: string;
-}> {
+): Promise<HttpRequestResult> {
+  const isBun = typeof process !== 'undefined' && !!process.versions?.bun;
```
sfc-gh-mcepiga, line 168: Should we consider adding bun setup to the ci matrix?

sfc-gh-pbulawa (approved):

lib/telemetry/platform_detection.ts
sfc-gh-rsavenok: I tried, but the current test runner is not compatible with Bun. I will have another PR adding a simple smoke test for bun, but full matrix support for bun would be only in UD

PR #1415, opened 2026-05-25
SNOW-3411971 AWS Outbound Identity Federation Support for WIF

sfc-gh-rsavenok:
### Description

SNOW-3411971: Switch AWS Workload Identity attestation to STS Outbound Identity Federation.

The AWS provider for `WORKLOAD_IDENTITY` previously produced a token by building a SigV4-signed `GetCallerIdentity` HTTP request and base64-encoding the `{url, method, headers}` envelope. This PR replaces that flow with STS's new outbound identity federation, calling `GetWebIdentityToken` and forwarding the signed JWT (audience `snowflakecomputing.com`, `ES384`) directly as the attestation token.

Changes:

- `lib/authentication/auth_workload_identity/attestation_aws.ts`: call `STSClient.send(new GetWebIdentityTokenCommand({ Audience: ['snowflakecomputing.com'], SigningAlgorithm: 'ES384' }))` and return `response.WebIdentityToken`. Throw a clear error if STS returns no token. The impersonation chain (`AssumeRole` hops) is unchanged.
- Removed the now-unused helpers (`getStsHostname`) and the manual SigV4 signing path.
- `package.json`: dropped `@aws-crypto/sha256-js`, `@smithy/protocol-http`, and `@smithy/signature-v4` — no longer imported anywhere.
- Tests:
  - `attestation_aws_test.ts`: rewiremock now exposes both `AssumeRoleCommand` and `GetWebIdentityTokenCommand` as distinct fake classes, with `STSClient.send` dispatching by `instanceof` to per-command stubs. Added cases for the new "no `WebIdentityToken` returned" error and for the JWT being returned both with and without an impersonation path. Removed the obsolete `getStsHostname` suite.
  - `auth_workload_identity_test.ts`: the AWS path now hits STS unconditionally, so the suite mocks `@aws-sdk/client-sts` minimally (returns the JWT). The end-to-end impersonation behavior is covered by `attestation_aws_test.ts` and `test/auth-workload-identity-e2e.ts`, so no duplicate impersonation case was added here.
  - `test_utils.ts`: dropped the `assertAwsAttestationToken` helper (SigV4 envelope assertions no longer apply); added `AWS_WEB_IDENTITY_TOKEN` constant.
- `CHANGELOG.md`: entry under Upcoming Release → Changes.

### Checklist

- [x] Extend the types in index.d.ts file (if necessary) — N/A, no public API changes
- [x] Extend the README / documentation and ensure is properly displayed (if necessary) — N/A; consider updating the stale JSDoc on `workloadIdentityProvider` in `lib/connection/types.ts` (currently still says "encodes signed GetCallerIdentity request as token")
- [x] Provide JIRA issue id (if possible) or GitHub issue id in commit message (SNOW-3411971)

### DIFF
--- a/lib/authentication/auth_workload_identity/attestation_aws.ts
+++ b/lib/authentication/auth_workload_identity/attestation_aws.ts
@@ -59,6 +59,5 @@ export async function getAwsAttestationToken(impersonationPath?: string[]) {
     throw new Error('Failed to obtain AWS web identity token from STS');
   }
 
-  Logger().debug(`AWS outbound token prefix: ${token.slice(0, 10)}`);
   return token;
 }
--- a/lib/authentication/auth_workload_identity/attestation_aws.ts
+++ b/lib/authentication/auth_workload_identity/attestation_aws.ts
@@ -1,9 +1,6 @@
 import { defaultProvider } from '@aws-sdk/credential-provider-node';
-import { STSClient, AssumeRoleCommand } from '@aws-sdk/client-sts';
+import { STSClient, AssumeRoleCommand, GetWebIdentityTokenCommand } from '@aws-sdk/client-sts';
 import { MetadataService } from '@aws-sdk/ec2-metadata-service';
-import { HttpRequest } from '@smithy/protocol-http';
-import { SignatureV4 } from '@smithy/signature-v4';
-import { Sha256 } from '@aws-crypto/sha256-js';
 import Logger from '../../logger';
 
 export async function getAwsCredentials(region: string, impersonationPath: string[] = []) {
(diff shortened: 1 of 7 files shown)


lib/authentication/auth_workload_identity/attestation_aws.ts
```diff
@@ -45,42 +42,23 @@ export async function getAwsRegion() {
...
+    throw new Error('Failed to obtain AWS web identity token from STS');
+  }
 
-  const token = {
-    url: `https://${stsHostname}/?Action=GetCallerIdentity&Version=2011-06-15`,
-    method: 'POST',
-    headers: signedRequest.headers,
-  };
-  return btoa(JSON.stringify(token));
+  Logger().debug(`AWS outbound token prefix: ${token.slice(0, 10)}`);
```
sfc-gh-rkowalski: Why do we need this? I think effectively this is always the same value

sfc-gh-rkowalski (approved):
LGTM

lib/authentication/auth_workload_identity/attestation_aws.ts
sfc-gh-rsavenok: Same was added in python, but you're right, I'll remove it

PR #1416, opened 2026-05-25
NO-SNOW browserResponseRenderer connection option

sfc-gh-rsavenok:
### Description

Adds a new `browserResponseRenderer` connection option that lets users customize the HTML page shown in the browser after an `EXTERNALBROWSER` (SAML) or `OAUTH_AUTHORIZATION_CODE` authentication callback completes, instead of the hardcoded default confirmation message.

**Public API**
- New optional connection option `browserResponseRenderer?: (result: { error: string | null }) => string`. The function returns the full HTML page to display. On success `error` is `null`; on failure it contains the IdP error message (already HTML-escaped and safe to embed). Documented in `lib/connection/types.ts` and registered in `lib/connection/connection_config.js`.

**Behavior changes**
- The local callback server now responds with `Content-Type: text/html; charset=utf-8` (previously `text/plain`) so custom HTML renders correctly. The default response is preserved and HTML-escaped.
- `createServer` in `authentication_util.js` is now shared between the SAML (`auth_web.js`) and OAuth (`auth_oauth_authorization_code.js`) flows, accepting an `options.renderer` and a `reject` callback.

**Robustness fixes** (surfaced while wiring up the shared server)
- Wrapped the socket `data` handler in try/catch so a throwing renderer or malformed redirect rejects the auth promise instead of crashing the process via `uncaughtException`.
- Socket `error` handler now rejects the promise instead of `throw`ing from an `EventEmitter` listener (which could not reach the caller and would crash the process). `ECONNRESET` is still treated as an expected browser disconnect.
- `prepareError` now guards against a missing `error_description` from the IdP rather than throwing.
- OAuth flow now `await`s the server `listening`/`error` events before opening the browser and closes the server in a `finally` block to avoid leaking the listener on timeouts or browser-launch failures.

### Checklist


lib/authentication/authentication_util.js
```diff
@@ -71,7 +93,13 @@ const withBrowserActionTimeout = (millis, promise) => {
 function prepareError(rejected) {
   const errorResponse = querystring.parse(rejected.substring(rejected.indexOf('?') + 1));
   const error = errorResponse['error'];
```
sfc-gh-rkowalski, line 95: `error` is always passed or can it be missing like `error_description`?

sfc-gh-rkowalski (approved):

lib/authentication/authentication_util.js
sfc-gh-rsavenok: `prepareError` is called only if queryString includes `error`

Issue #1418, opened 2026-05-31
SNOW-3589640: streamRows() hangs forever when a chunk\'s getRows() returns undefined (slice TypeError gets swallowed)


matanmaze:
Version: snowflake-sdk 2.4.2, Node 22, Linux
On large result sets we occasionally get this logged, after which the stream never ends or errors and our consumer hangs:
```
Callback failed with TypeError: Cannot read properties of undefined (reading 'slice')
```
It only happens on big results (ours was ~108k rows / ~21 chunks), so it's in the large_result_set chunk path.
Looks like `chunk.getRows()` (lib/connection/result/chunk.js) can return undefined when `_rowsetAsString`/`_rowset` aren't set, and the callers slice it without checking:
```js
// lib/connection/result/row_stream.js:183
rowBuffer = chunk.getRows().slice(...)
```
The throw then gets swallowed here, so the stream goes silent instead of erroring:
```js
// lib/services/large_result_set.js:101
try { options.callback(err, body); }
catch (e) { Logger.getInstance().error(`Callback failed with ${e}`); }
```
Repro (intermittent, easier under memory pressure / `--max-old-space-size`):
```js
const stmt = connection.execute({
  sqlText: 'select * from big_table',  // large enough to spill into multiple chunks
  streamResult: true,
});
await new Promise((resolve, reject) => {
  const stream = stmt.streamRows();
  stream.on('data', () => {});
  stream.on('end', resolve);
  stream.on('error', reject);   // sometimes never fires -> hangs
});
```
Two things here: getRows() being undefined and sliced unguarded, and the swallowed callback error never surfacing as a stream 'error'. The second is the worse one since it turns a chunk failure into a silent hang. Possibly related to #694 (same slice error in chunk.js, different cause, closed). Is this fixed on a newer release?

sfc-gh-dszmolka:
thank you for reporting this isssue @matanmaze , we'll check it out.

sfc-gh-rsavenok:
Hi @matanmaze!

It seems like a malformed chunk was downloaded (probably empty string), but I'm not able to replicate this when quering large table :/

Few questions:
* Could you pls provide query id + deployment + datetime when this happened? This would help us find server-side logs for the chunk generation
* Are you able to consistently replicate the bug locally? Would it be possible to set up a proxy (I'll provide details on how to set it up) and collect all server responses to see what chunk metadata looked like and what it actually downloaded from the chunk url?
[...]

PR #1420, opened 2026-06-03
SNOW-3589640 Fix http error being hidden when streaming

sfc-gh-rsavenok:
### Description

Fixes SNOW-3589640 / #1418.

Interrupted streaming HTTP responses were being misreported to callbacks as
successful empty responses instead of errors. In `HttpClient.request`'s error
handler (`lib/http/base.js`), any failure that had an `err.response` was unwrapped
and passed to the callback as a success (`callback(null, response, err.response.data)`).
When a streamed response was interrupted after the status line arrived but before
the body did, `err.response.data` is `undefined`, so the callback received
`err = null` and an `undefined` body.

For large result sets this `undefined` body was stored as the chunk's rowset,
`chunk.getRows()` returned `undefined`, and `row_stream.js` called `.slice()` on it,
throwing a `TypeError`. That throw was then swallowed in `large_result_set.js`, so
the stream never emitted `'error'` or `'end'` — `streamRows()` hung forever.

The fix tightens the condition so an error response is only unwrapped as a
successful body when a body is actually present: the check changes from
`if (err.response)` to `if (err.response && err.response.data !== undefined)`.

Now any interrupted streamed response (status but no data) is routed through the
error path and surfaced to the callback as a real error, so consumers get an
`'error'` event instead of a silent hang. This applies to all callback-based HTTP
requests, not just large-result-set chunk downloads.

Added unit tests in `test/unit/http/base_test.ts`:
- an interrupted streaming response (`err.response` present, `data === undefined`)
  now invokes the callback with an error and a `null` body — this test fails without
  the change;
- a non-2xx response that carries a body (e.g. an S3 XML error) still unwraps the
  body and invokes the callback as a success (regression guard).

### Checklist

- [x]  Extend the types in index.d.ts file (if necessary)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Upcoming Release
 
-TBA
+Bugfixes:
+
+- Fixed interrupted streaming HTTP responses being treated as successful empty responses instead of errors, which could cause `streamRows()` to hang silently on large result sets (snowflakedb/snowflake-connector-nodejs#1420)
 
 ## 2.4.3
 
--- a/lib/http/base.js
+++ b/lib/http/base.js
@@ -89,7 +89,7 @@ HttpClient.prototype.request = function (options) {
               requestUtil.describeRequestFromOptions(requestOptions),
               err.response.status,
             );
-            options.callback(null, normalizeResponse(err.response), err.response.data);
+            options.callback(err, normalizeResponse(err.response), err.response.data);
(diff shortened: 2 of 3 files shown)


sfc-gh-pbulawa (approved):

sfc-gh-mcepiga (approved):
lgtm

sfc-gh-rkowalski (approved):

PR #1430, opened 2026-06-09
fix: replace hardcoded CRL nextUpdate expiry date in test utils

sfc-gh-dszmolka:
### Description

The default nextUpdate in createTestCRL() was hardcoded to 2026-06-08T00:00:00Z, which caused 5 CRL validator unit tests to fail once that date passed. Replaced with Date.now() + 1 year so tests remain valid regardless of when they run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Upcoming Release
 
-TBA
+Internal:
+
+- Fixed CRL validator unit tests failing due to hardcoded `nextUpdate` expiry date; replaced with a dynamic 1-year-from-now value (snowflakedb/snowflake-connector-nodejs#XXXX).
 
 ## 2.4.3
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## Upcoming Release
 
-TBA
+Internal:
+
+- Fixed CRL validator unit tests failing due to hardcoded `nextUpdate` expiry date; replaced with a dynamic 1-year-from-now value (snowflakedb/snowflake-connector-nodejs#1430).
```
sfc-gh-rsavenok: Use "no changelog" for this PR. Its a test only change so no need to list in the log

sfc-gh-rsavenok (approved):

CHANGELOG.md
sfc-gh-dszmolka: ack - done

PR #1438, opened 2026-06-24
SNOW-2013822 add customLogger option to funnel driver logs into user-supplied logger

### LINKED ISSUE #1045: Ability to provide own Logger
## What is the current behavior?

Not sure, as I can't find clear documentation or reference to how one might provide their own logging utility.

## What is the desired behavior?

The ability to provide, through a typed configuration property, a custom logger that receives log events from the Snowflake SDK and can handle them in any way it chooses. It should override all other forms of logging.

## How would this improve `snowflake-connector-nodejs`?

More flexibility for consumers that most likely have their own logging implementation. Writing to log files is not how everyone expects logs to be output - giving them the option to provide their own logger would cover more bases.

## References, Other Background

[This documentation](https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-logs) mentions having this available in the Node driver:

```
Logger.setInstance(new NodeLogger({ logFilePath: 'STDOUT'}));
```

But I need it available within docker containers on SPCS, running Node and the snowflake sdk.

Kremzon:
Adds a `customLogger` option to `snowflake.configure()` so users can route driver log messages through their own logging pipeline instead of the built-in file/console output.

Closes #1045

## How it works

When `customLogger` is set, the built-in winston file/console logging is bypassed entirely. Messages reach the custom logger already level-filtered, formatted, and secret-masked — the same processing that would normally happen before writing to `snowflake.log`. The method matching the SDK level is called (`error`/`warn`/`info`/`debug`/`trace`).

## Interface and compatibility

The expected interface is the same five-method shape used by pino, bunyan, and the built-in `console` object — all of which work drop-in:

```js
snowflake.configure({ customLogger: require("pino")() });
snowflake.configure({ customLogger: console });
```

Winston uses different level names and has no `.trace()`; the type definition includes a small adapter example for that case.

No config-time validation is done on the supplied logger — this matches the convention across the Node.js ecosystem (knex, kafkajs, fastify, etc.): duck typing, TypeError at runtime if a method is missing.

## Testing

- Unit tests: routing to a custom logger with no file side-effect, level filtering, and applying via `configure()` after construction
- Tested live against a real Snowflake account using pino (drop-in), a winston adapter, and a plain in-memory collector — all confirmed: logs routed correctly, no `snowflake.log` written

## Checklist

- [x] Tests which fail without the change
- [x] Unit tests pass (`npm run test:unit`)
- [x] Types extended in `index.d.ts`
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 New features:
 
 - Added `workloadIdentityAwsUseOutboundToken` connection option (default `false`) that switches AWS Workload Identity attestation to STS `GetWebIdentityToken` JWTs. This method is recommended and may become the default in a future release (snowflakedb/snowflake-connector-nodejs#1437)
+- Added a `customLogger` option to `snowflake.configure()` to route the driver's log messages through your own logger (e.g. winston, pino, OpenTelemetry) instead of the built-in file/console logging (snowflakedb/snowflake-connector-nodejs#1045).
 
 Bugfixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 New features:
 
 - Added `workloadIdentityAwsUseOutboundToken` connection option (default `false`) that switches AWS Workload Identity attestation to STS `GetWebIdentityToken` JWTs. This method is recommended and may become the default in a future release (snowflakedb/snowflake-connector-nodejs#1437)
-- Added a `customLogger` option to `snowflake.configure()` to route the driver's log messages through your own logger (e.g. winston, pino, OpenTelemetry) instead of the built-in file/console logging (snowflakedb/snowflake-connector-nodejs#1045).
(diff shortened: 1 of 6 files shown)


Kremzon:

sfc-gh-rsavenok:
Hi! @Kremzon 

Will review your PR next week. Need to check how the other drivers' logger customization APIs look so we have a consistent config.

Next release is scheduled for the first half of July, will make sure to include your PR

sfc-gh-rsavenok (approved):

sfc-gh-rsavenok:
PR looks good. Extend it with a fix to not include timestamp in the message string.

Will build and merge as soon as CI issues are resolved

Kremzon:
> PR looks good. Extend it with a fix to not include timestamp in the message string.
> 
> Will build and merge as soon as CI issues are resolved

Thanks :) 

@sfc-gh-rsavenok Do I need to resolve something in the CI? Or is there some broader unrelated issue?

sfc-gh-rsavenok:
> Do I need to resolve something in the CI? Or is there some broader unrelated issue?

I'll take care of it

PR #1440, opened 2026-06-26
SNOW-3709067 Skip GCS stage re-resolution when using a downscoped access token

sfc-gh-jchernandez:
## What

For a GCS **versioned-stage** upload (e.g. `PUT … 'snow://workspace/<fqn>/versions/live/…'`), the connector resolved the stage **twice** for a single PUT and ended up using the access token from the first resolution against the staging path from the second. When the server scopes upload credentials with an access token (rather than a presigned URL), the token is scoped to the wrong path and the upload fails with:

```
403 AccessDenied: …does not have storage.objects.create access …
  objects/stages/<uuid>/versions/<ver>/<rank>/…
```

GCS versioned-stage uploads only. **AWS and Azure are unaffected**, and **downloads are unaffected**.

## Root cause

`FileTransferAgent.updateFileMetasWithPresignedUrl()` re-runs the PUT statement **once per destination file**. Its sole purpose is to mint a **presigned URL** per object. But when the server scopes credentials with an **access token** instead, that re-resolution returns no presigned URL — just another token — and it overwrites `meta.stageInfo` with a freshly re-resolved staging path while the upload still authorizes with the token built from the *original* resolution.

## Server-side trigger (why the two resolutions diverge)

For certain versioned stages (e.g. `snow://workspace/…` live versions), the physical staging path contains a **per-PUT segment that the server advances on every resolution**. The access token's boundary is scoped to that per-PUT prefix. So two resolutions of the *same* logical destination produce two different physical paths:

- resolution #1 → `…/<rank_A>/`, token scoped to `…/<rank_A>/`
- resolution #2 → `…/<rank_B>/` (rank advanced)
- upload → URL from `<rank_B>` with the `<rank_A>`-scoped token → **403**

Both PUT requests carry the **identical** destination (`…/versions/live/<dir>/`); the diverging `<rank>` is assigned server-side per resolution, not sent by the client. Where the per-PUT segment does not advance (other deployments), both resolutions return the same path and the bug is invisible.

## Fix

A GCS upload session uses **one credential style throughout** — an access token **or** presigned URLs, never a mix. So when the stage resolution returns an access token, the connector can **skip the per-file re-resolution entirely**: one resolution already yields a token covering the whole staging directory, and every file uploads under that prefix.

```js
if (commandType === CMD_TYPE_UPLOAD) {
  if (stageInfo['creds'] && stageInfo['creds']['GCS_ACCESS_TOKEN']) {
    return; // access-token mode: one resolution covers all files; no presigned URL to fetch
  }
  // …legacy presigned-URL flow (re-resolve per file to mint each URL) unchanged…
}
```

This also removes a redundant per-file round-trip in access-token mode (1 PUT instead of N+1), matching the single-resolution behavior of the Python connector.

## Test correction

The request-retry mock for `Query PUT with GCS presigned URL refresh` modeled an **impossible mid-session credential-style flip** — an access token on the first resolution, then a presigned URL on the refresh. A session does not switch styles mid-stream. The mock now represents a presigned-URL session **consistently** (no access token in `creds`), so it still exercises the per-file presigned refresh it was written to cover.

## Testing

- **New unit test** `test/unit/file_transfer_agent/file_transfer_agent_test.js`:
  - access-token mode, **3-file folder**: asserts **zero** re-resolutions and that every file uploads under the single original resolution. Fails without the fix; passes with it.
  - presigned-URL mode: asserts **one** re-resolution per file (legacy flow preserved).
- **`test/integration/wiremock/testRequestRetry.ts` passes locally (118/118)**, including the corrected `Query PUT with GCS presigned URL refresh` suite.
- **Existing `gcs_test.js` unchanged and green.**
- **Live end-to-end** against a GCS preprod versioned stage:
  - single-file `PUT`: 403 → UPLOADED, now with a **single** resolution (token and upload path share the same `<rank>`).
  - **4-file folder**: all UPLOADED.
  - AWS and Azure uploads: still UPLOADED (regression check).

## Checklist

- [x] Tests which fail without the change
- [x] Unit tests pass (`npm run test:unit` for the affected area)
- [ ] index.d.ts — n/a (no public type change)
- [ ] README/docs — n/a
- [ ] JIRA/GitHub issue id — none yet; happy to attach


wiremock/mappings/query_put_gcs_presigned_ok.json.template
```diff
@@ -25,9 +25,7 @@
               "locationType": "GCS",
               "location": "gcs-bucket/path/",
               "presignedUrl": null,
-              "creds": {
-                "GCS_ACCESS_TOKEN": "test_gcs_token"
-              },
+              "creds": {},
```
sfc-gh-jchernandez, line 29: Why `creds` is empty here (and why the original `GCS_ACCESS_TOKEN` was removed):

A GCS upload session uses one credential style throughout — presigned URLs **or** a scoped access token, never a mix. This suite exercises the **presigned-URL** flow, so every stage resolution in it returns a presigned URL and no access token; `creds` is intentionally empty.

The previous fixture put an access token in this *initial* response and a presigned URL in the *refresh* below. That models an impossible mid-session credential-style switch. It also masks what this suite is meant to verify: once the connector holds an access token it skips the per-file re-resolution (one resolution covers all files), so the per-file presigned refresh this suite counts would never run. Keeping the session presigned-consistent (no token) preserves the refresh behaviour the test was written to cover, and is what makes the fix in `file_transfer_agent.js` observable here.

sfc-gh-rsavenok (approved):

PR #1457, opened 2026-07-23
Add `blockFilesystemAccess` configuration option

yehorkardash:
### Description

Adds `blockFilesystemAccess` to `snowflake.configure()`. When enabled, filesystem access used by file-transfer operations such as `PUT` and `GET` is blocked. The option defaults to `false`.

It does not affect driver-managed logging, OCSP/CRL caches, credential and private-key loading, configuration discovery, or fixed system-file reads.

### Testing

Added coverage for configuration validation and blocked `PUT`/`GET` filesystem access.

### Checklist


yehorkardash:

yehorkardash:
recheck

yehorkardash:
@sfc-gh-rsavenok, this is a fix for an issue reported in several CVEs([1](https://cvefeed.io/vuln/detail/CVE-2026-28381),[2](https://www.sentinelone.com/vulnerability-database/cve-2026-6442/)), would be grateful to get your eyes on this PR 🙏

Sorry for pinging!

sfc-gh-dszmolka:
this also makes performing large INSERTs with variable binding impossible

yehorkardash:
@sfc-gh-dszmolka, I think big file uploads should continue to function as usual. They're using `strbuffer` that is passed to FileTransferAgent as `filestream` parameter. When filestream parameter is present, the code doesn't try to read the file from filesystem

```
// file_transfer_agent.js
this.execute = async function () {
    if (fileStream) {
      // upload directly
    } else {
      // call guarded fs proxy
    }
```

sfc-gh-rsavenok:
Hi @yehorkardash!

We've seen Mythos flagging this and similar issues, but our current position is that the driver is not responsible for the SQL text provided in `.execute()`. Developers can just filter the SQL text on their end.

That being said, I think it makes sense to make it easier for developers to validate the SQL, but instead of `blockFilesystemAccess` it should be a query/connection option named `allowPutGet` or something like that.

All drivers are going to be rewritten in Rust (see https://github.com/snowflakedb/snowflake-connector-nodejs/issues/1256#issuecomment-3852678552) in the next 3-6 months, and such a feature is not on its roadmap.

I will close the PR, as we're not going to implement it in the current driver. Please file an issue with a detailed description of the use cases, and we'll add it to the [Universal Driver](https://github.com/snowflakedb/universal-driver) roadmap.

sfc-gh-rsavenok:
UPD: We already have `enablePutGet` config option in JDBC driver, so we need to implement similar in all UD drivers

PR #1464, opened 2026-08-10
Fix tests

sfc-gh-rsavenok:
### Description

GS released a change and error now has an id suffix:
* Before - `Incorrect username or password was specified.`
* After - `Incorrect username or password was specified. [263c39b3-e774-43d3-a1df-89083dda6a02]`

This resulted in our tests breaking. Fixed by matching pattern instead of exact string

### Checklist


sfc-gh-mcepiga (approved):
I know it's quick and small, but it'd really make me happy to not have this sort of inconsistencies (props access per dot notation vs brackets one)

PR #1475, opened 2026-09-03
harden: fix security issue in auth_oauth_authorization_code.js

anupamme:
## Summary
Harden input handling in `lib/authentication/auth_oauth_authorization_code.js` (flagged by multi_agent_ai).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `lib/authentication/auth_oauth_authorization_code.js:73` |
| **Assessment** | Defensive hardening |
| **Chain Complexity** | 2-step |

**Description**: OAuth access tokens and refresh tokens are cached using writeToCache() without visible encryption. The cache keys are built from token type, OAuth IDP URL, Snowflake host, username, and role. Without encryption at rest, an attacker with filesystem access to the cache storage could extract these tokens and replay them to impersonate users.

## Threat Model Context

This is a Node.js library - vulnerabilities affect downstream consumers who use this package.

## Changes
- `lib/authentication/auth_oauth_authorization_code.js`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/lib/authentication/auth_oauth_authorization_code.js
+++ b/lib/authentication/auth_oauth_authorization_code.js
@@ -13,6 +13,48 @@ const Util = require('../util');
 const AuthenticationTypes = require('./authentication_types');
 const { coordinateAuth } = require('./auth_coordinator');
 const { buildCacheKey, CacheTokenType } = require('./cache_key_builder');
+const crypto = require('crypto');
+const os = require('os');
+
+// Tokens are persisted on disk via writeToCache()/readCache(). Encrypt them at
+// rest so filesystem access to the cache file alone doesn't yield a directly
+// replayable OAuth access/refresh token.
+const CACHE_ENCRYPTION_KEY = crypto.scryptSync(
+  `${os.hostname()}:${os.homedir()}`,
+  'snowflake-oauth-token-cache',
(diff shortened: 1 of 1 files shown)


sfc-gh-rsavenok:
This encryption is pointless. If you know how encryption key is constructed, you know how to decrypt it.

Next major node driver release is going to replace disk credential manager with a keychain as default

anupamme:
Thanks for the review and for the clarification. I agree that the locally derived key doesn’t provide meaningful protection against someone who can access the cache, so I won’t pursue this encryption approach.

The keychain-backed credential manager sounds like the right solution for the underlying issue. I’ll leave this closed rather than introducing crypto/key-management complexity that will be superseded by that work.

PR #1479, opened 2026-09-09
SNOW-4017205 Make STS hostname configurable

sfc-gh-pfus:
### TL;DR

Adds a `workloadIdentityHost` connection option that lets users override the AWS STS endpoint used during Workload Identity authentication, enabling connectivity to AWS partitions not natively known to the connector.

### What changed?

- Introduced a new `workloadIdentityHost` connection option (string) for AWS Workload Identity flows. It accepts a bare host (`sts.sc2s.sgov.gov`), a host with port (`sts.example.com:8443`), or a full URL (`https://sts.example.com`). When not set, the connector defaults to the regional `sts.<region>.amazonaws.com` endpoint.
- Added `parseWorkloadIdentityHost` to normalize and validate the user-supplied value, and `regionalStsEndpoint` to construct the default regional endpoint. Both produce a typed `StsEndpoint` object consumed by the signing and SDK client code.
- The parsed endpoint is threaded through `getAwsAttestationToken`, `getCallerIdentityToken`, `getOutboundWebIdentityToken`, and `getAwsCredentials` (for role impersonation), replacing the previously hardcoded `getStsHostname` calls.
- When an override is active, the AWS SDK `STSClient` is configured with an explicit `endpoint` and FIPS/dualstack flags are suppressed (with a warning if those env vars are set), since the SDK rejects combining them with a custom endpoint.
- Validation that `workloadIdentityHost` is a string is performed in `ConnectionConfig`. A further check in `AuthWorkloadIdentity` rejects the option with a clear error when the provider is not AWS.
- Malformed overrides (wrong scheme, query strings, fragments, user info, empty value) are caught and reported as configuration errors before any region or credential lookups occur.
- Bumped the `toml` dependency from `^3.0.0` to `^5.0.0`.

### How to test?

- Unit tests cover `parseWorkloadIdentityHost` (valid and invalid inputs), `regionalStsEndpoint`, and the full `getAwsAttestationToken` flow with and without a `workloadIdentityHost` override, including impersonation and outbound token paths.
- A connection config test verifies that a non-string `workloadIdentityHost` is rejected at configuration time.
- Integration: configure a connection with `workloadIdentityProvider: 'AWS'` and `workloadIdentityHost: 'sts.sc2s.sgov.gov'` (or another non-commercial STS host) and confirm that the signed token's `url` and `host` header reflect the override.
- Confirm that setting `workloadIdentityHost` with a non-AWS provider (Azure, GCP, OIDC) produces an `InvalidParameterError`.

### Why make this change?

The connector previously hardcoded STS hostname resolution to commercial and China AWS partitions. Users operating in other AWS partitions (e.g., US government isolated regions) had no way to reach their STS endpoint without waiting for a connector release. The `workloadIdentityHost` option lets them supply the correct endpoint directly, decoupling partition support from the release cycle.


lib/connection/types.ts
```diff
@@ -339,6 +339,24 @@ export interface WIP_ConnectionOptions {
...
+  /**
+   * When workloadIdentityProvider=AWS, overrides the AWS STS host used by the attestation
+   * flows, so AWS partitions the connector does not know about can be reached without a
+   * connector release.
+   *
+   * Accepts a bare host (`sts.example.com`), a host with a port (`sts.example.com:8443`) or a
+   * full URL (`https://sts.example.com`). The value is used as given - no partition suffix is
+   * appended or rewritten - and `https` is assumed when no scheme is provided.
+   *
+   * When not set, the regional endpoint `sts.<region>.amazonaws.com` is used
```
sfc-gh-rsavenok: This comment should not say about china/non china. Later we might fix default behavior to support more regions and forget to list them here

lib/connection/types.ts
```diff
@@ -339,6 +339,24 @@ export interface WIP_ConnectionOptions {
    */
   workloadIdentityAwsUseOutboundToken?: boolean;
 
+  /**
+   * When workloadIdentityProvider=AWS, overrides the AWS STS host used by the attestation
+   * flows, so AWS partitions the connector does not know about can be reached without a
+   * connector release.
```
sfc-gh-rsavenok: what "without a connector release" means?

lib/connection/connection_config.js
```diff
@@ -724,6 +725,16 @@ function ConnectionConfig(options, validateCredentials, qaMode, clientInfo) {
     this.queryTag = options.queryTag;
   }
 
+  if (Util.exists(options.workloadIdentityHost)) {
```
sfc-gh-rsavenok: We don't do param validation anymore and trust typescript to do it. This is because new driver won't have them.

Follow https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/.cursor/rules/code.mdc

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Upcoming Release
 
+New features:
+
+- Added the `workloadIdentityHost` connection option that overrides the STS host used by the AWS Workload Identity flows, so AWS partitions the connector does not know about can be reached without a connector release. The value is used as given and defaults to the regional `sts.<region>.amazonaws.com` (snowflakedb/snowflake-connector-nodejs#1479)
```
sfc-gh-rsavenok: same question about "without connector release" and same comment about not including default driver `sts.<region>.amazonaws.com`

sfc-gh-rsavenok (commented):
Quick review done. I'll do a deeper look at attestation_aws and tests in few weeks closer to the next release

lib/connection/types.ts
sfc-gh-rsavenok: Ok, not I get that this is "without releasing new driver version" :) 

Even though the repo is named "snowflake-connector-nodejs", we mostly use "driver" terminology

I'd suggest something like "so AWS partitions unknown to the driver can be reached"

lib/connection/types.ts
sfc-gh-pfus: Good point - dropped the enumeration. The doc now just says the driver derives the STS host from the AWS region when the option is not set, so it will not go stale if we extend the default resolution later.

lib/connection/types.ts
sfc-gh-pfus: Yes, exactly that - "without releasing a new driver version". Adopted your wording: "so AWS partitions unknown to the driver can be reached". I also switched the StsEndpoint doc comment in attestation_aws.ts to the same phrasing so the terminology is consistent there too.

lib/connection/connection_config.js
sfc-gh-pfus: Removed the isString check (and its negative test) - TypeScript covers it.

Reading code.mdc I also noticed the rule against createInvalidParameterError / new ErrorCode entries, which the rest of this change was breaking, so the workloadIdentityHost parsing failures and the AWS-only guard now throw plain Errors instead:

- `Invalid workloadIdentityHost "ftp://sts.custom.snowflake.com": must use https or http, got scheme "ftp"`
- `workloadIdentityHost is supported only for AWS, got AZURE`

The second one now also matches the plain-Error style of the workloadIdentityImpersonationPath guard directly above it. I left the pre-existing createInvalidParameterError calls in that file alone, as they are out of scope for this PR.

CHANGELOG.md
sfc-gh-pfus: Reworded, no default host listed anymore: "... so AWS partitions unknown to the driver can be reached. The value is used as given; when not set, the STS host is derived from the AWS region."

sfc-gh-rsavenok (approved):

PR #1484, opened 2026-09-18
SNOW-4017205 - Polish sts endpoint logic

sfc-gh-rsavenok:
* StsEndpoint doesn't need to store complex metadata, we can store URL as StsClient accepts it. and we don't need overridden flag either
* e2e test added


lib/authentication/auth_workload_identity/attestation_aws.ts
```diff
@@ -112,25 +84,14 @@ export function parseWorkloadIdentityHost(workloadIdentityHost: string): StsEndp
       `Invalid workloadIdentityHost "${trimmed}": must use https or http, got scheme "${url.protocol.slice(0, -1)}"`,
     );
   }
-  if (!url.hostname) {
```
sfc-gh-rsavenok, line 115: URL always have a hostname, otherwise the construction `new URL(withScheme);` would crush

sfc-gh-pfus (approved):

